PR #15, opened 2019-05-20
Update googleapis version

scotthart:


WORKSPACE
```diff
@@ -14,15 +14,122 @@
 
 workspace(name = "com_github_googleapis_google_cloud_cpp_spanner")
 
+###
```
coryan: Shouldn't this be in the google_cloud_spanner_deps.bzl file too?  Yes, it is temporary, but we want that file to load all the necessary dependencies, right?

bazel/googleapis.BUILD
```diff
@@ -0,0 +1,198 @@
+# Copyright 2018 Google LLC
```
coryan: s/2018/2019/

bazel/google_cloud_cpp_spanner_deps.bzl
```diff
@@ -50,28 +50,28 @@ def google_cloud_cpp_spanner_deps():
...
+            build_file = "//bazel:googleapis.BUILD",
         )
 
     # Load gRPC and its dependencies, using a similar pattern to this function.
     # This implictly loads "com_google_protobuf", which we use.
     if "com_github_grpc_grpc" not in native.existing_rules():
         http_archive(
             name = "com_github_grpc_grpc",
-            strip_prefix = "grpc-1.19.1",
+            strip_prefix = "grpc-1.21.0-pre1",
```
coryan: Do you really want to depend on a pre-release?  Should we wait a couple of days and depend on the actual release?

scotthart: updated to grpc-1.21.0

scotthart:
Toggled CLA to yes as my email address in prior commits had the string "[diff]" appended.

WORKSPACE
scotthart: I'll refactor this in a follow on PR.

PR #37, opened 2019-05-31
Fix a few shellcheck warnings in the ci scripts.

mr-salty:


ci/kokoro/docker/build.sh
```diff
@@ -103,6 +102,7 @@ mkdir -p "${BUILD_HOME}"
 
 sudo docker run \
      --cap-add SYS_PTRACE \
+     "${interactive_flag}" \
```
coryan: As you probably noticed, when `${interactive_flag}` is empty docker reads this as "a command-line flag that does not start with --, it must be the name of the image".

mr-salty: blah...  i copied it from [here](https://github.com/googleapis/google-cloud-cpp/blob/559f5c94c65ee12902fa239c3dfad1777f3c8911/ci/travis/build-linux.sh#L52) but added the quotes, which is what sunk me... fixed (confirmed it works now with `< /dev/null` on my local machine)

PR #38, opened 2019-06-01
Implements spanner::Value skeleton with support for BOOL, INT64, FLOAT64, and STRING

### LINKED ISSUE #17: Implement spanner::Value skeleton
Implement a `spanner::Value` class. This should be a non-template class that can represent any of the allowed spanner types as defined in this proto: https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/type.proto

A rough sketch of the class would be something like the following:

```cc

namespace google {
namespace cloud {
namespace spanner {

class Value {
 public:
  // XXX: Use appropriate enablers for T
  template <typename T>
  explicit Value(T&& v);

  // Returns true if the value's type is T.
  //   bool b = value.is<int64_t>();
  template <typename T>
  bool is() const;

  // Type-safe getter, allows
  //   auto opt = value.get<int64_t>();
  //
  // Returns nullopt if the "null" value.
  // Crashes (or UB) if T is the wrong type.
  template <typename T>
  std::optional<T> get() const;

 private:
   // Probably holds a proto from spanner of some sort
};

}  // namespace spanner
}  // namespace cloud
}  // namespace google
```

devjgm:
closes #17 

This is just a skeleton of the class. It does not fully support all Spanner value types. Specifically it's missing

TIMESTAMP, DATE, BYTES, ARRAY, and STRUCT.

I will file separate GH issues to implement each of those separately.

(Note, I'm not actually working on a Saturday. I implemented this yesterday, but I wanted to trigger all the CI builds to see if this broke stuff.)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/38)

### DIFF
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -13,9 +13,10 @@
 // limitations under the License.
 
 #include "google/cloud/spanner/value.h"
+#include "google/cloud/terminate_handler.h"
 #include <cmath>
 #include <ios>
-#include <sstream>
+#include <string>
 
 // Implementation note: See
 // https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/type.proto
@@ -32,9 +33,7 @@ Value::Value(bool v) {
 
 Value::Value(std::int64_t v) {
   type_.set_code(google::spanner::v1::TypeCode::INT64);
-  std::stringstream ss;
-  ss << std::dec << v;
-  value_.set_string_value(ss.str());
+  value_.set_string_value(std::to_string(v));
 }
 
 Value::Value(double v) {
@@ -55,6 +54,7 @@ Value::Value(std::string v) {
 
 bool operator==(Value a, Value b) {
   if (a.type_.code() != b.type_.code()) return false;
+  if (a.is_null() && b.is_null()) return true;
   switch (a.type_.code()) {
     case google::spanner::v1::TypeCode::BOOL:
       return a.get<bool>() == b.get<bool>();
@@ -127,9 +127,16 @@ bool Value::GetValue(bool) const {
 std::int64_t Value::GetValue(std::int64_t) const {
   if (is_null() || !is<std::int64_t>())
     return {};  // XXX: Potentially UB or crash?
-  std::int64_t n;
-  std::istringstream(value_.string_value()) >> std::dec >> n;
-  return n;
+  auto const& s = value_.string_value();
+  std::size_t processed = 0;
+  long long x = std::stoll(s, &processed, 10);
+  if (processed != s.size()) {
+    // This should never happen, because we only parse strings that we created
+    // using std::to_string(), but it's good to be careful anyway.
+    std::string const err = "Failed to parse number from string: \"" + s + "\"";
+    google::cloud::Terminate(err.c_str());
+  }
+  return {x};
 }
 
 double Value::GetValue(double) const {
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -17,6 +17,10 @@
 #include <ios>
 #include <sstream>
 
+// Implementation note: See
+// https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/type.proto
+// for details about how Spanner types should be encoded.
+
 namespace google {
 namespace cloud {
 namespace spanner {
@@ -38,8 +42,7 @@ Value::Value(double v) {
   if (std::isnan(v)) {
     value_.set_string_value("NaN");
   } else if (std::isinf(v)) {
-    std::string const s = v < 0 ? "-Infinity" : "Infinity";
-    value_.set_string_value(s);
+    value_.set_string_value(v < 0 ? "-Infinity" : "Infinity");
   } else {
     value_.set_number_value(v);
   }
@@ -47,30 +50,46 @@ Value::Value(double v) {
 
 Value::Value(std::string v) {
   type_.set_code(google::spanner::v1::TypeCode::STRING);
-  value_.set_string_value(v);
+  value_.set_string_value(std::move(v));
 }
 
 bool operator==(Value a, Value b) {
-  auto const& at = a.type_;
-  auto const& bt = b.type_;
-  auto const& av = a.value_;
-  auto const& bv = b.value_;
-  if (at.code() != bt.code() || av.kind_case() != bv.kind_case()) return false;
-  switch (av.kind_case()) {
-    case google::protobuf::Value::kNullValue:
-      return true;
-    case google::protobuf::Value::kNumberValue:
-      return av.number_value() == bv.number_value();
-    case google::protobuf::Value::kStringValue:
-      return av.string_value() == bv.string_value();
+  if (a.type_.code() != b.type_.code()) return false;
+  switch (a.type_.code()) {
+    case google::spanner::v1::TypeCode::BOOL:
+      return a.get<bool>() == b.get<bool>();
+    case google::spanner::v1::TypeCode::INT64:
+      return a.get<std::int64_t>() == b.get<std::int64_t>();
+    case google::spanner::v1::TypeCode::FLOAT64:
+      return a.get<double>() == b.get<double>();
+    case google::spanner::v1::TypeCode::STRING:
+      return a.get<std::string>() == b.get<std::string>();
     default:
       return false;
   }
 }
 
-//
-// Value::is_null
-//
+std::ostream& operator<<(std::ostream& os, Value v) {
+  auto const& code = v.type_.code();
+  auto const* const descriptor = google::spanner::v1::TypeCode_descriptor();
+  os << "Value{" << descriptor->FindValueByNumber(code)->name() << ":";
+  if (v.is_null()) return os << "(null)}";
+  switch (code) {
+    case google::spanner::v1::TypeCode::BOOL:
+      os << v.get<bool>();
+      break;
+    case google::spanner::v1::TypeCode::INT64:
+      os << v.get<std::int64_t>();
+      break;
+    case google::spanner::v1::TypeCode::FLOAT64:
+      os << v.get<double>();
+      break;
+    case google::spanner::v1::TypeCode::STRING:
+      os << v.get<std::string>();
+      break;
+  }
+  return os << "}";
+}
 
 bool Value::is_null() const {
   return value_.kind_case() == google::protobuf::Value::kNullValue;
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/value.h
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+
+  friend bool operator==(Value a, Value b);
+  friend bool operator!=(Value a, Value b) { return !(a == b); }
+
+  // Returns true if there is no contained value.
+  bool is_null() const;
+
+  // Returns true if the contained value is of the specified type `T`. All
```
coryan: I think you want a Doxygen block here:

```C++
/**
 * Returns true if the contained value is of the specified type `T`.
 *
 * All `Value` instances ...
... ...
 */
```

Note how the first sentence is on its own, that makes it the `@brief` description for the function:

http://www.doxygen.nl/manual/commands.html#cmdbrief

google/cloud/spanner/value.h
```diff
@@ -0,0 +1,179 @@
...
+  //   spanner::Value v{true};
+  //   assert(v.is<bool>());
+  //   assert(v.is<google::cloud::optional<bool>>());
+  //
+  template <typename T>
+  bool is() const {
+    return IsType(T{});
+  }
+
+  // Returns the contained value if the specified type `T` matches the Value's
```
coryan: Ditto re: doxygen block.

google/cloud/spanner/value.h
```diff
@@ -0,0 +1,179 @@
...
+  bool IsType(bool) const;
+  bool IsType(std::int64_t) const;
+  bool IsType(double) const;
+  bool IsType(std::string) const;
+  template <typename T>
+  bool IsType(google::cloud::optional<T> opt) const {
+    return IsType(T{});
+  }
+
+  // Tag-dispatched function overloads. The arugment type is important, the
```
coryan: typo: ar**ug**ment.

google/cloud/spanner/value.cc
```diff
@@ -0,0 +1,156 @@
...
+    default:
+      return false;
+  }
+}
+
+std::ostream& operator<<(std::ostream& os, Value v) {
+  auto const& code = v.type_.code();
+  auto const* const descriptor = google::spanner::v1::TypeCode_descriptor();
+  os << "Value{" << descriptor->FindValueByNumber(code)->name() << ":";
+  if (v.is_null()) return os << "(null)}";
```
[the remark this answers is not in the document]
devjgm: Ack. FWIW, clang-format's Google style sets `AllowShortBlocksOnASingleLine: false`, which means that putting braces on the `if` will force it to wrap to 3-lines. So we can only achieve 1-line `if` statements without braces.

My *preference* is to allow one-line ifs without braces, but I definitely do not feel strongly about it. Do you mind if I leave this as-is for now, and see if the rest of the team has a preference and/or take an informal poll? If we agree to require braces and wrapping, that's fine with me and then I'll fix these cases and the few other places where we use one-line `if`s.

google/cloud/spanner/value.cc
```diff
@@ -0,0 +1,156 @@
...
+bool Value::GetValue(bool) const {
+  if (is_null() || !is<bool>()) return {};  // XXX: Potentially UB or crash?
+  return value_.bool_value();
+}
+
+std::int64_t Value::GetValue(std::int64_t) const {
+  if (is_null() || !is<std::int64_t>())
+    return {};  // XXX: Potentially UB or crash?
+  std::int64_t n;
+  std::istringstream(value_.string_value()) >> std::dec >> n;
```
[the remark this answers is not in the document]
devjgm: Good point. I changed to using `std::to_string` and `std::stoll` to check for errors.

google/cloud/spanner/value.cc
```diff
@@ -0,0 +1,156 @@
...
+//
+
+bool Value::GetValue(bool) const {
+  if (is_null() || !is<bool>()) return {};  // XXX: Potentially UB or crash?
+  return value_.bool_value();
+}
+
+std::int64_t Value::GetValue(std::int64_t) const {
+  if (is_null() || !is<std::int64_t>())
+    return {};  // XXX: Potentially UB or crash?
```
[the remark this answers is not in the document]
devjgm: No change yet, mainly because I still have a question about whether or not this is actually UB. In the header docs for `get<T>()` I say

```cc
   * Returns the contained value if the specified type `T` matches the Value's
   * type and if the contained value is not null. Otherwise, a default
   * constructed `T` is returned.
```

So as of right now, it's explicitly not UB. But of course, I contradict myself (a little) by still including this comment in the implementation about whether this decision is correct and whether perhaps it *should* be UB.

FWIW, the current guidance about returning a default constructed `T` comes from following the behavior of protobuf's and their `oneof` behavior. There is no UB or crashes. Callers can test for null and check the type (with `is<T>()`), but if they then try to access the value the assumption is that they want the behavior of "return a default constructed T if it doesn't have a T". If they don't want that behavior, they can avoid it by checking for null and the type explicitly. 

Additionally, callers could call `v.get<optional<T>>()`, in which case it'll return a disengaged optional if the value was null.

.... sooooooo, do you have thoughts on this point?

---

A 2-step alternative design:

1. Change get to look like:

```cc
template <typename T>
StatusOr<T> get() const;
```
This gives us a non-lethal way to signal the error to the caller... but it's a little more cumbersom to use, because StatusOr.

2. Add an explicit conversion operator, that does NOT return StatusOr<>, and declare that using this interface is UB to cast to the wrong type or if the Value is null. E.g.,

```cc
  template <typename T>
  explicit operator T() const;
...
  Value v = ...;
  std::int64_t i = static_cast<std::int64_t>(v);
```
This lets callers avoid the StatusOr if they want to, but then they have to code around UB.

... I kinda like this alternative design.

google/cloud/spanner/value.h
devjgm: Done. I think. FWIW, if there's a good doc showing your preferred doxygen style, that'd be great. Or if there's a good tutorial, that'd be helpful too. Otherwise, feel free to be nitpicky until I learn :-)

google/cloud/spanner/value.h
devjgm: Done, I think.

google/cloud/spanner/value.h
devjgm: Fixed.

google/cloud/spanner/value.cc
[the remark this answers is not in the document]
coryan: Please leave as-is.

google/cloud/spanner/value.cc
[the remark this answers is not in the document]
coryan: I like option (2) also.

google/cloud/spanner/value.cc
```diff
@@ -0,0 +1,163 @@
...
+  auto const& s = value_.string_value();
+  std::size_t processed = 0;
+  long long x = std::stoll(s, &processed, 10);
+  if (processed != s.size()) {
+    // This should never happen, because we only parse strings that we created
+    // using std::to_string(), but it's good to be careful anyway.
```
coryan: nit: Some of these strings will be generated by the service, so the comment is slightly wrong?

google/cloud/spanner/value.h
coryan: The standard documentation is pretty good.

http://www.doxygen.nl/manual/docblocks.html

I have settled on these things:

1. All types, functions, macros, namespaces in the public API get documentation, even "obvious" classes and functions.
1. Prefer `///` for types or functions that just have a one-liner documentation line.
1. For types or functions that get a long comment start with: (a) a one sentence description of that the thing is or does, then (b) the rest of the documentation. In other words, we use `JAVADOC_AUTOBRIEF=yes`.
1. We use `/** ... */` for long comments, we do not use `/*! ... */`.
1. Use `@` in preference `\` for commands (e.g. `@param` or `@return` instead of `\param` or `\return), no good reason, we just started with those...
1. For functions if you document one parameter you need to document all of them.
1. For functions document add the `@return` section if non-obvious.
1. Large important classes get large documentation, including links to the relevant docs in `cloud.google.com`, but add those late in the development cycle.
1. Using Doxygen for private members is entirely optional, we do not extract this documentation, but it is easier to use the same style for all comments.

google/cloud/spanner/value.cc
devjgm: Good point. Removed the comment that was misleading.

google/cloud/spanner/value.h
devjgm: Yes, this! Carlos's doxygen style tips. Thanks. I made a few updates to the doxygen comments for this file based on this.

google/cloud/spanner/value.cc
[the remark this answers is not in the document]
devjgm: I'm surprised, but this static_cast<> idea will be a little harder than I thought, because of an oddity of optional (including std::optional). See https://godbolt.org/z/Zw2pWQ for an example.

Basically, `static_cast<std::optional<int>>(foo)` will end up calling foo's `operator int()`, **not** its `operator std::optional<int>`. The reason is because of optional's conversion operator precedence and the definition of how static_assert has to behave. But this was a surprise to me.

Ultimately, this means that from within our templated `explicit operator T()` we'll never see `T` as an optional, only the optional's contained value. And that would mean that we could not correctly convert to an `optional<T>` that is disengaged for "null" Value objects.

That *might* be the nail in the coffin for option (2). I'll look at again tomorrow when I'm rested, but I wanted to record these findings somewhere in the meantime.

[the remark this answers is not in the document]
devjgm: OK, I just published a new version of this CL. Call it option 2a :-) Here are the major differences.

1. There are no optionals in the API at all. The only way to construct a "null" Value is via a new factory function `Value::MakeNull<T>()`

2. `Value::get<T>` now returns a `StatusOr<T>`. This object will only have a non-OK status if the programmer passed the wrong T or if the Value was null. There is no potential UB when calling `Value::get`.

3. There's an `explicit operator T() const` operator, which callers can use to `static_cast` Value objects to the desired type. This operator requires that the caller verify on their own that the specified `T` is correct and that the Value is not null. Otherwise, the result is undefined. This is an alternative to calling `Value::get`. It might be convenient for callers writing generic code where using a cast is desirable.

PTAL and re-verify approval.

google/cloud/spanner/value.cc
```diff
@@ -0,0 +1,173 @@
...
+
+template <typename T>
+Status CheckValidity(Value const& v) {
+  if (!v.is<T>()) return Status(StatusCode::kInvalidArgument, "wrong type");
+  if (v.is_null()) return Status(StatusCode::kInvalidArgument, "null value");
+  return {};  // OK status
+}
+
+}  // namespace
+Value::Value(bool v) {
```
coryan: nit: blank line after closing namespace?

google/cloud/spanner/value.cc
```diff
@@ -0,0 +1,173 @@
...
+}
+
+StatusOr<std::int64_t> Value::GetValue(std::int64_t) const {
+  auto const status = CheckValidity<std::int64_t>(*this);
+  if (!status.ok()) return status;
+  auto const& s = value_.string_value();
+  std::size_t processed = 0;
+  long long x = std::stoll(s, &processed, 10);
+  if (processed != s.size()) {
+    GCP_LOG(FATAL) << "Failed to parse number from string: \"" << s << "\"";
```
coryan: FYI: Right now `GCP_LOG(FATAL)` does not really terminate, we should probably fix that.

google/cloud/spanner/value.cc
devjgm: Fixed.

google/cloud/spanner/value.cc
devjgm: Ahh, ok. well, that's probably fine for now.

devjgm:
I'll let the test pass, then merge.

PR #39, opened 2019-06-01
Add CMake files and a CI build using them.

coryan:
These files allow us to build the spanner client using CMake, which we
need to support `make install` and package maintainers. We can also use
these to create code coverage builds, builds that run clang-tidy, and
builds generate the Doxygen documentation, all of which could be done
with Bazel, but would require rediscovering how to do it.

These files make no attempt at installing the dependencies for the
project. They assume these dependencies are installed and the build
fails if they are not. The CI build pre-installs these dependencies, a
future CI build will also test support for super builds.

Provide instructions for developers that want to use CMake with `vcpkg`
or by installing the dependencies in `/usr/local`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/39)


coryan:
PTAL.

PR #55, opened 2019-06-04
Refactoring to make it easier to add spanner ARRAY support.

devjgm:
Removed the switch statement in operator==() in favor of a custom
protobuf message differencer, which will grow/expand with any
data/values that we stuff in the proto. A custom differencer was needed
so that we could make encoded "NaN" values still appear unequal to each
other, as they're supposed to.

Removed the switch statement in operator<<() and changed to directly
printing the internal protobuf contents with `.ShortDebugString()`.
Since this format exposes internal implementation details that we don't
want external users or computers to depend on, I changed op<< ->
PrintTo() to make it clearer that this "formatting" is only inteded for
nice output during failed gtest assertions.

Changed the IsType() functions that used to return bool to instead
return the type code so that it can later be used in other places. Also
refactored those functions into a small, private TypeCodeMap struct.

	modified:   value.cc
	modified:   value.h
	modified:   value_test.cc

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/55)

### DIFF
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -14,8 +14,6 @@
 
 #include "google/cloud/spanner/value.h"
 #include "google/cloud/log.h"
-#include <google/protobuf/util/field_comparator.h>
-#include <google/protobuf/util/message_differencer.h>
 #include <cmath>
 #include <ios>
 #include <string>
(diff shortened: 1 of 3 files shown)


devjgm:
OK, for now I've reverted the MessageDifferencer changes, so that op==() still uses a switch and will do the Right Thing. The other changes will still help with spanner ARRAY types a little.

PTAL.

PR #57, opened 2019-06-04
Refactor the CI scripts for Bazel and CMake.

coryan:
Having two separate set of scripts helped me understand what was common,
but now it is time to refactor things, as I am about to add a *third*
type of build: CMake used as a super build.


coryan:
Ping.

PR #27, opened 2019-05-28
 Split the spanner sketch into pieces and make it build

mr-salty:
This is basically just Greg's sketch code (`jgm.cc`) - I split it up, added it to the bazel build, converted it to use cloud types, and it was already C++11 since we dropped variant. I did some formatting/cleanup.

Fixes #26

PR #60, opened 2019-06-05
Add skeleton classes based on the design doc.

mr-salty:
The methods are mostly no-ops.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/60)

### DIFF
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -21,6 +21,16 @@ find_package(googleapis CONFIG REQUIRED)
 configure_file(version_info.h.in version_info.h)
 add_library(spanner_client
             ${CMAKE_BINARY_DIR}/google/cloud/spanner/version_info.h
+            client.h
+            column.h
+            key.h
+            mutation.h
+            partition.h
+            result_set.h
+            result_stats.h
+            row.h
+            sql_statement.h
+            transaction.h
             value.cc
             value.h
             version.cc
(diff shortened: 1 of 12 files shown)


mr-salty:
This is an updated version of #27 - I think I addressed most/all of the comments from there (copyrights, comments, grouping methods, etc).  I also updated everything to what's in the current design doc and removed the "dummy" method implementations in favor of having them return empty/unimplemented.

I have a .cc file i was using to make sure this compiled and linked, but it's not really worth checking in... I should work on a proper example though.

devjgm:
Thanks, Todd.

First a general question.

I have a mild concern about adding skeleton code en mass based on the very rough sketch code from [jgm.cc](https://github.com/devjgm/google-cloud-cpp-spanner/blob/spanner-sketch/google/cloud/spanner/jgm.cc). That code was written quickly, with only enough thought or care to make it compile (probably with warnings). Its only goal was to give us some way to quickly experiment with and decide between various possible API choices. I think every line of that sketch code needs to be reconsidered and likely rewritten with intentional engineering. I'm hesitant to start with any of that sketch code in the code base for fear that it may accidentally live on.

We filed separate [issues](https://github.com/googleapis/google-cloud-cpp-spanner/issues) to implement the basics for each of the classes described in the design doc. The intent was to thoughtfully translate the abstractions from the design doc into properly written, well engineered code with unit tests. The "skeleton" part of that is that each abstraction need not be perfect or complete on the first PR, just some of the basic parts and a unit test.

So we could either tackle these abstractions one at a time, handling and closing the filed GH issues. Or would could do a more bulk laying down of skeleton code as in this PR. I was thinking we'd be doing the former. Do you think it's better to do the latter? Or some other option that I've missed?

mr-salty:
ok, after chatting with Greg, I'm going to abandon this - we should (collectively) split it into smaller and more fleshed out chunks.

PR #63, opened 2019-06-07
Make spanner::Value::is_null() a template

### LINKED ISSUE #49: Consider: making Value::is_null() a template
The [spanner::Value](https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/value.h) class currently has a non-templated `is_null()` method, because we do not need to know the contained value's type in order to determine if the contained value is null. That said, it might make sense to make the `is_null()` method a template so that it checks both the nullness and the type in one call. For example, we might want to consider something like the following:

```cc
class Value {
 public:
  ...
  template <typename T>
  bool is_null() const;
  ...
};
```

devbww:
We choose to treat `spanner::Value` as `T(null, v1, v2, ...)`
rather than `null + T(v1, v2, ...)` or something between the
two.  That is, the null value is typed.  `v.is_null<T>()` is false if
`!v.is<T>()`, even when v is null, but the typical decision tree
would always test `v.is<T>()` first.

Fixes #49 .

### DIFF
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -26,6 +26,7 @@ namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
+
 /**
  * The Value class represents a type-safe, nullable Spanner value.
  *
@@ -58,7 +59,7 @@ inline namespace SPANNER_CLIENT_NS {
  *     std::string msg = "hello";
  *     spanner::Value v(msg);
  *     assert(v.is<std::string>());
- *     assert(!v.is_null());
+ *     assert(!v.is_null<std::string>());
  *     StatusOr<std::string> copy = v.get<std::string>();
  *     if (copy) {
  *       std::cout << *copy;  // prints "hello"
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/value.h
```diff
@@ -135,6 +132,23 @@ class Value {
...
+   *     spanner::Value v{true};
+   *     assert(!v.is_null<bool>());
+   *
+   *     spanner::Value null_v = spanner::Value::MakeNull<bool>();
+   *     assert(v.is_null<bool>());
+   *     assert(!v.is_null<std::int64_t>());
+   */
+  template <typename T>
+  bool is_null() const {
+    return is<T>() && value_.kind_case() == google::protobuf::Value::kNullValue;
```
devjgm: Consider instead `return !HasValue<T>(type_, value_).ok();`

devbww: That looks different.  In particular, ...
```
Value null_a = spanner::Value::MakeNull<A>();
EXPECT_FALSE(null_a.is<B>());
```
would fail.  I added a test for that case.

devjgm: Ahh, right.

coryan:
I think the CMake build is failing because the config files did not exist when you created the branch (GitHub is dumb that way, it does not update your base when you push more commits.) You need to sync to `master`, rebase and push again, sorry.

PR #64, opened 2019-06-07
Create file to support super-builds and test them.

### LINKED ISSUE #40: Add instructions to installed CMake dependencies in $HOME/local
Some developers may prefer to install the dependencies for the project manually, but in `$HOME/local` instead of `/usr/local/`. The current instructions do not cover this case.

### LINKED ISSUE #42: Create a CI build to test the CMake files as a super-build.
One of the expected use-cases for the CMake files is to compile the project as part of a larger super-build. We need to test this in at least one of the CI builds.

coryan:
Create configuration files to verify our CMake files can be used as part
of a super-build.  These should probably live in google-cloud-cpp, but we
need to do a lot of refactoring there to allow this.

As a side benefit, we get a relatively easy way to install all the
dependencies in $HOME/local (or a similar directory). This is more
flexible than `vcpkg`, and less cumbersome than installing in
`/usr/local/`.

Add CI build configuration to automatically test this, though an additional
internal CL will be needed to enable this. 

Fixes #40 and fixes #42.

### DIFF
--- a/doc/setup-cmake-environment.md
+++ b/doc/setup-cmake-environment.md
@@ -39,6 +39,34 @@ The build takes a few minutes the first time, as it needs to create a Docker
 image with all the development tools and dependencies installed. Future builds
 (with no changes to the source) take only a few seconds.
 
+## Install the dependencies in `$HOME/local`.
+
+This is the recommended way to develop `google-cloud-cpp-spanner` using CMake.
+You will install the dependencies in `$HOME/local` (or a similar directory), and
+compile the project against these libraries. The installation needs to be done
+every time the version of the dependencies is changed, and it is not done
+automatically. But once installed you can use them for any build.
+
+Configure the super-build to install in `$HOME/local`:
+
+```console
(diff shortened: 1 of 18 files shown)


doc/setup-cmake-environment.md
```diff
@@ -47,8 +75,8 @@ However, some of the dependencies, such as `google-cloud-cpp` are only updated
 monthly. If at some point the Spanner client depends on recent changes on
 `google-cloud-cpp` this strategy will not work for you.
 
-With that out of the way, installing Ninja is recommended, because it makes the
-`vcpkg` builds faster, but not required:
+Though not required, we recommend that you install Ninja as it significantly
+speeds up the `vcpkg` builds:
```
ras0219-msft: We should actually download ninja automatically, so this shouldn't affect vcpkg total runtime. Are you seeing significant perf improvements by pre-installing ninja?

coryan: I have just been cargo-culting this since forever, thanks for the heads up, I will change the instructions.

PR #76, opened 2019-06-09
Cleanup: Prefer GetType() over type_.set_code; MakeNull is now a non-member

devjgm:
With this refactoring, the GetType() helper is now the only place that
maps C++ types to spanner type codes (e.g., std::int64_t ->
TypeCode::INT64). All the constructors now call GetType() to set the
type_ data member.

Now that "MakeNull" no longer needs to call a private constructor, a
MakeNullValue() factory is only a convenience function and does not need
any private or friend access to the value class. Therefore, it's nice to
make this a non-member function, which increases encapsulation since
fewer functions have access to Value's non-public members (this is a
pattern encouraged by many including Scott Meyers). Also,
spanner::MakeNullValue() is two fewer characters (removes an unneeded
"::").

	modified:   value.cc
	modified:   value.h
	modified:   value_test.cc

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/76)

### DIFF
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -53,7 +53,7 @@ inline namespace SPANNER_CLIENT_NS {
  * etc, but there is no default constructor because there is no default type.
  * Callers may create instances by passing any of the supported values (shown
  * in the table above) to the constructor. "Null" values are created using the
- * `Value::MakeNull<T>()` factory function or by passing an empty `optional<T>`
+ * `MakeNullValue<T>()` factory function or by passing an empty `optional<T>`
  * to the Value constructor..
  *
  * Example with a non-null value:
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/value.h
```diff
@@ -305,6 +294,18 @@ class Value {
   google::protobuf::Value value_;
 };
 
+/**
+ * Factory to construct a "null" Value of the specified type `T`.
+ *
+ * This is equivalent to passing to the constructor an `optional<T>` without a
```
devbww: Perhaps "equivalent to passing \<thing\> to the constructor" is clearer than "equivalent to passing to the constructor \<thing\>".

PR #77, opened 2019-06-09
Refactored c'tor bodies to MakeValueProto() function.

devjgm:
This refactoring makes all constructor bodies look the same: 1 statment
setting the type_ member using `MakeTypeProto()`, and 1 statement
setting the value_ member using `MakeValueProto()`. This way we can
re-use the make-a-value-proto logic without having to invoke the
constructor then steal the value_ data member from it.

This refactoring was done in preparation for adding Spanner Struct
support, which I suspect will involve some form of std::typle, which
should work well with helper functions like this.

	modified:   value.cc
	modified:   value.h

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/77)

### DIFF
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -20,10 +20,6 @@
 #include <ios>
 #include <string>
 
-// Implementation note: See
-// https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/type.proto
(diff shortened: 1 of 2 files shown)


devbww:
PS: There are some tidy errors.

PR #84, opened 2019-06-12
Update google-cloud-cpp.

coryan:
We need to pick up the fixes to `google::cloud::optional` to support
gcc-4.8.


coryan:
They are related. The new version of `google-cloud-cpp` removes the `internal::gitrev()` function, so I had to change them to use the its equivalent `internal::build_metadata()`.
BTW, something is not right with Kokoro, it is not creating all the builds. I am hoping it will be just a backlog that will clear by tomorrow.

devbww:
> They are related.

I see.  Thanks.

PR #86, opened 2019-06-12
Run all the formatters in the clang-tidy build.

coryan:
The intention was to create a single build that we could use to test
"most" things. I forgot to install the necessary tools in the Fedora
image, and to setup the right step in the `build-in-docker-cmake.sh`
script.

Thanks to @bww and @mr-salty for looking into this.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/86)


ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -35,6 +35,11 @@ echo
 echo "${COLOR_YELLOW}Starting docker build $(date) with $(nproc) cores${COLOR_RESET}"
 echo
 
+echo "================================================================"
+echo "Verify formatting $(date)"
```
devbww: While you're in this file, lines 61 and 62 (the "ctest" and the "echo Running tests") seem to be out of order.

ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -35,6 +35,11 @@ echo
...
+(cd "${PROJECT_ROOT}" ; ./ci/check-style.sh)
```
devbww: It looks like the image is missing "shellcheck".

PR #87, opened 2019-06-12
Add the package for "shellcheck"

devbww:
Install the ShellCheck package needed for check-style.sh.  Also move
the "chmod" in replace_original_if_changed to a more logical location.

Emit the "Running the unit tests" message in build-in-docker-cmake.sh
before actually running the tests.


coryan:
Thanks!

PR #90, opened 2019-06-13
Add build with C++17 turned on.

### LINKED ISSUE #89: Create a CI build that compiles with C++17 enabled.
This is to verify that our library can be used by developers using C++17, we cannot use any C++17 features though.

coryan:
Compile our code (but not our dependencies) with C++17. I needed to
introduce an extra configuration option because setting
-DCMAKE_CXX_STANDARD=17 in the command-line does not seem to work.

I tried to do this with Bazel, but without a CROSSTOOL file (which I
dread to tackle) the only way seems to add --cxxopt=-std=c++17. Bazel
*also* adds --cxxopt=-std=c++0x in that case, and I am not sure what the
result is. It also seems counterproductive to try to compile all our
dependencies with C++17.

Fixes #89.


coryan:
@devjgm I think you can rely on `__cplusplus >= 201703` because even MSVC has that defined correctly for C++17. Would need some experimentation I think.

PR #92, opened 2019-06-13
Static assert that vector of vector is not allowed.

### LINKED ISSUE #59: spanner::Value should disable construction from vector<vector<T>>
[Spanner documentation](https://cloud.google.com/spanner/docs/data-types) says that ARRAYs cannot directly include other ARRAYs. In C++, we use `std::vector` to implement ARRAYs. The templated constructor in `spanner::Value` is flexible enough to allow the construction of `std::vector<std::vector<T>>`, but since Spanner says this is not allowed, we should probably use `std::enable_if<>` to disable this constructor for vectors of vectors.

devjgm:
Fixes #59

This PR adds an `is_vector` metafunction and `static_assert`s in the
constructor's body to disallow vector of vector. I choose to use a
static_assert in the body here instead of an enabler in the template
parameter list for several reasons:

1. Using `std::enable_if` in the template parameter list requires that
the metafunction be defined first, so it would have to live above the
constructors in the Value class, instead of being "hidden" at the bottom
of the class in the private section.

2. The static_assert() produces a slightly clearer error message in this
case.

3. Enablers are most useful when the goal is to disable an overload so
that *some other overload* can be used instead. That's not the goal
here. In this case, we're just detecting an error case. There's no
alterative better constructor to use instead, so an enabler isn't really
useful.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/92)

### DIFF
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -159,8 +159,10 @@ class Value {
    * elements must have exactly the same field names. Any mismatch in in field
    * names results in undefined behavior.
    */
-  template <typename T>  // TODO(#59): add an enabler to disallow T==vector
+  template <typename T>
   explicit Value(std::vector<T> const& v) {
+    static_assert(!is_vector<typename std::decay<T>::type>::value,
+                  "vector of vector not allowed. See value.h documentation.");
     type_ = MakeTypeProto(v);
     value_ = MakeValueProto(v);
   }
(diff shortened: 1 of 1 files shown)


coryan:
I just pinged b/135102447 because there are builds stuck all over the place for spanner.

PR #96, opened 2019-06-13
Simplifies the Value constructor implementations.

devjgm:
All public constructors now call one private templated constructor that
sets the type_ and value_ members.

@coryan you previously referred to a simpler constructor /implementation/. I'm not sure if this is what you were referring to (maybe?), but I think I like this better than the repeated c'tor bodies that this code previously had.

Thoughts?

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/96)


devjgm:
PTAL

PR #97, opened 2019-06-14
cleanup: Mark uncommon CMake options as "advanced".

coryan:
This removes the options from the CMake UIs, and makes it easier for our
users to discover the interesting tweaks.


coryan:
Rebased to pickup the C++17 configs.

PR #98, opened 2019-06-14
Add the build badges to the top-level README.

coryan:


coryan:
To see how these look you need to check my branch:

https://github.com/coryan/google-cloud-cpp-spanner/tree/add-build-badges

PR #103, opened 2019-06-17
Sql statement skeleton

### LINKED ISSUE #33: Implement spanner::SqlStatement skeleton
Implement the skeleton of the SqlStatement classes, which should initially look something like the following:

```cc
namespace google {
namespace cloud {
namespace spanner {

class SqlStatement {
 public:
  using param_type = std::map<std::string, Value>;
  explicit SqlStatement(std::string sql);
  explicit SqlStatement(std::string sql, param_type params);
  // Value type, op==, op!=
  // ...

  std::string sql() const;
  param_type params() const;
};

}  // namespace spanner
}  // namespace cloud
}  // namespace google
```

## Design notes

This class represents an SQL statement with optional parameter placeholders of the form @param1 and their bound values.

scotthart:
fixes #33


google/cloud/spanner/sql_statement.cc
```diff
@@ -0,0 +1,40 @@
...
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+bool operator==(SqlStatement const& lhs, SqlStatement const& rhs) {
+  std::string s1 = lhs.statement_;
+  std::string s2 = rhs.statement_;
+  std::transform(std::begin(s1), std::end(s1), std::begin(s1),
+                 [](unsigned char c) { return std::tolower(c); });
+  std::transform(std::begin(s2), std::end(s2), std::begin(s2),
```
devjgm: It's too bad that we need to copy both statement strings. What do you think about writing a little helper function (in an unnamed namespace in this file) that takes two `std::string const&` arguments and does a case insensitive comparison without copying? It could quickly look at string sizes to short circuit the loop, then iterate both strings comparing the results of `std::tolower(s1[i]) != std::tolower(s2[i])`... or something?

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+#include "google/cloud/spanner/value.h"
+#include <string>
+#include <unordered_map>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+/**
+ * The SqlStatement class representis a SQL statement, that may or may not be
```
devjgm: s/a SQL/an SQL/g? I guess it depends on how one reads it "a sequel" vs "an ess-queue-ell".

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+ * parameterized.
+ *
+ * Parameter placeholders are specified by @<param name> in the SQL string.
```
devjgm: I wonder if this `@<param name>` thing here messes up doxygen, since it also uses like @foobar directives.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+ * Values for parameters are a collection of std::pair<std::string,
+ * google::cloud:spanner::Value>.
+ * Example:
+ *   google::cloud::spanner::SqlStatement stmt("select * from scmods where
```
devjgm: If you make the code example indented 4 spaces, doxygen will automatically put it in code font.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+ *   last_name = @param1", {{"param1", Value("Blues")}});
+ */
+class SqlStatement {
+ public:
+  /// Type alias for parameter collection.
+  using param_type = std::unordered_map<std::string, Value>;
+
+  /// Copy and move.
```
devjgm: Does this also support default construction? It seems like it could. It'd likely be equivalent to `SqlStatement("")` maybe?

google/cloud/spanner/sql_statement.h
coryan: `@` is magic in Doxygen, consider using backticks around `@<param name>`.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+/**
+ * The SqlStatement class representis a SQL statement, that may or may not be
+ * parameterized.
+ *
+ * Parameter placeholders are specified by @<param name> in the SQL string.
+ * Values for parameters are a collection of std::pair<std::string,
```
coryan: Also consider backticks around the long type.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+ * google::cloud:spanner::Value>.
+ * Example:
```
coryan: Consider `@par Example` so it is formatted as a new paragraph.

google/cloud/spanner/sql_statement.h
coryan: I think you need to insert 4 spaces to get a block.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+ *   google::cloud::spanner::SqlStatement stmt("select * from scmods where
+ *   last_name = @param1", {{"param1", Value("Blues")}});
+ */
+class SqlStatement {
+ public:
+  /// Type alias for parameter collection.
+  using param_type = std::unordered_map<std::string, Value>;
```
coryan: The naming convention is `ParamType` right?

google/cloud/spanner/sql_statement_test.cc
```diff
@@ -0,0 +1,76 @@
...
+  SqlStatement s1("SELECT * FROM TABLE SCMODS;",
+                  {{"last", Value("Blues")}, {"first", Value("Elwood")}});
+  SqlStatement s2("select * from table scmods;", {{"last", Value("Blues")}});
+  EXPECT_TRUE(s1 != s2);
+}
+
+TEST(SqlStatementTest, SqlAccessor) {
+  const char* statement = "select * from foo";
+  SqlStatement stmt(statement);
+  EXPECT_STREQ(statement, stmt.sql().c_str());
```
devjgm: Maybe just `EXPECT_EQ(statement, stmt.sql())` ?

google/cloud/spanner/sql_statement.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "google/cloud/spanner/sql_statement.h"
+#include <algorithm>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+bool operator==(SqlStatement const& lhs, SqlStatement const& rhs) {
+  std::string s1 = lhs.statement_;
```
coryan: Hmmm, I think `SELECT * FROM my_table WHERE foo='a capitalized string'` is a different statement than `SELECT * FROM my_table WHERE foo='A Capitalized String'`.  I know the keywords and functions are case-insensitive, but I sure the string values are not, and and I am not sure about table names and such.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+      : statement_(std::move(statement)), params_(std::move(params)) {}
+
+  /// Relational operators.
+  friend bool operator==(SqlStatement const& lhs, SqlStatement const& rhs);
+  friend bool operator!=(SqlStatement const& lhs, SqlStatement const& rhs);
+
+  std::string const& sql() const { return statement_; }
+  param_type const& params() const { return params_; }
+
+ private:
```
coryan: Should we have some kind of `operator<<()` to make it easier to debug these things?

google/cloud/spanner/sql_statement.cc
devjgm: That's a good point. Perhaps we should just punt on the case insensitivity of the keywords and stuff and just do a dead-simple case *sensitive* comparison here? I mean, in the C++ realm, "foo" != "FOO", so it may not be that surprising of SqlStatement behaves the same way for its stringy parts.

google/cloud/spanner/sql_statement.h
scotthart: I think it depends on if we're following google or STL conventions. I am comfortable with either and will defer to the greater consensus.

google/cloud/spanner/sql_statement.h
scotthart: If we allowed an empty SqlStatement, would it also follow that we would want mutators for statement_ and params_?

google/cloud/spanner/sql_statement.h
scotthart: Done.

google/cloud/spanner/sql_statement.h
scotthart: Backticks added.

google/cloud/spanner/sql_statement.h
scotthart: Indented

google/cloud/spanner/sql_statement_test.cc
scotthart: Done.

google/cloud/spanner/sql_statement.h
scotthart: Done.

google/cloud/spanner/sql_statement.h
scotthart: Done

google/cloud/spanner/sql_statement.h
scotthart: Done

google/cloud/spanner/sql_statement.h
scotthart: Done.

google/cloud/spanner/sql_statement.h
devjgm: Not necessarily. `SqlStatement` already supports mutations via assignment. So a default constructor could be useful like:

```cc
SqlStatement sql;
...
sql = ComputeStatement(...);
```

coryan: Not necessarily, you may still use the default constructor in something like this:

```C++
SqlStatement sql;
if (foo) { sql = ...; } else { sql = ...other-dots...; }
```

google/cloud/spanner/sql_statement.h
devjgm: You're right. FWIW, In a case like this, I'd say `ParamType` would be slightly preferred. We tend to use STL conventions when we either specifically need to "look like" an STL container and thus need to use their spellings, or when we're doing something very much analogous to an STL thing so we sometimes borrow their spelling. But this is a bit of a gray area: I'd probably write ParamType here.

google/cloud/spanner/sql_statement.h
scotthart: Punting on this until we resolve #105

google/cloud/spanner/sql_statement.h
scotthart: Done

google/cloud/spanner/sql_statement.h
scotthart: Done

google/cloud/spanner/sql_statement.cc
scotthart: Done.

google/cloud/spanner/sql_statement.cc
```diff
@@ -0,0 +1,58 @@
...
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace {
+bool CaseInsensitiveStringCompare(std::string const& lhs,
+                                  std::string const& rhs) {
+  if (lhs.size() == rhs.size()) {
+    for (int i = 0; i < static_cast<int>(lhs.size()); ++i) {
+      if (std::tolower(lhs[i]) != std::tolower(rhs[i])) {
```
coryan: I still believe that case-insensitive is not the right thing, I believe `select * from tbl where foo = 'FOO'` should be different than `select * from tbl where foo = 'foo'`.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,75 @@
...
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+/**
+ * The SqlStatement class represents an SQL statement, that may or may not be
+ * parameterized.
+ *
+ * Parameter placeholders are specified by `@<param name>` in the SQL string.
+ * Values for parameters are a collection of `std::pair<std::string,
+ * google::cloud:spanner::Value>`.
+ * @par Example:
```
coryan: Did you look at the Doxygen output?  It does not look right for me. You can check using:

```sh
./ci/kokoro/docker/build.sh clang-tidy
google-chrome cmake-out/spanci-fedora-install-30-clang-tidy/google/cloud/spanner/html/index.html
```

google/cloud/spanner/sql_statement.cc
remyabel: FYI, here are the [docs](https://cloud.google.com/spanner/docs/lexical#case-sensitivity) on case sensitivity.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,75 @@
...
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+/**
+ * The SqlStatement class represents an SQL statement, that may or may not be
+ * parameterized.
+ *
+ * Parameter placeholders are specified by `@<param name>` in the SQL string.
+ * Values for parameters are a collection of `std::pair<std::string,
+ * google::cloud:spanner::Value>`.
```
remyabel: Isn't the `value_type` of `unordered_map` actually `std::pair<const Key, T>`?

google/cloud/spanner/sql_statement.h
scotthart: Inserting a blank line before the indented code example appears to be critical.

google/cloud/spanner/sql_statement.h
scotthart: Done.

google/cloud/spanner/sql_statement.cc
scotthart: I have opted to remove equality comparison and make SqlStatement a semi-regular type. In addition to the intricacies of case sensitivity, I find it difficult to wrap my head around whether SqlStatement("select * from foo where colX = 1234) is equal or not equal to SqlStatement("select * from foo where colX = @value", {{"value", Value(1234.0)}}). Until such time as we determine equality is a necessary operator for this type, I would like to leave it out.

remyabel: I can see comparing the *serialized* strings of the statements being a valid use-case (i.e, in unit testing). Comparing the raw strings as-is may not make much sense.

ci/kokoro/docker/build.sh
```diff
@@ -33,7 +33,8 @@ elif [[ -n "${KOKORO_JOB_NAME:-}" ]]; then
   # files `$foo.cfg` for continuous builds and `$foo-presubmit.cfg` for
   # presubmit builds. Here we extract the value of "foo" and use it as the build
   # name.
-  export BUILD_NAME="$(basename "${KOKORO_JOB_NAME}" "-presubmit")"
+  BUILD_NAME="$(basename "${KOKORO_JOB_NAME}" "-presubmit")"
```
coryan: Seems like you `git merge` the changes from `master`, in the future please `git rebase`, it is less confusing as I do not need to review other changes that are really no-ops.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+#include "google/cloud/spanner/value.h"
+#include <string>
+#include <unordered_map>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+/**
+ * The SqlStatement class represents an SQL statement, that may or may not be
```
coryan: nit: (And I should have thought about this earlier) Consider a briefer "brief description" like:

```C++
/**
 * Represents a, potentially parameterized SQL statement.
 ...
..
 */
```

google/cloud/spanner/sql_statement.cc
devjgm: It's OK with me to punt on `op==` for now, but if you do, please file a new issue for us to consider adding it later.

Semiregular exists as a concept... because it has to. Semiregular types are still not necessarily something to strive for. It's like a use them when you must, but "regular" is to be preferred. See https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c11-make-concrete-types-regular

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+ * parameterized.
+ *
+ * Parameter placeholders are specified by `@<param name>` in the SQL string.
+ * Values for parameters are a collection of `std::pair<std::string const,
+ * google::cloud:spanner::Value>`.
+ * @par Example:
+ *
+ *     google::cloud::spanner::SqlStatement stmt("select * from scmods where
+ *     last_name = @param1", {{"param1", Value("Blues")}});
```
devjgm: The wrapping of this string param makes the example slightly difficult to read, IMO. Consider something like:
```
Example:

    using google::cloud::spanner::SqlStatement;
    using google::cloud::spanner::Value;
    SqlStatement stmt("select * from scmods where last_name = @param1",
                      {{"param1", Value("Blues")}});
```

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+  /// Type alias for parameter collection.
+  using ParamType = std::unordered_map<std::string, Value>;
+
+  /// Copy and move.
+  SqlStatement(SqlStatement const&) = default;
+  SqlStatement(SqlStatement&&) = default;
+  SqlStatement& operator=(SqlStatement const&) = default;
+  SqlStatement& operator=(SqlStatement&&) = default;
+
+  SqlStatement() = default;
```
devjgm: Nit: Please move this to be the first c'tor.

google/cloud/spanner/sql_statement.h
```diff
@@ -0,0 +1,72 @@
...
+  explicit SqlStatement(std::string statement)
+      : statement_(std::move(statement)) {}
+  /// Constructs a SqlStatement with specified parameters.
+  SqlStatement(std::string statement, ParamType params)
+      : statement_(std::move(statement)), params_(std::move(params)) {}
+
+  std::string const& sql() const { return statement_; }
+  ParamType const& params() const { return params_; }
+
+  friend std::ostream& operator<<(std::ostream& os, SqlStatement const& stmt);
```
devjgm: I thought you were punting on this until #105 was resolved? Anyway, adding it now is fine with me.

I think it would be worth putting a comment here that the output is intended for human consumption for debugging. The output format may change without notice in the future, so users should not rely on the format. I think this is important for stringifying objects which do not have an obvious canonical representation.

scotthart: I already had the code written prior to our discussion and figured I could change it if necessary later.

google/cloud/spanner/sql_statement.h
scotthart: Done

google/cloud/spanner/sql_statement.h
scotthart: Done

ci/kokoro/docker/build.sh
scotthart: I did attempt a git rebase, but obviously did something wrong as it later made me merge.

google/cloud/spanner/sql_statement.h
scotthart: Done.

PR #109, opened 2019-06-19
Support code coverage builds.

### LINKED ISSUE #7: Setup a CI build code coverage analysis.
We should have one CI building with code coverage and uploading to codecov.io.

coryan:
Introduce a new build type (`coverage`) which compiles our code,
but not the dependencies, with the --coverage flag. It then runs
the unit tests and uploads the code coverage stats to codecov.io.

Once enabled this fixes #7.


cmake/EnableCoverage.cmake
```diff
@@ -0,0 +1,68 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ~~~
+
+# For the GCC and Clang compiler families, enable a Coverage build type.
+if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "^(Clang|GNU)$")
+
```
devbww: Do we need/like the blank line after "if" here and below?

coryan: Maybe, cmake-format used to be stupid about formatting the comments all wrong without these extra lines. Let me try again with a newer version and fix them (in a separate PR).

PR #113, opened 2019-06-21
chore: Update google-cloud-cpp to version that installs googleapis proto files

scotthart:


cmake/external/google-cloud-cpp.cmake
```diff
@@ -25,10 +25,10 @@ if (NOT TARGET google-cloud-cpp-project)
     # downloaded from GitHub.
     set(
         GOOGLE_CLOUD_CPP_URL
-        "https://github.com/googleapis/google-cloud-cpp/archive/75b62d0bb931d475137342d2d3932e588161fe95.tar.gz"
+        "https://github.com/googleapis/google-cloud-cpp/archive/0126fe419e92ceef5c35cc0452adfd1224190393.tar.gz"
```
coryan: Do you mind making the same change in `bazel/google_cloud_cpp_spanner_deps.bzl`?  I would prefer if we kept those in sync.

scotthart: Done.

PR #115, opened 2019-06-22
Implement relational ops for spanner::Row

### LINKED ISSUE #112: Implement `op<` for spanner::Row
`spanner::Row` is conceptually a `std::tuple<Ts...>` so there's a logical way to implement the relational operators like <, <=, etc. We should probably implement these for Row, and test them. They should behave the same as they do for std::tuple.

remyabel:
Fixes #112

### DIFF
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -166,7 +166,7 @@ class Row {
     return a.values_ == b.values_;
   }
   friend bool operator!=(Row const& a, Row const& b) {
-    return !(a == b);
+    return a.values_ != b.values_;
   }
 
   /// Relational operators.
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -165,7 +165,9 @@ class Row {
   friend bool operator==(Row const& a, Row const& b) {
     return a.values_ == b.values_;
   }
-  friend bool operator!=(Row const& a, Row const& b) { return !(a == b); }
+  friend bool operator!=(Row const& a, Row const& b) {
+    return !(a == b);
+  }
 
   /// Relational operators.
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/row.h
```diff
@@ -167,6 +167,20 @@ class Row {
   }
   friend bool operator!=(Row const& a, Row const& b) { return !(a == b); }
 
+  /// Relational operators.
+  friend bool operator<(Row const& a, Row const& b) {
+    return a.values_ < b.values_;
+  }
+  friend bool operator<=(Row const& a, Row const& b) {
+    return std::rel_ops::operator<=(a.values_, b.values_);
```
devjgm: I don't think we gain much by using `std::rel_ops` here. I think it should be equivalent and perhaps a bit simpler to just have each of our relational operators forward to tuple's. So, how about instead just doing:

```cc
friend bool operator<=(Row const& a, Row const& b) {
  return a.values_ <= b.values_;
}
...
```
I think that'd be shorter too.

If you got this route, you probably want to fix the existing `operator!=` to follow this same pattern.

google/cloud/spanner/row.h
```diff
@@ -165,7 +165,9 @@ class Row {
   friend bool operator==(Row const& a, Row const& b) {
     return a.values_ == b.values_;
   }
-  friend bool operator!=(Row const& a, Row const& b) { return !(a == b); }
+  friend bool operator!=(Row const& a, Row const& b) {
+    return !(a == b);
```
devjgm: No, I'm thinking 
```cc
return a.value_ != b.value_;
```

that way all the relational operators are implemented by exactly forwarding the operation to the `.value_` (a std::tuple).

PR #120, opened 2019-06-24
feat: Implement spanner::Value support for TIMESTAMP and DATE

### LINKED ISSUE #50: Implement spanner::Value support for TIMESTAMP
Current `spanner::Value` implementation: https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/value.h

Spanner types to be supported: https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/type.proto

devbww:
TIMESTAMP is represented by `std::chrono::system_clock::time_point`,
and DATE by a new, simple "YMD" type, `google::cloud::spanner::Date`.

Also add conversion functions for `google::protobuf::Duration` and
`google::protobuf::Timestamp` to/from `std::chrono::nanoseconds`
and `std::chrono::system_clock::time_point` respectively.

Fixes #50 and #51.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/120)


google/cloud/spanner/date.h
```diff
@@ -0,0 +1,59 @@
...
+ * normalization. No calendar arithmetic is provided.
+ */
+class Date {
+ public:
+  Date() : Date(0, 0, 0) {}  // Note: bad month/day
+  Date(std::int64_t year, int month, int day)
+      : year_(year), month_(month), day_(day) {}
+
+  std::int64_t year() const { return year_; }
+  int month() const { return month_; }
```
coryan: nit: I think we should document if this are zero-based or 1-based.

google/cloud/spanner/internal/time.cc
```diff
@@ -0,0 +1,216 @@
...
+    t -= 1;
+    ss += seconds(1);
+  }
+  google::protobuf::Timestamp proto;
+  proto.set_seconds(t);
+  proto.set_nanos(ss.count());
+  return proto;
+}
+
+void ToProto(time_point tp, google::protobuf::Timestamp* proto) {
```
coryan: Also not tested.

google/cloud/spanner/internal/time.cc
```diff
@@ -0,0 +1,216 @@
...
+google::protobuf::Duration ToProto(nanoseconds ns) {
+  auto nanos = ns.count();
+  google::protobuf::Duration proto;
+  proto.set_seconds(nanos / nanoseconds::period::den);  // rounds toward zero
+  proto.set_nanos(nanos % nanoseconds::period::den);    // (a/b)*b + a%b == a
+  return proto;
+}
+
+void ToProto(nanoseconds ns, google::protobuf::Duration* proto) {
+  *proto = ToProto(ns);
```
coryan: In case you care, this function is not tested.

google/cloud/spanner/internal/time.cc
```diff
@@ -0,0 +1,216 @@
...
+  tp += nanoseconds(proto.nanos());
+  return tp;
+}
+
+namespace {
+
+// A duration capable of holding subsecond values at high precision.
+using femtoseconds = std::chrono::duration<std::int64_t, std::femto>;
+
+// Convert a std::time_t into a Zulu std::tm.
```
coryan: There is a lot of magic here that I am certain is "normal" if you have been doing cctz for a while, but it is all magic to me.  I am sure it is all correct, but maybe document it for future selves?

google/cloud/spanner/internal/time.cc
```diff
@@ -0,0 +1,216 @@
...
+std::tm ZTime(std::time_t const t) {
+  std::time_t sec = t % (24 * 60 * 60);
+  std::time_t day = t / (24 * 60 * 60);
+  if (sec < 0) {
+    sec += 24 * 60 * 60;
+    day -= 1;
+  }
+  day += 719468;  // since 1970-01-01
```
coryan: Since 1970-01-01 to when?  I guess 1900 as `std::tm` is 1900 based?

google/cloud/spanner/internal/time.cc
```diff
@@ -0,0 +1,216 @@
...
+  std::time_t const era = (day >= 0 ? day : day - 146096) / 146097;
```
coryan: Can we explain the magic here?

google/cloud/spanner/internal/time.cc
```diff
@@ -0,0 +1,216 @@
...
+  tm.tm_mon = m - 1;
+  tm.tm_mday = d;
+  tm.tm_hour = sec / (60 * 60);
+  tm.tm_min = (sec / 60) % 60;
+  tm.tm_sec = sec % 60;
+  return tm;
+}
+
+// Convert a Zulu std::tm into a std::time_t.
+std::time_t TimeZ(std::tm const& tm) {
```
coryan: Some of the same concerns about magic numbers here.

google/cloud/spanner/date.h
```diff
@@ -0,0 +1,59 @@
...
+ private:
+  std::int64_t year_;
+  int month_;
+  int day_;
+};
+
+inline bool operator==(Date const& a, Date const& b) {
+  return a.year() == b.year() && a.month() == b.month() && a.day() == b.day();
+}
+inline bool operator!=(Date const& a, Date const& b) { return !(a == b); }
```
devjgm: *No change necessary*

*If* we wanted to easily support ordering too, you could make the `year_` , `month_`, `day_` data members, into a `std::tuple<std::int64_t, int, int>`, and `std::tuple` already supports all the [relational operators](https://en.cppreference.com/w/cpp/utility/tuple/operator_cmp). So it would be easy for us to forward them all to tuple's implementation.

I realize that we're actively trying *not* provide a robust "Date" type here. So perhaps doing this would be a step in the wrong direction. But I thought I'd drop it here for your consideration. Not doing this is totally fine with me.

google/cloud/spanner/internal/date.cc
```diff
@@ -0,0 +1,60 @@
...
+
+namespace {
+
+// RFC3339 "full-date".
+char const kDateFormat[] = "%Y-%m-%d";
+
+}  // namespace
+
+std::string DateToString(Date d) {
+  std::tm tm;
```
devjgm: Does it matter that `tm` may have uninitialized fields when we call `std::put_time`? We are setting tm_year, tm_mon, and tm_mday, and I *assume* no other fields are accessed, but if they are, they'd be uninitialized.

We could avoid the question with `std::tm tm{};`. Thoughts?

google/cloud/spanner/internal/date.cc
```diff
@@ -0,0 +1,60 @@
...
+  return output.str();
+}
+
+Date DateFromString(std::string const& s) {
+  std::tm tm{};
+  std::istringstream input(s);
+  input >> std::get_time(&tm, kDateFormat);
+  if (!input || input.tellg() >= 0) {
+    // Failed to match kDateFormat, or did not consumed the entire string.
+    return Date{};
```
devjgm: Given how `Value` is currently implemented (its internal `GetValue`) function, I see what you're returning a default-constructed Date here instead of an error status. However, I'm thinking that `Value::GetValue` is going to need to change to return a `StatusOr<T>`, because each Value type could fail to convert a proto to the requested C++ type. In that case case, `Value::get<T>()` already has a `StatusOr<T>` return value, so we have room to return a nice error status, but it will need to be plumbed down into `Value::GetValue`.... but that's not yet done.

Sorry for how poorly worded and confusing I made that. Here's an issue I just filed to capture this: https://github.com/googleapis/google-cloud-cpp-spanner/issues/122

Anyway, in order to support that going forward, you might want to change these functions now to return StatusOr if they could possibly fail. That way they'll be easy to switch to the new `Value::GetValue()` model.

Thoughts?

google/cloud/spanner/internal/time.cc
```diff
@@ -0,0 +1,216 @@
...
+
+google::protobuf::Duration ToProto(nanoseconds ns) {
+  auto nanos = ns.count();
+  google::protobuf::Duration proto;
+  proto.set_seconds(nanos / nanoseconds::period::den);  // rounds toward zero
+  proto.set_nanos(nanos % nanoseconds::period::den);    // (a/b)*b + a%b == a
+  return proto;
+}
+
+void ToProto(nanoseconds ns, google::protobuf::Duration* proto) {
```
devjgm: Do we need this overload, or can we have just the one that returns by value?

coryan:
I think you need to rebase to fix the build failures.

devbww:
> General question: Is there any reasonable way for us to use cctz directly (https://github.com/google/cctz) for these time functions? Either as a real dependency, or maybe even as a copy in our source in a unique namespace so it never conflicts with anything else? Or is this too much work for the minimal time processing that we need?

I don't think this is worth doing anything about until we can just move to Abseil.

google/cloud/spanner/internal/time.cc
devbww: Removed.

google/cloud/spanner/internal/time.cc
devbww: Removed.

google/cloud/spanner/date.h
devbww: Noted that month will be [1 .. 12] and day will be [1 .. 31].

google/cloud/spanner/internal/time.cc
devbww: Removed.

google/cloud/spanner/internal/date.cc
devbww: As discussed offline, as long as we set the fields that `std::put_time()` is defined to read for the given format specifiers, we're in good shape.

The same is true for `std::get_time()` ... if it succeeds it is OK to read the fields corresponding to the format specifiers ... but clang-tidy doesn't understand that, so unfortunately we have to initialize them first.

google/cloud/spanner/date.h
devbww: Acknowledged.  I think it it best if we provide the fewest operations possible, until we can move the representation to something like `absl::CivilDay`.  Equality is only there because Value requires it.

google/cloud/spanner/internal/time.cc
devbww: It's pretty much pure magic!  But I added a link to a document that (attempts to) explain it, plus a few inline descriptions that might satisfy the slightly curious.

google/cloud/spanner/internal/time.cc
devbww: 719468 is the difference in days between 1970-01-01 and 0000-03-01.  I added a direct comment about that (and the use of Mar 1st instead of Jan 1st).  The 1900 for `tm_year` appears only in its immediate reads/writes.

google/cloud/spanner/internal/time.cc
devbww: 146097 is the number of days in the 400-year Gregorian cycle.  Added a direct comment about that.

google/cloud/spanner/internal/time.cc
devbww: I made it so the comment applies to both `ZTime()` [absolute -> civil] and `TimeZ()` [civil -> absolute].

Aside: `ZTime()` is just like `gmtime()`.  Unfortunately there is no standard inverse function, so we have to write our own, and at that point we're better off writing a `gmtime()` replacement too so that there is some correspondence between them, and so we can avoid `gmtime()` portability issues.

google/cloud/spanner/internal/date.cc
devbww: Done.

google/cloud/spanner/internal/time.h
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Convert a system_clock::time_point to a google::protobuf::Timestamp.
+ */
+google::protobuf::Timestamp ToProto(std::chrono::system_clock::time_point tp);
+
+/**
+ * Convert a google::protobuf::Timestamp to a system_clock::time_point.
+ */
+std::chrono::system_clock::time_point FromProto(
```
devjgm: There are some platforms where the system_clock duration is not nanoseconds. IIRC, Mac OS X has microsecond resolution on the system clock. Does it matter for these functions if we lose precision when going from protos to system_clock::time_point? 

As an alternative, we could define a nanosecond-resolution time_point on the system_clock, like:

```cc
using time_point_ns = std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds>;
```

... but then the range would be lessened. Bummer.

Another thought would be to require the caller to specify the time-point's duration (resolution), and make all of our functions that take or produce a time_point a template that uses `std::chrono::time_point<std::chrono::system_clock, D>`

Thoughts?

google/cloud/spanner/internal/date.cc
```diff
@@ -0,0 +1,62 @@
...
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+
+namespace {
+
+// RFC3339 "full-date".
+const char* const kDateFormat = "%Y-%m-%d";
```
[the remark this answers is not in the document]
devbww: Changed to `constexpr auto`.

google/cloud/spanner/value_test.cc
```diff
@@ -104,6 +106,48 @@ TEST(Value, BasicSemantics) {
...
+           Date(2262, 4, 12)    // after system_clock limit
+       }) {
+    SCOPED_TRACE("Testing: google::cloud::spanner::Date " +
+                 internal::DateToString(x));
+    TestBasicSemantics(x);
+    TestBasicSemantics(std::vector<Date>(5, x));
+    std::vector<optional<Date>> v(5, x);
+    v.resize(10, x);
+    TestBasicSemantics(v);
+  }
```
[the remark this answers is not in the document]
devbww: Done.

google/cloud/spanner/value_test.cc
```diff
@@ -104,6 +106,48 @@ TEST(Value, BasicSemantics) {
...
+           Date(2019, 6, 21),   // contemporary
+           Date(2038, 1, 19),   // around max 32-bit seconds limit
+           Date(2262, 4, 12)    // after system_clock limit
+       }) {
+    SCOPED_TRACE("Testing: google::cloud::spanner::Date " +
+                 internal::DateToString(x));
+    TestBasicSemantics(x);
+    TestBasicSemantics(std::vector<Date>(5, x));
+    std::vector<optional<Date>> v(5, x);
+    v.resize(10, x);
```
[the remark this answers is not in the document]
devbww: Fixed in #137.

google/cloud/spanner/internal/time.h
devbww: My thoughts were that `system_clock::time_point` would be sufficient, but I'm happy to revisit.  I concur that the other alternatives are a particular time_point of sufficient resolution and portable width, or rep/period templates all the way down.  I filed #138 to track this.

[Aside: Our spanci-fedora-libcxx-msan-30 image also has a microsecond system_clock.  The current test cases assume at least that resolution.]

devbww:
> nit (maybe): blank line?

So added.

devbww:
> Actually it is stdlibc++ that does not provide those functions right? We just don't compile with Clang on a distro that has such an old version of the library.

Acknowledged.  By "GCC" I was trying to refer to the entire (default) toolchain/libraries.  If there is a better term for that, I'm happy to add it.

That said, the code actually reads `__GNUC__ < 5`, so even if you think of "GCC" as just the compiler, the comment reflects that.

And all that said, I'd be happy if you came up with some cmake magic that could say whether the standard library it thinks I'm eventually going to link against has `std::get_time()` or `std::put_time()`, but that's no panacea either.

devbww:
> So we already have a FormatRfc3339 in google::cloud::internal, why not use that one? Or if this one is better, why not put this implementation there?
> 
> Similarly, we have a ParseRfc3339 function in google::cloud::internal, why not use that one and/or improve it?

Why not use them?  Because they do things like use `mktime()` and then try to determine the local UTC offset to compensate, and assume that the `system_clock` epoch is the UNIX epoch, etc.  In general I would say we're best off with `std::get_time()` and `std::put_time()` [modulo having to deal with libstdc++ history, which hopefully we can stop doing at some point].  Those APIs also don't fully support dates.

Why not improve them? That's what this is trying to do.

Why not put this implementation there?  Only because that is a bigger step.  In the end I imagine they wouldn't be there or here, but rather in the "common" repository.  But you have to start somewhere.

devbww:
> The include order is backwards here, we should include <time.h> after all the other headers. And do we want HAVE_GET_TIME defined before we included system headers?

OK, revamped the ordering.  Not defining something locally, like `HAVE_GET_TIME`, means we have to repeat the test, but that's OK.

devbww:
> Note that this is now ~1200 lines, should we split this?

OK.  Closing.

coryan:
I know this is now closed, just a couple of things to follow up on:

> Why not improve them? That's what this is trying to do.
> Why not put this implementation there?  Only because that is a bigger step.  In the end I imagine they wouldn't be there or here, but rather in the "common" repository.  But you have to start somewhere.

Ack. That makes sense, maybe a TODO+GitHub issue?

> OK, revamped the ordering. Not defining something locally, like `HAVE_GET_TIME`, means we have to repeat the test, but that's OK.

I did not express myself well.  I was thinking that the `#include <time.h>` should go last, then you can have the local defines without accidentally changing how the system includes interpret `HAVE_GET_TIME`.

devbww:
> That makes sense, maybe a TODO+GitHub issue?

Shall do (re: reconciling string <=> time/date functions).

> I was thinking that the `#include <time.h>` should go last

Ack, although I'm thinking that repeating the include test in the code (as opposed to introducing new, local defines) isn't so bad.  It is straightforward, and it is self testing (if it links, it is right).

PR #130, opened 2019-06-24
feat: Implement skeleton SpannerStub

### LINKED ISSUE #61: Implement a "SpannerStub" grpc wrapper
Since we will not have a gapic generated API to program to, we should implement our own class, I'll call it "internal::SpannerStub" (but it can be renamed to something better), to wrap gRPC and the raw RPC interface to spanner. This class will not be user facing, but it will be used by our user-facing `spanner::Client` class.

This class should keep in mind how callers will use it and test with it, so it may need to be gmock friendly.

mr-salty:
gRPC/GAPIC-like wrapper for the Spanner API.

Fixes #61

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/130)

### DIFF
--- a/google/cloud/spanner/internal/spanner_stub.cc
+++ b/google/cloud/spanner/internal/spanner_stub.cc
@@ -1,0 +1,257 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/spanner/internal/spanner_stub.h"
+#include <google/spanner/v1/spanner.grpc.pb.h>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+namespace {
+
+namespace spanner_proto = ::google::spanner::v1;
+
+/**
+ * DefaultSpannerStub - a stub that calls Spanner's gRPC interface.
+ */
+class DefaultSpannerStub : public SpannerStub {
+ public:
+  DefaultSpannerStub(
+      std::unique_ptr<spanner_proto::Spanner::StubInterface> grpc_stub)
+      : grpc_stub_(std::move(grpc_stub)) {}
+
+  DefaultSpannerStub(DefaultSpannerStub const&) = delete;
+  DefaultSpannerStub& operator=(DefaultSpannerStub const&) = delete;
+
+  StatusOr<spanner_proto::Session> CreateSession(
+      grpc::ClientContext& client_context,
+      spanner_proto::CreateSessionRequest const& request) override;
+  StatusOr<spanner_proto::Session> GetSession(
+      grpc::ClientContext& client_context,
+      spanner_proto::GetSessionRequest const& request) override;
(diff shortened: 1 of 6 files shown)


google/cloud/spanner/internal/spanner_stub.cc
```diff
@@ -0,0 +1,257 @@
...
+
+}  // namespace
+
+SpannerStub::~SpannerStub() = default;
+
+std::shared_ptr<SpannerStub> CreateDefaultSpannerStub(
+    ClientOptions const& options) {
+  auto channel =
+      grpc::CreateChannel(options.admin_endpoint(), options.credentials());
+  auto spanner_grpc_stub = spanner_proto::Spanner::NewStub(channel);
```
devjgm: Should the NewStub argument be `std::move(channel)`? Even if `CreateChannel` returns a `std::shared_ptr`, moving is more efficient if it's no longer needed in the current scope.

google/cloud/spanner/internal/spanner_stub.h
```diff
@@ -0,0 +1,109 @@
...
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+
+/**
+ * SpannerStub is a thin stub layer that avoids exposing the underlying
+ * transport stub directly (similar to what GAPIC will generate).
+ */
+class SpannerStub {
```
devjgm: For our future selves and other maintainers, it may be convenient to drop a link to the spanner.proto file in this comment. Here it is https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/spanner.proto

google/cloud/spanner/internal/spanner_stub.cc
```diff
@@ -0,0 +1,257 @@
...
+    spanner_proto::ListSessionsRequest const& request) {
+  spanner_proto::ListSessionsResponse response;
+  grpc::Status grpc_status =
+      grpc_stub_->ListSessions(&client_context, request, &response);
+  if (!grpc_status.ok()) {
+    return GRPCStatusToStatus(grpc_status);
+  }
+  return response;
+}
+Status DefaultSpannerStub::DeleteSession(
```
coryan: nit: insert blank line.

google/cloud/spanner/internal/spanner_stub.cc
coryan: `NewStub()` takes a `std::shared_ptr<grpc::Channel> const&`  a `std::move()` here would not help :disappointed:

google/cloud/spanner/internal/spanner_stub.cc
```diff
@@ -0,0 +1,257 @@
...
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+namespace {
+
+namespace spanner_proto = ::google::spanner::v1;
+
+/**
+ * DefaultSpannerStub - a stub that calls Spanner's gRPC interface.
+ */
+class DefaultSpannerStub : public SpannerStub {
```
coryan: While writing a unit test for this class is a pain, writing integration tests is needed. Can you at add a bug + TODO for the integration tests?

google/cloud/spanner/internal/spanner_stub.cc
devjgm: Oh, hmm. Thanks for pointing that out. Well, I guess this API did exist before C++11, and at that point passing shared_ptrs by value everywhere involved lots of unnecessary atomic ops, so passing by const-ref was probably better. In a C++11-world, they may have chosen differently.

**No change necessary here** :-)

[But while we're on the topic, I like that as a caller I kinda don't need to care about how the function accepts an argument. As a caller I only need to know if I am finished with a variable, in which case a `std::move()` is always safe (and never less efficient). And if I'm not finished with a variable, then a `std::move()` is never the right thing. This is nice because I can reason about the code locally without caring about the called function's actual signature. It also allows my code to magically become a bit more efficient if one day the function's signature changes to instead accept a by-value shared_ptr<T>.]

PR #132, opened 2019-06-25
Create integration tests for DatabaseAdminClient.

coryan:
Write actual integration tests, that is, tests that validate the
results, for spanner::DatabaseAdminClient. Change the CI build to run
this and future integration tests.

Remove the spanner_tool binary, it has become redundant now that we have
integration tests.

I think this approach is an improvement over what we do in `google-cloud-cpp`,
where we must add a little bit of bash code for each integration test.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/132)


google/cloud/spanner/integration_tests/README.md
```diff
@@ -0,0 +1,22 @@
+# Cloud Spanner C++ Client integration tests.
+
+This directory contains integration tests for the Cloud Spanner C++ client
+library. These tests are written using the Google Test framework, but can only
+execute when the environment properly configured.
```
devbww: s/environment/environment is/?

Issue #122, opened 2019-06-24
spanner::Value::GetValue should return StatusOr


devjgm:
The private (implementation detail) `Value::GetValue()` function *could* fail to parse a C++ value from a protobuf, but the function's signature currently leaves no way to report that error. See https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/value.cc#L190 for an example. Given that we'll be getting protobufs off the wire from the server, and it's *possible* that these protos are invalid and may fail to "parse", we should probably change `Value::GetValue` to return `StatusOr<T>` so that each implementation has room to return a nice error status.

After fixing this issue, all the private Value::GetValue functions will return `StatusOr<T>`. For example:

BEFORE CHANGE
```cc
  static double GetValue(double, google::protobuf::Value const&,
                         google::spanner::v1::Type const&);
  ...
```

AFTER CHANGE
```cc
  static StatusOr<double> GetValue(double, google::protobuf::Value const&,
                                   google::spanner::v1::Type const&);
  ...
```

devjgm:
Fixed in https://github.com/googleapis/google-cloud-cpp-spanner/pull/152

PR #137, opened 2019-06-25
fix: Value::GetValue(optional<T>, ...) must check for nullness

devbww:
`Value::GetValue(optional<T>, ...)` must check the `google::protobuf::Value`
for nullness if it is ever going to produce an `optional<T>{}`.

Change `Value.BasicSemantics` to correctly produce `vector<optional<T>>`
values that have some elements present and some missing.

Don't forward the `Value` private ctor argument to `MakeTypeProto()`.
`MakeTypeProto()` is responsible for taking its argument by value or
`const&`, and the ctor argument is only forwarded to `MakeValueProto()`.

Also use a non-throwing string->integer conversion function in
`Value::GetValue(std::int64_t, ...)`. This will help with #122.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/137)

### DIFF
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -32,6 +32,7 @@ namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
+
 class Value;  // Defined later in this file.
 // Internal implementation details that callers should not use.
 namespace internal {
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/value.h
```diff
@@ -447,6 +448,9 @@ class Value {
   static optional<T> GetValue(optional<T> const&,
                               google::protobuf::Value const& pv,
                               google::spanner::v1::Type const& pt) {
+    if (pv.kind_case() == google::protobuf::Value::kNullValue) {
+      return optional<T>{};
```
devjgm: Does `return {};` work?

devbww: Yes, and so modified.

PR #142, opened 2019-06-26
Add an abstraction layer for string <=> std::tm conversions

devbww:
For most toolchains we can use the C++11 `std::get_time()` and
`std::put_time()` functions, but these were not implemented in
the standard library used by GCC until version 5. Before that
we fallback to `strptime()` and `strftime()` respectively.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/142)


devbww:
> FYI: You and @mr-salty are racing

I win!

PR #144, opened 2019-06-27
Add support for converting Spanner DATE values to/from strings

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/144)


devbww:
I'm seeing these "leak" errors in random builds.  Given that it is unrelated to this PR, I'll merge.

```
google/cloud/spanner/database_admin_client_test.cc:148: ERROR: this mock object (used in test DatabaseAdminClientTest.HandleOperationDoneWithInvalidContents) should be deleted but never is. Its address is @0x22785a0.
ERROR: 1 leaked mock object found at program exit. Expectations on a mock object is verified when the object is destructed. Leaking a mock means that its expectations aren't verified, which is usually a test bug. If you really intend to leak a mock, you can suppress this error using testing::Mock::AllowLeak(mock_object), or you may use a fake or stub instead of a mock.
```

PR #146, opened 2019-06-27
Iterate tuple

devjgm:
Refactors the `IterateTuple` function from the private section of `Value` into `internal/tuple_utils.h` so that it can be used from other places (I'll need this from an upcoming "RowParser" class). The function is renamed to `internal::ForEach`.

This PR also makes the implementation a bit more generic so that it will work with not just `std::tuple`, but also any class that looks like a tuple, that is any class template that takes a variable number of template parameters. This will make `ForEach` work with the `Row<Ts...>` class.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/146)

### DIFF
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -41,6 +41,7 @@ add_library(spanner_client
             internal/time.h
             internal/time_format.cc
             internal/time_format.h
+            internal/tuple_utils.h
             row.h
             sql_statement.cc
             sql_statement.h
@@ -89,6 +90,7 @@ function (spanner_client_define_tests)
         internal/spanner_stub_test.cc
         internal/time_format_test.cc
         internal/time_test.cc
+        internal/tuple_utils_test.cc
         row_test.cc
         spanner_version_test.cc
         sql_statement_test.cc
--- a/google/cloud/spanner/internal/tuple_utils.h
+++ b/google/cloud/spanner/internal/tuple_utils.h
@@ -1,0 +1,72 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


devbww:
gcc-4.8 says:
```
google/cloud/spanner/internal/tuple_utils_test.cc: In member function 'virtual void google::cloud::spanner::v0::TupleUtils_ForEachStruct_Test::TestBody()':
google/cloud/spanner/internal/tuple_utils_test.cc:73:57: error: converting to 'std::tuple<bool, int>' from initializer list would use explicit constructor 'constexpr std::tuple<_T1, _T2>::tuple(_U1&&, _U2&&) [with _U1 = bool; _U2 = int; <template-parameter-2-3> = void; _T1 = bool; _T2 = int]'
   auto not_a_tuple = ns::NotATuple<bool, int>{{true, 42}};
                                                         ^
```

clang-tidy says:
```
/v/google/cloud/spanner/internal/tuple_utils_test.cc:66:58: error: invalid case style for function 'get' [readability-identifier-naming,-warnings-as-errors]
typename std::tuple_element<I, std::tuple<Ts...>>::type& get(
                                                         ^~~
                                                         Get
```

clang-3.8 says:
```
google/cloud/spanner/internal/tuple_utils_test.cc:73:47: error: chosen constructor is explicit in copy-initialization
  auto not_a_tuple = ns::NotATuple<bool, int>{{true, 42}};
                                              ^~~~~~~~~~
```

devjgm:
> Do you mean these to be Doxygen comments or just plain old comments? For a class in `internal` both are fine. Though I always dream of having a custom doxygen site that exposes the documentation for classes in `internal` so we (the developers of the libraries) can browse those too.

No, I didn't intend to use doxygen comments in this internal thing. AFAIU, doxygen doesn't do well with having one comment apply to several symbols/classes/functions, and sometimes that's needed for things like overload sets, template specializations, metafunctions, etc. Rather than trying to make doxygen understand these things, or caving to doxygen's requirements and perhaps making the header itself less clear, I opted to forget about doxygen and try to make the .h file clear since we don't use doxygen for "internal" things anyway currently.

But if we (the developers of the libraries) decide to start using doxygen for internal things too, then it'd probably be worth the effort to figure out how to both make doxygen happy and not harm devs who still read the .h files.

Also, I know very little about doxygen, so it's quite possible that there are good doxygen-y ways to document these things without harming .h readability, but I don't know it. Would be happy to learn!

PR #147, opened 2019-06-27
Add support for converting Spanner TIMESTAMP values to/from strings

devbww:
Also add `ToProto()`/`FromProto()` for `google::protobuf::Duration`
and `google::protobuf::Timestamp`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/147)


devbww:
> FYI: this code is not tested, in case you want to add a unit test for it.

Yeah ... this depends on the rounding behavior of the `std::chrono::system_clock::to_time_t(tp)` implementation.  It's allowed to do what it likes, although it probably always does the same thing.  I would floor in this case, which would mean that `tp - from_time_t(to_time_t))` is always positive.  If the implementation actually does that then the un-covered code is unreachable,

PR #150, opened 2019-06-27
Document the return value of ParseTime()

devbww:
Also use `std::size_t` instead of `std::string::size_type`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/150)

### DIFF
--- a/google/cloud/spanner/internal/time_format.h
+++ b/google/cloud/spanner/internal/time_format.h
@@ -16,6 +16,7 @@
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_SPANNER_INTERNAL_TIME_FORMAT_H_
 
 #include "google/cloud/spanner/version.h"
+#include <cstddef>
 #include <ctime>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/internal/time_format.h
```diff
@@ -33,10 +34,11 @@ std::string FormatTime(char const* fmt, std::tm const& tm);
...
 /**
  * Parse the date/time string `s` according to format string `fmt`, storing
- * the result in the std::tm addressed by `tm`.
+ * the result in the std::tm addressed by `tm`. Returns std::string::npos if
+ * the format string could not be matched. Otherwise returns the position of
+ * the first character not consumed (s.size() if the entire string matched).
  */
-std::string::size_type ParseTime(char const* fmt, std::string const& s,
-                                 std::tm* tm);
+std::size_t ParseTime(char const* fmt, std::string const& s, std::tm* tm);
```
devjgm: You probably want to change the return type in the .cc file too.

devbww:
> Is std::string::npos a good value to indicate error if the return type is no longer std::string::size_type?

I would say, yes.  @devjgm will remind us that the types are the same ... we're just being a little more succinct.

google/cloud/spanner/internal/time_format.h
devbww: Yep.

PR #154, opened 2019-06-28
cleanup: Make .clang-tidy consistent.

coryan:
I would like us to share these files, but that sounds hard, we should
keep them consistent in any case.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/154)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,7 +2,7 @@
 # Configure clang-tidy for this project.
 
 # Disabled:
-#  -google-readability-namespace-comments the BIGTABLE_CLIENT_NS is a macro, and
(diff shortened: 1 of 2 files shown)


devbww:
New version appears to pick up ...
```
/v/google/cloud/spanner/version.cc:25:28: error: invalid case style for static constant 'version' [readability-identifier-naming,-warnings-as-errors]
  static std::string const version = [] {
                           ^~~~~~~
                           kVersion
```

coryan:
> New version appears to pick up ...

Yeah... keep forgetting to cleanup my local build after changing .clang-tidy files. PTAL.

PR #156, opened 2019-06-28
feat: Add a default c'tor for Value

devjgm:
A default-constructed Value represents no value with no type. This is
needed in order to create a `std::array` of Value objects.

	modified:   value.h
	modified:   value_test.cc

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/156)

### DIFF
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -68,7 +68,7 @@ std::pair<google::spanner::v1::Type, google::protobuf::Value> ToProto(Value v);
  *     ARRAY/`std::vector`.
  *
  * Value is a regular C++ value type with support for copy, move, equality,
- * etc, but there is no default constructor because there is no default type.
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/value.h
```diff
@@ -136,7 +137,13 @@ std::pair<google::spanner::v1::Type, google::protobuf::Value> ToProto(Value v);
...
 class Value {
  public:
-  Value() = delete;
+  /**
+   * Constructs a non-null `Value` that holds nothing.
+   *
+   * All calls to `is<T>()`, `is_null<T>()`, and `get<T>()` will return false
+   * or an error as appropriate.
+   */
+  Value() = default;
```
devbww: I would consider

`type_.set_code(google::spanner::v1::TypeCode::TYPE_CODE_UNSPECIFIED);`

here, even if it may be the default.

PR #164, opened 2019-06-30
chore: Update google-cloud-cpp dep version.

remyabel:
This is to pick up the testing_util and grpc_util changes.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/164)


remyabel:
Actually, this won't work because required files will be missing.

PR #165, opened 2019-07-01
Remove dependency on bigtable_client.

### LINKED ISSUE #129: Remove dependency on bigtable_client.
We needed a function (`internal::MakeStatusFromGrpcError`) in the bigtable_client library, so I added it as a dependency. We should refactor the function to a new library. Note that the function cannot be moved to `google_cloud_cpp_common` as it depends on gRPC and we do not want Cloud Storage to depend on gRPC.

remyabel:
Fixes #129 

This updates the `google-cloud-cpp` dep and changes any appropriate references to `grpc_utils`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/165)


coryan:
Thanks!

PR #169, opened 2019-07-05
feat: Use a fixed type for TIMESTAMP

### LINKED ISSUE #138: Finalize representation of TIMESTAMP values
Currently we use `std::chrono::system_clock::time_point` as the C++ representation of a Spanner TIMESTAMP value.  This has one possible shortcoming in that the resolution of `system_clock` is platform dependent, so a timestamp generated on one machine (possibly by the Spanner implementation, or as user data) may be truncated when reconstituted on another machine.

We should decide if this should be changed.  One possibility is using a platform-independent representation which is deemed of sufficient range and resolution (like a 64-bit count of nanoseconds, which would span 1678 to 2261).  Another is to allow the user to specify the rep/period of the `time_point` when we convert to/from the RFC3339 string.

devbww:
Previously we chose `std::chrono::system_clock::time_point` to represent
a spanner TIMESTAMP, but it has a platform-specific resolution. Now
we use `time_point<system_clock, nanoseconds>` to explicitly match the
resolution required by Spanner. This allows a timestamp generated on one
machine to be reconstituted intact on another machine.

Revert to testing with nanoseconds. Previously we used microseconds as
that was the common resolution of `system_clock::time_point` over all
supported platforms.

Fixes #138.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/169)

### DIFF
--- a/google/cloud/spanner/internal/time.cc
+++ b/google/cloud/spanner/internal/time.cc
@@ -15,7 +15,6 @@
 #include "google/cloud/spanner/internal/time.h"
 #include "google/cloud/spanner/internal/time_format.h"
 #include <chrono>
-#include <cstdint>
 #include <cstring>
 #include <ctime>
 #include <iomanip>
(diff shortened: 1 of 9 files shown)


google/cloud/spanner/internal/time.h
```diff
@@ -42,26 +42,24 @@ std::chrono::nanoseconds FromProto(google::protobuf::Duration const& proto);
...
 
 /**
  * Convert a google::protobuf::Timestamp to a system_clock::time_point.
  */
-std::chrono::system_clock::time_point FromProto(
-    google::protobuf::Timestamp const& proto);
+Timestamp FromProto(google::protobuf::Timestamp const& proto);
 
 /**
  * Convert a system_clock::time_point to an RFC3339 "date-time".
```
coryan: nit: the comment does not match the function any more?

google/cloud/spanner/internal/time.h
```diff
@@ -42,26 +42,24 @@ std::chrono::nanoseconds FromProto(google::protobuf::Duration const& proto);
...
  */
-std::string TimestampToString(std::chrono::system_clock::time_point tp);
+std::string TimestampToString(Timestamp ts);
 
 /**
  * Convert an RFC3339 "date-time" to a system_clock::time_point.
```
coryan: nit: the comment does not match the function.

google/cloud/spanner/internal/time.h
```diff
@@ -42,26 +42,24 @@ std::chrono::nanoseconds FromProto(google::protobuf::Duration const& proto);
 /**
  * Convert a system_clock::time_point to a google::protobuf::Timestamp.
```
coryan: nit: the comment does not match the function.

google/cloud/spanner/internal/time.cc
```diff
@@ -29,156 +28,161 @@ namespace spanner {
...
-  if (sec < 0) {
-    sec += 24 * 60 * 60;
-    day -= 1;
+//
+// All the civil-time code assumes the proleptic Gregorian calendar, with
+// 24-hour days divided into 60-minute hours and 60-second minutes.
+std::tm ZTime(std::chrono::seconds s) {
+  using rep = std::chrono::seconds::rep;
+  constexpr auto kSecsPerDay = std::chrono::hours::period::num * 24;
+  using days = std::chrono::duration<rep, std::ratio<kSecsPerDay>>;  // C++20
```
coryan: nit: what does the `// C++20` comment mean?  That you miss that feature from C++20?

google/cloud/spanner/internal/time.h
devbww: Fixed.

google/cloud/spanner/internal/time.h
devbww: Fixed.

google/cloud/spanner/internal/time.h
devbww: Fixed.

google/cloud/spanner/internal/time.cc
devbww: It was supposed to hint that the std::chrono::days duration exists in C++20, but the short hint is indeed more confusing that it is worth, and nor is it useful to go to any lengths to describe the future.    So, deleted.

PR #171, opened 2019-07-08
Add RowParser

### LINKED ISSUE #32: Implement Row<...> and RowRange<...> skeleton
Implement the skeleton of the `Row<Ts..>` and `RowRange<Ts...>` classes, which should initially look something like the following:

(See also https://godbolt.org/z/sP4DTt)

```cc
namespace google {
namespace cloud {
namespace spanner {

template <typename... Ts>
class Row {
 public:
  using TupleType = std::tuple<Ts...>;

  Row() = default;
  explicit Row(TupleType tt) : tuple_(std::move(tt)) {}

  TupleType const& get() { return tuple_; }

  template <std::size_t... Is>
  auto get() -> decltype(std::get<Is...>(get())) {
    // This doesn't actually work. But it gives you the idea of what we want. Should be pretty
    // easy to implement ourselves.
    return std::get<Is...>(get());
  }

 private:
  TupleType tuple_;
};

template <typename... Ts>
class RowRange {
 public:
  using value_type = StatusOr<Row<Ts...>>;
  using iterator = typename std::vector<value_type>::iterator;
  using const_iterator = typename std::vector<value_type>::const_iterator;

  const_iterator begin() const;
  const_iterator end() const;
};

}  // namespace spanner
}  // namespace cloud
}  // namespace google
```

## Design notes

A `spanner::RowRange<Ts..>` represents a range of rows, with each row containing the specified C++ types. A caller constructs a RowRange by calling `spanner::ResultSet::rows<Ts…>()`, and specifying the C++ types that correspond to the columns of each row. The rows can be accessed by iterating the row range, which will produce a `StatusOr<Row<Ts…>>`. 

Note that the specified column types must exactly match the supported Spanner types. The code will not compile if a field is specified as, say, `std::int32_t`, which is not a supported type. 

A `spanner::Row<Ts…>` represents a range of values for each column in the row. Callers can access the values in a type-safe way using the `spanner::Row<Ts…>::get()` functions, which can return tuples or values depending on how they're called.

devjgm:
Fixes #32

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/171)

### DIFF
--- a/google/cloud/spanner/row_parser.h
+++ b/google/cloud/spanner/row_parser.h
@@ -60,10 +60,61 @@ inline namespace SPANNER_CLIENT_NS {
  */
 template <typename ValueIterator, typename... Ts>
 class RowParser {
-  class RowIterator;  // Defined in private section below.
-
  public:
-  using iterator = RowIterator;
+  /// A single-pass input iterator that coalesces multiple `Value` results into
+  /// a `Row<Ts...>`.
+  class iterator {
+   public:
+    using RowType = Row<Ts...>;
+    using iterator_category = std::input_iterator_tag;
+    using value_type = StatusOr<RowType>;
+    using difference_type = std::ptrdiff_t;
+    using pointer = value_type*;
+    using reference = value_type&;
+
+    reference operator*() { return curr_; }
+    pointer operator->() { return &curr_; }
+
+    iterator& operator++() {
+      curr_ = value_type{};
+      if (it_ == end_) return *this;
+      std::array<Value, RowType::size()> values;
+      for (auto& e : values) {
+        if (it_ == end_) {
+          curr_ = Status(StatusCode::kUnknown, "incomplete row");
+          return *this;
+        }
+        e = *it_++;
+      }
+      curr_ = ParseRow<Ts...>(values);
+      return *this;
+    }
+
+    iterator operator++(int) {
+      auto const old = *this;
+      operator++();
+      return old;
+    }
+
+    friend bool operator==(iterator const& a, iterator const& b) {
+      return std::tie(a.curr_, a.it_, a.end_) ==
+             std::tie(b.curr_, b.it_, b.end_);
+    }
+    friend bool operator!=(iterator const& a, iterator const& b) {
+      return !(a == b);
+    }
+
+   private:
+    friend RowParser;
+    explicit iterator(ValueIterator begin, ValueIterator end)
+        : it_(begin), end_(end) {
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/row_parser.h
```diff
@@ -0,0 +1,165 @@
...
+  /// Returns the begin iterator.
+  iterator begin() { return iterator(it_, end_); }
+
+  /// Returns the end iterator.
+  iterator end() { return iterator(end_, end_); }
+
+ private:
+  // The RowParser's interator type. Models InputIterator so it can only
+  // iterate forward and can only do so once.
+  class RowIterator {
```
coryan: Why not just a public class called `iterator`?  In any case, this is part of the public interface, it needs documentation like any other class...  Either here or in line 66 something like this maybe?

```C++
/// A single-pass input iterator that coalesces multiple `Value` results into a `RowType`.
```

google/cloud/spanner/row_parser.h
```diff
@@ -0,0 +1,165 @@
...
+#include <tuple>
+#include <utility>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+/**
+ * A RowParser takes a range of `Value` objects and a list of types, and it
```
coryan: I would expect you need to receive a range of `StatusOr<Value>` objects? The "unchunker" is likely to produce ranges of `StatusOr<Value>` because the iteration might fail?

google/cloud/spanner/row_parser.h
devjgm: Good point. Done. This is better because the iterator's member aliases and stuff really are public API points.

FTR, the reason it was private was to prevent others from directly creating instances of this iterator. But I can do the same thing by making the class public but making its c'tor private (and friending `RowParser`), which I've done.

google/cloud/spanner/row_parser.h
devjgm: That, uhh, might be true? I'm not sure yet exactly what the connection will be between the RowParser and the stream of [PartialResultSet](https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/result_set.proto#L61) will be. Yes, something will need to unchunk it, and I'm not sure if there's another way to expose Status for a failed unchunking.

But it should be easy to adapt this code to accept an input range of `StatusOr<Value>` instead (I think).

PR #172, opened 2019-07-09
fix: RowParser consumes Values from functor, not input range

devjgm:
`RowParser<Ts...>` no longer accepts an input range of `Value` objects. Instead, it now consumes `Values`s from a caller-provided functor. This functor will allow the `Client` class to inject code to read protos and values from a gRPC stream.

Other smaller changes in this PR:

*  RowParser now requires at least 1 template parameter. It made not sense to have a row parser that was parsing nothing for each row.
* Any failure during iteration will halt further iteration by advancing the iterator to end.
* The caller-provided `ValueSource` functor may return a `StatusOr` to indicate a failure producing the `Value` object. (@coryan 's prediction yesterday was correct :-) )

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/172)

### DIFF
--- a/google/cloud/spanner/row_parser.h
+++ b/google/cloud/spanner/row_parser.h
@@ -21,6 +21,7 @@
 #include "google/cloud/status.h"
 #include "google/cloud/status_or.h"
 #include <array>
+#include <functional>
 #include <tuple>
 #include <utility>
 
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/row_parser.h
```diff
@@ -30,42 +31,60 @@ namespace spanner {
...
- *         Value(true), Value(1),  // Row 1
- *         Value(true), Value(2),  // Row 2
- *         Value(true), Value(3),  // Row 3
- *     };
- *     for (auto row : MakeRowParser<bool, std::int64_t>(values)) {
+ *     ValueSource vs = ...
+ *     RowParser<bool, std::int64_t> rp(std::move(vs));
+ *     for (auto row : rp) {
  *       if (!row) {
  *         // handle error
```
devbww: Should we explicitly `continue` (or perhaps `break`) here?

google/cloud/spanner/row_parser.h
```diff
@@ -97,30 +128,29 @@ class RowParser {
     }
 
     friend bool operator==(iterator const& a, iterator const& b) {
-      return std::tie(a.curr_, a.it_, a.end_) ==
-             std::tie(b.curr_, b.it_, b.end_);
+      // Input iterators may only be compared to themselves and end.
```
devbww: Perhaps s/themselves/(copies of) themselves/?

google/cloud/spanner/row_parser.h
devjgm: a `break` sounds right.

google/cloud/spanner/row_parser.h
devjgm: done.

PR #173, opened 2019-07-09
Improved some doxygen commenting.

devjgm:
Changed to using @code, @par, @note, @warning, etc.

Inspected doxygen output and iterated until it looked slightly better.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/173)

### DIFF
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -183,15 +183,18 @@ class Row {
   std::tuple<Types...>&& get() && { return std::move(values_); }
   std::tuple<Types...> const&& get() const&& { return std::move(values_); }
 
-  // Equality operators.
+  /// @name Equality operators
+  ///@{
   friend bool operator==(Row const& a, Row const& b) {
     return a.values_ == b.values_;
   }
   friend bool operator!=(Row const& a, Row const& b) {
     return a.values_ != b.values_;
   }
+  ///@}
 
-  // Relational operators.
+  /// @name Relational operators
+  ///@{
   friend bool operator<(Row const& a, Row const& b) {
     return a.values_ < b.values_;
   }
@@ -204,6 +207,7 @@ class Row {
   friend bool operator>=(Row const& a, Row const& b) {
     return a.values_ >= b.values_;
   }
+  ///@}
 
  private:
   std::tuple<Types...> values_;
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -45,23 +45,26 @@ inline namespace SPANNER_CLIENT_NS {
  * function template (see below). Once an instance is created, you can access
  * the values in the row using the `Row::get()` overloaded functions.
  *
- * Example:
+ * @par Example
  *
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/row.h
```diff
@@ -156,32 +164,34 @@ class Row {
...
+  // Equality operators.
   friend bool operator==(Row const& a, Row const& b) {
     return a.values_ == b.values_;
   }
   friend bool operator!=(Row const& a, Row const& b) {
     return a.values_ != b.values_;
   }
 
-  /// Relational operators.
+  // Relational operators.
```
coryan: No comments are fine, but also this could work:

```C++
//@{
/// @name Relational operators.
friend bool operator<(...) { ... }
friend bool operator<=(...) { ... }
...
//@}
```

The operators will be grouped in the Doxygen documentation, under the "Relational operators" heading.

coryan: And if you like how that looks, consider the same for the equality operators.

google/cloud/spanner/value.h
```diff
@@ -194,10 +209,10 @@ class Value {
    * The type `T` may be any valid type shown above, except vectors of vectors
    * are not allowed.
    *
-   * NOTE: If `T` is a `std::tuple` with field names (i.e., at least one of its
-   * element types is a `std::pair<std::string, T>`) then, all of the vector's
-   * elements must have exactly the same field names. Any mismatch in in field
-   * names results in undefined behavior.
+   * @warning If `T` is a `std::tuple` with field names (i.e., at least one of
+   * its element types is a `std::pair<std::string, T>`) then, all of the
```
coryan: Should we indent these?  We have done so in the other libraries, but I know it is tedious...

google/cloud/spanner/value.h
```diff
@@ -311,27 +332,27 @@ class Value {
...
   }
 
-  /**
-   * Allows Google Test to print internal debugging information when test
-   * assertions fail.
-   *
-   * @warning This is intended for debugging and human consumption only, not
-   * machine consumption as the output format may change without notice.
-   */
+  // Allows Google Test to print internal debugging information when test
```
coryan: This one no longer has Doxygen comments, is that intentional?  I think it should have them, as the function is part of the public API (sadly that is what gtest requires).

google/cloud/spanner/value.h
coryan: Ditto for the `@note` paragraphs.

google/cloud/spanner/row.h
devjgm: Added groups for these operators.

google/cloud/spanner/value.h
devjgm: SGTM. Indented wrapped lines 2 spaces

google/cloud/spanner/value.h
devjgm: I did remove the doxygen comment on purpose, but I've added it back.

PR #174, opened 2019-07-10
feat: Use policies to control polling.

### LINKED ISSUE #126: Refactor ToGoogleCloudStatus(google::rpc::Status) to common location.
This function converts a protobuf `Status` into a `google::cloud::Status`. It should be in a common location for both `google-cloud-cpp` and `google-cloud-cpp-spanner`.

### LINKED ISSUE #128: Introduce a polling policy to control loop for long running operations.
We need a wait to automatically poll long running operations, and we need a way for applications to control this polling loop. In other libraries we used a polling policy to do so.

coryan:
This PR introduces a `spanner::PollingPolicy` to control for how long we
poll longrunning operations. We use this policy in the
`internal::DatabaseAdminRetry` decorator, and refactored the polling
loop to a function, so we can test it.

I removed a number of tests for DatabaseAdminClient because they
basically repeat the (more generic) tests for `internal::PollingLoop`.

Fixes #128 and fixes #126.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/174)

### DIFF
--- a/google/cloud/spanner/internal/database_admin_retry.cc
+++ b/google/cloud/spanner/internal/database_admin_retry.cc
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 #include "google/cloud/spanner/internal/database_admin_retry.h"
+#include "google/cloud/spanner/internal/polling_loop.h"
 #include "google/cloud/spanner/internal/retry_loop.h"
+#include "google/cloud/grpc_utils/grpc_error_delegate.h"
 
 namespace google {
 namespace cloud {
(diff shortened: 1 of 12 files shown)


google/cloud/spanner/internal/database_admin_retry.cc
```diff
@@ -54,11 +56,42 @@ std::unique_ptr<BackoffPolicy> DefaultAdminBackoffPolicy() {
       .clone();
 }
 
+#ifndef GOOGLE_CLOUD_CPP_SPANNER_ADMIN_DEFAULT_POLLING_TIMEOUT
```
devjgm: These are settings that we need to tweak at build time? How often do these change?

google/cloud/spanner/internal/polling_loop_test.cc
```diff
@@ -0,0 +1,254 @@
...
+  operation.set_done(true);
+  operation.mutable_response()->PackFrom(expected);
+
+  StatusOr<google::protobuf::Value> actual =
+      PollingLoop<google::protobuf::Value>(
+          TestPollingPolicy(),
+          [](grpc::ClientContext&,
+             google::longrunning::GetOperationRequest const&) {
+            // The polling operation should not be called for requests
+            EXPECT_TRUE(false);
```
devjgm: You may be able to use the `FAIL()` googletest macro here to make this error case more obvious when visually scanning the file.

coryan: `FAIL()` (and `GTEST_FAIL()`) have an embedded `return;` (note the semi-colon), so they do not work here.  Is there a `FAIL_WITH_RETURN()` or something?  I have never been able to find an alternative (ditto for `ASSERT_*()`).

google/cloud/spanner/internal/database_admin_retry.cc
coryan: > These are settings that we need to tweak at build time?

Our users might.

> How often do these change?

Approximately never.

I do not like libraries that hard-code things for me, but I am happy to remove this if you think it is too much.

devjgm: No change necessary; keeping these is totally fine with me.

But I'll clarify my thinking on this topic a little.

These are only user-configurable if the user is compiling the source from scratch. If they are linking against a library, these defines will be ignored and the user will get whatever settings the library maintainer compiled the library with. 

Since these are just the settings for the default policy, users can already customize the timeouts by using a non-default policy.

I'd be happy to omit these preprocessor knobs until someone needs them. But I'm also totally happy to keep them if you see them being useful.

(I imagine you may now feel pressure to remove them to appease the reviewer. Please don't :-) Really, feel free to keep them unless you agree. Either is fine with me. My approval stands as-is.)

coryan:
Ack, I am less convinced about the compile-time configuration, though I picked that habit in a former life developing libraries.  Created #177 to reconsider.

PR #175, opened 2019-07-10
feat: Implement examples for DatabaseAdminClient.

coryan:
This implements the first examples, and creates the scaffolding to run
them as part of the CI build. The examples are like integration tests
from the point of view of the CI builds. They are compiled as usual, and
executed with all the other integration tests. The examples check if the
environment variable `GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES` is set to
`yes`. If it is, then the examples run a predefine set of commands,
using a random name for the database.

The user (typically a developer) can also execute the examples manually,
as seems to be customary in the cloud.google.com site, for example see:

https://cloud.google.com/spanner/docs/getting-started/go/#create_a_database

in this case the user provides all the parameters (database name, project
id, etc.) in the command-line.

This PR also makes the sample code visible as snippets in the Doxygen
documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/175)


google/cloud/spanner/examples/database_admin_snippets.cc
```diff
@@ -0,0 +1,180 @@
...
+}
+
+bool AutoRun() {
+  return google::cloud::internal::GetEnv("GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES")
+             .value_or("") == "yes";
+}
+
+}  // namespace
+
+int main(int ac, char* av[]) try {
```
devjgm: My understanding of [function try blocks](https://en.cppreference.com/w/cpp/language/function-try-block) is that they're intended to catch exceptions from member initializers in a constructor. They do work on any function (including `main()`), but I *think* for regular functions they're equivalent to the normal form of 
```cc
int main() {
  try {
    ...
  } catch (...) {
    ...
  }
}
``` 

I have a slight preference for using the vanilla try/catch block form for functions since I think it's more typical and less surprising. Of course, we can always feel free to use a function-try-block in a c'tor where it may be needed to catch  member initialization exceptions.

Thoughts?

coryan: > They do work on any function (including `main()`), but I _think_ for regular functions they're equivalent to the normal form of

Agree.

> I have a slight preference for using the vanilla try/catch block form for functions

I have a slight preference to the function try block, because it saves on the extra indentation.

> since I think it's more typical and less surprising.

I do not find them surprising at all, specially around `main()`, but that might be just me.  As to "typical", I am not sure, in a previous life, when we used exceptions a lot (I mean, we did not *throw* them a lot, but we assumed they could happen anywhere) we wrote `main()` like this.

Seems like we have a tie on "slight preference", should we flip a coin?

coryan:
> This is excellent.

Thanks!

> So that I'm clear, these "examples" (or "samples"?)

I am terribly inconsistent about naming these, happy for somebody else to pick one.

> are what we'll write to show that each of our needed use-cases is done and working. Is that right?

I think that is the main reason to write them for the Alpha release.

We *also* want to show how to use these functions in the reference documentation (here is an [example](https://googleapis.github.io/google-cloud-cpp/latest/storage/classgoogle_1_1cloud_1_1storage_1_1v1_1_1Client.html#a985880a314d312836387ea3ed2d60598) for the storage library), and we *also* extract this examples into the cloud.google.com website (here is a an example for [Spanner C#](https://cloud.google.com/spanner/docs/getting-started/csharp/#create_a_database)).

I believe those are requirements for *Beta*, but since they are useful to convince ourselves we are "done" for Alpha I think we might as well write them now.

google/cloud/spanner/examples/database_admin_snippets.cc
devjgm: Keeping it sounds fine. Saving a little indentation is a fair point, and it's more concrete than my opinion about typical and surprising.

Maybe I'll grow to like it, like I've grown to like east const :-)

PR #197, opened 2019-07-12
feat: Add CMake install target.

remyabel:
This PR adds the CMake install target, a small program (copy/pasted from the database admin snippet), tests that the install works and that the small program that depends upon spanner compiles.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/197)


remyabel:
> I think we should start a ci/kokoro/install subdirectory, which would just Docker to test the make + make install + use-installed-files-test process. We can keep that fairly simple first, with a single distribution as a target and a really small build.sh file.

If we're going that route we may as well cache the images like in `google-cloud-cpp` unless you want to keep it simple for now? If the former I will hold off on this PR until those change are implemented.

remyabel:
I believe all your concerns have been addressed. PTAL.

PR #201, opened 2019-07-12
cleanup: Support metadata as result in PollingLoop.

coryan:
Some operations (like the upcoming `UpdateMetadata()`) use long-running
operations, but all the interesting results are in the metadata field,
not the response field. This change refactors the code to support
extracting the result from either.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/201)

### DIFF
--- a/google/cloud/spanner/internal/polling_loop.h
+++ b/google/cloud/spanner/internal/polling_loop.h
@@ -28,6 +28,65 @@ namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
 namespace internal {
+/**
+ * Extract the result of a long-running operation from the `response` field.
+ */
+template <typename ResultType>
+struct PollingLoopResponseExtractor {
+  using ReturnType = StatusOr<ResultType>;
+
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/internal/polling_loop.h
```diff
@@ -28,6 +28,65 @@ namespace cloud {
...
 namespace internal {
+/**
+ * Extract the result of a long-running operation from the `response` field.
+ */
+template <typename ResultType>
+struct PollingLoopResponseExtractor {
+  using ReturnType = StatusOr<ResultType>;
+
+  static ReturnType Extract(google::longrunning::Operation const& operation,
+                            char const* location) {
```
devjgm: What is the `location` argument, and why is it named "location"? I see it used in some Status error messages, so it looks like it could be any string. But what's the intended meaning of location?

coryan: Typically the `__func__` that started the loop, that way we report something like "error in UpdateDatabase" instead of "error in PollingLoop for mystery function".

PR #204, opened 2019-07-12
chore: Speedup install builds.

coryan:
This makes the install builds go faster when running them on your
workstation, I expect no effect for the CI environment.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/204)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+# We do not need the git repository in docker images.
+.git/
+
+# We do not need the typical build directories in docker images.
+.idea/
+.build/
+_build/
+build-output/
+cmake-build-*/
+cmake-out/
+
+# Skip these directories, the files here are not needed to build any of
+# the Docker images. Furthermore, when creating new Dockerfiles the files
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -0,0 +1,17 @@
...
+build-output/
+cmake-build-*/
+cmake-out/
+
+# Skip these directories, the files here are not needed to build any of
+# the Docker images. Furthermore, when creating new Dockerfiles the files
+# here change a lot, but do not change the image, so they trigger image rebuids
+# without any benefit.
+ci/kokoro/
+ci/test-readme/
```
remyabel: I think you mean `ci/test-install` but let's also leave `test-readme` here if that's coming in the future.

coryan: Hmmm, if we add `ci/test-install` then those files don't get uploaded to the Docker daemon, and we need them inside the Docker build for `ci/kokoro/install/Dockerfile.fedora`, right?  I removed the directory to avoid confusion.

coryan:
Ping.

Issue #198, opened 2019-07-12
Implement spanner::Mutation Skeleton


devjgm:
The spanner::Mutation class represents an [insert, update, insert_or_update, replace, or delete](https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/mutation.proto#L79-L95) operation that can be applied to a Spanner table using the spanner::Client::Commit function. This class will be a value type with the expected copy/move operations, as well as support for equality. We don't have the details of this class designed yet, but we expect to provide the obvious accessor functions, and factory functions for creating instances. The following is an outline of what this class might look like:

```cc
namespace google {
namespace cloud {
namespace spanner {

class Mutation {
 public:
  // TODO: value type, ==, !=
  // TODO: expected accessors

 private:
  // Holds a google::spanner::v1::Mutation proto
};

Mutation MakeInsertMutation(...);
Mutation MakeUpdateMutation(...);
Mutation MakeInsertOrUpdateMutation(...);
Mutation MakeReplaceMutation(...);
Mutation MakeDeleteMutation(...);

}  // namespace spanner
}  // namespace cloud
}  // namespace google
```

coryan:
For what is worth, only delete mutations are missing, but we need key sets #202 to implement those.

coryan:
I am going to close this in favor of #277 because that is a more specific issue.

PR #211, opened 2019-07-13
feat: Implement Mutations skeleton.

coryan:
This is a very bare bones skeleton, it supports only one mutation type
(insert), and with a single row at that.

Part of the work for #198

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/211)


coryan:
Hmmm... on second thought, I do not like this. Closing while I write something else.

PR #212, opened 2019-07-13
feat: Implement Mutations skeleton.

coryan:
This PR introduces the `Mutation` class, some helper functions to create
simple mutations, and a couple of builders to create more complex
mutations.

I would like to hear if the `Mutation` class is useful, it is such a thin wrapper
around the proto that it seems tempting to just use the proto directly. Protos
are semi-regular, but it is unclear that one should be comparing mutations
anyway.

This is part of the work for #198.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/212)


google/cloud/spanner/mutations.h
```diff
@@ -0,0 +1,234 @@
...
+  Mutation(Mutation const&) = default;
+  Mutation& operator=(Mutation const&) = default;
+
+  friend bool operator==(Mutation const& lhs, Mutation const& rhs);
+  friend bool operator!=(Mutation const& lhs, Mutation const& rhs) {
+    return !(lhs == rhs);
+  }
+
+  /// Convert the mutation to the underlying proto.
+  google::spanner::v1::Mutation as_proto() && { return std::move(m_); }
```
devjgm: Oooo, for classes that wrap protos, pretty much 1:1, I like this technique (defined on rvalue). :-) Thanks!

google/cloud/spanner/mutations.h
```diff
@@ -0,0 +1,234 @@
...
+    for (auto& name : column_names) {
+      *field.add_columns() = std::move(name);
+    }
+  }
+
+  google::spanner::v1::Mutation Build() const& { return m_; }
+  google::spanner::v1::Mutation Build() && { return std::move(m_); }
+
+  template <typename... Ts>
+  void AddRow(Ts&&... values) {
```
devjgm: I wonder if things where we "add rows" should accept a `Row<Ts...>` maybe instead of, or in addition to, `Value` objects? I guess it depends on the use case, what type callers may be more likely to have. For example, I could imagine that it may be nice to write this like:

```cc
builder.AddRow(MakeRow("hello", 42, "world", true));
```
As opposed to
```cc
builder.AddRow(Value("hello"), Value(42), Value("world"), Value(true));
```

However, if we go toward accepting `Row`, we may want to wait for a PR I just sent #214 because then it'll be easier to extract the `Value` objects back out of the Row.

google/cloud/spanner/mutations_test.cc
```diff
@@ -0,0 +1,204 @@
...
+  std::string delta;
+  google::protobuf::util::MessageDifferencer differencer;
+  differencer.ReportDifferencesToString(&delta);
+  EXPECT_TRUE(differencer.Compare(actual, expected)) << delta;
+}
+
+TEST(MutationsTest, InsertComplex) {
+  Mutation empty;
+  auto builder = InsertMutationBuilder({"col1", "col2", "col3"})
+                     .AddRow(std::int64_t(42), std::string("foo"), false)
```
devjgm: (related to a previous comment)

If we switched to world where the argument type is `Row<Ts...>` and callers used `MakeRow()`, they would not need to go out of their way to cast `42` as `std::int64_t(42)`, because `MakeRow` would do that implicitly.

FWIW

google/cloud/spanner/mutations.h
coryan: I think we should go with an overload that takes `Row<Ts...>`. Not all applications will have a `Row` handy, the data they want to put into a mutation may come from anywhere.

google/cloud/spanner/mutations_test.cc
coryan: I changed the test to show that `42` also works. Under the hood all I am doing is `internal::ToProto(Value(parameter))` so any conversion that works for `Value` should also work here.

coryan:
Ping.

Issue #202, opened 2019-07-12
Implement spanner::Key, spanner::KeySet


scotthart:
Implement the Key and KeySet classes.

```cc
namespace google {
namespace cloud {
namespace spanner {
class Key { 
 public:  
  template <typename... Ts>  // TODO: Add appropriate enabler
  explicit Key(Ts&&... ts);
  // Value type, op==, op!=
  // ...
};

class KeySet {
 public:
  static KeySet All() { return KeySet(all_tag{}); }
  KeySet() = default;
   // TODO: Make this a value type with op==, op!=
  // TODO: Add support for key ranges

  void Add(Key key);
};
}  // namespace spanner
}  // namespace cloud
}  // namespace google
```
Please include all reasonable boilerplate code, copyright notices, basic doxygen comments, and `.h`, `.cc`, and `_test.cc` files.

## Design notes
A spanner::KeySet is an argument to the spanner::Client::Read functions. Keys may be specified as individual keys, key ranges (with open or closed intervals), or a special "all keys" token. A "key" consists of the value(s) for the column(s) used in the named index. This is encapsulated by the spanner::Key class. 

A spanner::Key is very similar to a `std::vector<spanner::Value>`, except that a Value can be "null" while a Key cannot. We may be able to invent a more general concept that covers both needs.

devjgm:
Just something for us to consider...

Now that we have `Row<Ts...>` (https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/row.h) we may want to consider whether a `Row` is a good replacement for `Key`. I do think we probably want a `KeySet` still, but I could imagine a `KeySet` object on which we call `.AddRow(...)`. 

We'd have to think about how that plays with templates, how we convert the Row -> proto values, etc. But something to consider.

devjgm:
... more thinking....

We already have `ParseRow` (https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/row.h#L289) that converts from an `std::array<Value, N>` to a `Row<Ts...>`. We'd need the opposite function to convert from a `Row<Ts...>` to a `std::array<Value, sizeof...(Ts)>`. Let me call that function `UnparseRow()` for a moment.

If we had `UnparseRow()`, we may be able to use the type `Row<Ts...>` as the "Key" type in the KeySet, and we'd be able to convert each "row key" into the appropriate Values and protos needed to create the gRPC requests.

[...]

PR #214, opened 2019-07-14
feat: Adds Row::values() accessor that returns std::array<Value, N>

devjgm:
This accessor lets callers extract all the columns from a row without
explicitly knowing the corresponding C++ types by getting a std::array
of Value objects. Those could then be interogated to figure out their
values, or the raw protos could be extracted.

I suspect this might be useful for the upcoming KeySet implementation: #202

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/214)

### DIFF
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -183,6 +183,25 @@ class Row {
   std::tuple<Types...>&& get() && { return std::move(values_); }
   std::tuple<Types...> const&& get() const&& { return std::move(values_); }
 
+  /**
+   * Returns a std::array of `Value` objects holding all the items in this row.
+   *
+   * @par Example
+   *
+   * @code
+   * auto row = MakeRow(true, "foo", 42);
+   * std::array<Value, 3> a = row.values();
+   * assert(a[0].get<bool>() == true);
+   * assert(a[1].get<std::string>() == "foo");
+   * assert(a[2].get<std::int64_t>() == 42);
+   * @endcode
+   */
+  std::array<Value, size()> values() const {
+    std::array<Value, size()> array;
+    internal::ForEach(values_, InsertValue{array, 0});
+    return array;
+  }
+
   /// @name Equality operators
   ///@{
   friend bool operator==(Row const& a, Row const& b) {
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/row.h
```diff
@@ -183,6 +183,25 @@ class Row {
...
+   *
+   * @code
+   * auto row = MakeRow(true, "foo", 42);
+   * std::array<Value, 3> a = row.values();
+   * assert(a[0].get<bool>() == true);
+   * assert(a[1].get<std::string>() == "foo");
+   * assert(a[2].get<std::int64_t>() == 42);
+   * @endcode
+   */
+  std::array<Value, size()> values() const {
```
coryan: Can we have a `&&` overload for this function?  Otherwise we are always going to copy all the values, and they could be  large (`BYTES`).

devjgm: Done. I added an overload for `values() &&`. I did not add overloads for `&`, because I it should behave the same as `const&`. I also didn't add an overload for `const&&`, because it's a really weird overload anyway, and I think it should behave the same as `const&` too (which it will).

Also, testing move semantics is really hard (not even sure how) with types that we don't own. So I wrote a test that verifies that properly written code will do what it's supposed to do, but I do not verify that the data (i.e., the std::string data) is actually moved.

devjgm:
PTAL.

google/cloud/spanner/row.h
```diff
@@ -183,6 +183,36 @@ class Row {
...
+   * // Potentially more efficient if `row` is no longer needed.
+   * std::array<Value, 3> b = std::move(row).values();
+   * @endcode
+   */
+  std::array<Value, size()> values() const& {
+    std::array<Value, size()> array;
+    internal::ForEach(values_, InsertValue{array, 0});
+    return array;
+  }
+  std::array<Value, size()> values() && {
```
coryan: Add the usual `// @copydoc` ?

devjgm: Good point. Done.

PR #215, opened 2019-07-15
feat: Implement google::cloud::spanner::Transaction

### LINKED ISSUE #30: Implement spanner::Transaction skeleton
Implement the skeleton of a `spanner::Transaction` class, which should initially look something like the following:

```cc
namespace google {
namespace cloud {
namespace spanner {

class Transaction {
 public:
  struct ReadOnlyOptions {
    // … like strong, time bounds, etc.
  };
  struct ReadWriteOptions {
    // … currently none
  };

  // Copy and move support. No default constructor.
  Transaction(Transaction&&) = default;
  Transaction& operator=(Transaction&&) = default;
  Transaction(Transaction const&) = default;
  Transaction& operator=(Transaction const&) = default;

  // Equality comparable.
  friend bool operator==(Transaction const& a, Transaction const& b);
  friend bool operator!=(Transaction const& a, Transaction const& b);

  // No other public API.
  // No accessors.
  // A Transaction is a fairly opaque handle.

 private:
  Transaction() = default;
};

}  // namespace spanner
}  // namespace cloud
}  // namespace google
```

Note that caller's cannot construct this class directly; it's constructed by `Client` and given to callers. However, we will need a way to construct instances for testing, so I imagine some friendship or a "spanner_internal" factory function may need to exist.

## Design notes for `spanner::Transaction`

Spanner transactions are represented by an instance of the `spanner::Transaction` class. Caller's cannot directly construct Transaction objects, rather Transaction objects are created by the Client class and passed as arguments to caller-provided function objects that are passed to a spanner::Client::RunTransaction method.

devbww:
A transaction is a set of reads and writes that execute atomically at a
single logical point in time across the columns/rows/tables in a database.
A `google::cloud::spanner::Transaction` is passed to each of those read and
write operations, tying them together.

Transaction is opaque to the user, but `google::cloud::spanner::Client` is
granted the friendship needed to implement its `Read()`and `ExecuteSql()`
calls.

Fixes #30.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/215)


google/cloud/spanner/internal/transaction_test.cc
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  // User-visible read operation.
+  ResultSet Read(Transaction txn, std::string const& table, KeySet const& keys,
+                 std::vector<std::string> const& columns) {
+    std::function<ResultSet(TransactionSelector&)> read =
+        [this, &table, &keys, &columns](TransactionSelector& selector) {
+          return this->Read(selector, table, keys, columns);
+        };
+    return txn->Visit(std::move(read));
```
devjgm: It looks like access to `Transaction::operator->` and `Transaction::Visit` are the reason that Transaction needs to friend Client. One alternative idea would be to have Transaction friend its own *internal* function instead, and then it would not need to friend `Client`. Maybe something like:

```cc
// transaction.h
namespace internal {
template <typename T>
T Visit(Transaction txn, std::function(T(TransactionSelector&)> f);
}  // namespace internal

class Transaction {
 private:
  template <typename T>
  friend T Visit(........);
};

// Then in this file (same as in client.cc, the unit tests, etc.)
...
  return internal::Visit(txn, std::move(read));
...
```

This would mean that Transaction's friends are all "close friends", meaning that its only friends are classes/functions that live in transaction.h (or a related file that it owns). This would also minimize the amount of faking you need to do in unit tests (e.g., no need to fake a Client class).

Thoughts?

google/cloud/spanner/internal/transaction_test.cc
```diff
@@ -0,0 +1,156 @@
...
+  int valid_visits_;  // GUARDED_BY(mu_);
+};
+
+// Transaction callback.  Normally we would use the TransactionSelector
+// to make a StreamingRead() RPC, and then, if the selector was a `begin`,
+// switch the selector to use the allocated transaction ID.  Here we use
+// the pre-assigned transaction ID after checking the read timestamp.
+ResultSet Client::Read(TransactionSelector& selector, std::string const&,
+                       KeySet const&, std::vector<std::string> const&) {
+  if (selector.has_begin()) {
```
coryan: Should we collapse the nested `if()` block to a single boolean expression?

google/cloud/spanner/transaction.h
```diff
@@ -0,0 +1,156 @@
...
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+namespace internal {
+class Transaction;
+}
+
+/**
+ * A transaction is a set of reads and writes that execute atomically at a
```
coryan: This is a bit long for the "brief" description... maybe: "A wrapper for Cloud Spanner transactions. Ties read and write operations together."

google/cloud/spanner/transaction.h
```diff
@@ -0,0 +1,156 @@
...
+    SingleUseOptions(std::chrono::nanoseconds max_staleness);
+
+   private:
+    friend Transaction;
+    google::spanner::v1::TransactionOptions_ReadOnly ro_opts_;
+  };
+
+  /// @name Construction of read-only and read-write transactions.
+  ///@{
+  Transaction(ReadOnlyOptions opts);
```
coryan: How do these constructors connect a transaction to a session and client?  Or do they create a transaction selector, but the transaction is not really created yet?

google/cloud/spanner/transaction.h
```diff
@@ -0,0 +1,156 @@
...
+inline namespace SPANNER_CLIENT_NS {
+
+namespace internal {
+class Transaction;
+}
+
+/**
+ * A transaction is a set of reads and writes that execute atomically at a
+ * single logical point in time across the columns/rows/tables in a database.
+ * A google::cloud::spanner::Transaction is passed to each of those read and
```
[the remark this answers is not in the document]
devbww: Done (backticks and unqualified).  Also tweaked to shorten the "short" part, as suggested by @coryan.

google/cloud/spanner/transaction.h
```diff
@@ -0,0 +1,156 @@
...
+ * All reads/writes in the transaction must be executed within the same
+ * session, and that session may have only one transaction active at a time.
+ *
+ * Spanner supports these transaction modes:
+ *   - ReadOnly. Provides guaranteed consistency across several reads, but does
+ *     not allow writes. Can be configured to read at timestamps in the past.
+ *     Does not need to be committed and does not take locks.
+ *   - ReadWrite. Supports reading and writing data at a single point in time.
+ *     Uses pessimistic locking and, if necessary, two-phase commit. May abort,
+ *     requiring the application to retry.
```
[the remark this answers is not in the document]
devbww: Done.

google/cloud/spanner/transaction.h
```diff
@@ -0,0 +1,156 @@
...
+    // committed before the start of the reads.
+    ReadOnlyOptions();
+
+    // Exact Staleness: Executes all reads at `read_timestamp`.
+    ReadOnlyOptions(Timestamp read_timestamp);
+
+    // Exact Staleness: Executes all reads at a timestamp `exact_staleness`
+    // old. The actual timestamp is chosen soon after the reads are started.
+    ReadOnlyOptions(std::chrono::nanoseconds exact_staleness);
+
```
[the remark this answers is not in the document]
devbww: I had thought about this and concluded that (1) providing such an accessor interface was tantamount to reimplementing the proto, but (2) that [I think] no one will really need it.

google/cloud/spanner/transaction.h
```diff
@@ -0,0 +1,156 @@
...
+ *   - ReadWrite. Supports reading and writing data at a single point in time.
+ *     Uses pessimistic locking and, if necessary, two-phase commit. May abort,
+ *     requiring the application to retry.
+ */
+class Transaction {
+ public:
+  /**
+   * Options for ReadOnly transactions.
+   */
+  class ReadOnlyOptions {
```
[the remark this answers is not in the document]
devbww: Similar to the above "no accessor" reasoning, it seems that these option values are short-lived ways to pass values to `Transaction` and so do not need to be rich.

google/cloud/spanner/transaction.h
```diff
@@ -0,0 +1,156 @@
...
+  }
+  friend bool operator!=(Transaction const& a, Transaction const& b) {
+    return !(a == b);
+  }
+  ///@}
+
+ private:
+  friend class Client;
+
+  /// Construction of single-use transaction.
```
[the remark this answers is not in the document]
devbww: Done.

google/cloud/spanner/transaction_test.cc
```diff
@@ -0,0 +1,71 @@
...
+    return Transaction(std::move(opts));
+  }
+};
+
+namespace {
+
+TEST(Transaction, RegularSemantics) {
+  Transaction::ReadOnlyOptions strong;
+  Transaction a(strong);
+  Transaction b(strong);
```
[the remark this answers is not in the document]
devbww: Re `A a = b;` ... I believe I already have `Transaction c = b;` a couple of lines down.

Re `A a; a = b;` ... Added new expectations after `c = a;`.

google/cloud/spanner/internal/transaction_test.cc
devbww: Done.

google/cloud/spanner/transaction.h
devbww: Done.

google/cloud/spanner/transaction.h
devbww: Right ... the "transaction-selector to transaction-id" transformation is not made until the `Transaction` is used in the context of a particular client/session.  Then it is up to the client to only use the `Transaction` in that context.

devbww:
> are transactions just "transaction selectors" that create the actual transaction on demand? Or can they represent an actual transaction, by which I mean the result of calling the BeginTransaction() rpc?

I'll say both.  They start off as "transaction selectors", but as soon as that selector is used in an RPC, `BeginTransaction()` or anything else, they become bound to the allocated "tranaction id".

We have an option open as to whether `Client::BeginTransaction()` will exist and the user can call that if they want to bind early, or if the `Transaction` might be marked as "eager" where `Client` would do the `BeginTransaction` RPC for you on first use.

google/cloud/spanner/transaction.h
coryan: Ack.

google/cloud/spanner/transaction_test.cc
```diff
@@ -0,0 +1,71 @@
...
+      Transaction::SingleUseOptions opts) {
+    return Transaction(std::move(opts));
+  }
+};
+
+namespace {
+
+TEST(Transaction, RegularSemantics) {
+  Transaction::ReadOnlyOptions strong;
+  Transaction a(strong);
```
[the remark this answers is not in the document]
devbww: Added `MakeReadOnlyTransaction()`, `MakeReadWriteTransaction()`, and `internal::MakeSingleUseTransaction()` factories.  Read-only and read-write `Transaction` constructors are still visible.

google/cloud/spanner/internal/transaction_test.cc
devbww: Done.  Also renamed `internal::Transaction` to `internal::TransactionImpl` to avoid ambiguity within `internal`.

devbww:
@devjgm, PTAL.  Thanks.

google/cloud/spanner/internal/transaction_impl_test.cc
```diff
@@ -0,0 +1,202 @@
...
+inline namespace SPANNER_CLIENT_NS {
+
+using TransactionSelector = google::spanner::v1::TransactionSelector;
+
+class KeySet {};
+class ResultSet {};
+
+// A fake `google::cloud::spanner::Client`, supporting a single `Read()`
+// operation that does nothing but track the expected transaction callbacks.
+class Client {
```
devjgm: To avoid ODR conflicts w/ the *real* `google::cloud::spanner::Client` class, what do you think about putting this one (and KeySet and ResultSet) in an unnamed namespace.

devbww: Done.  Thanks.

PR #216, opened 2019-07-15
Add equality operators to SqlStatement class

### LINKED ISSUE #107: Consider adding op== to SqlStatement
In #103 we added `SqlStatement`, but didn't include support for equality. This makes it a "semiregular" type. In general, we should prefer regular to semiregular types whenever possible. See [this item from the C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c11-make-concrete-types-regular).

We probably want to add equality support to SqlStatement, but we need to decide how to handle the case sensitivity of the sql string. Spanner's [case sensitivity docs are here](https://cloud.google.com/spanner/docs/lexical#case-sensitivity). Some options for `SqlStatement` that I see are:

0. Do nothing, and do not support equality of `SqlStatement`. This leaves the class semiregular.

1. 100% **case sensitive**. That is "select Foo" != "select foo". This is simple to implement in C++ and it's not unexpected behavior (case sensitivity is typical). However, this behavior differs from what [Spanner does on the server side](https://cloud.google.com/spanner/docs/lexical#case-sensitivity).

2. 100% case **insensitive**. That is "select Foo" == "SeLEcT FoO". This is also easy to implement in C++, but this behavior also differs from what Spanner does on the server side.

3. Parse the SQL string, and do equality exactly as Spanner does it on the server side. This would be harder to implement in C++, more difficult to keep *exactly* in sync w/ Spanner on the server side, but it would match what spanner does on the server.

---

FWIW,  I lean toward option 1. I think the only downside is that the behavior in C++ differs from the behavior on the Spanner servers. I wouldn't expect this to be an actual problem in practice. Additionally, there's precedent in other cases to do things differently in the client library than spanner does on the server. For example, spanner's STRUCT objects on the server *ignore* field names when doing equality comparisons, whereas our `Value` class (and the Value class for all the other client libraries) *include* the field name when doing equality comparison.

tmatsuo:
Fixes #107

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/216)

### DIFF
--- a/google/cloud/spanner/sql_statement.h
+++ b/google/cloud/spanner/sql_statement.h
@@ -67,6 +67,13 @@ class SqlStatement {
   google::cloud::StatusOr<Value> GetParameter(
       std::string const& parameter_name) const;
 
+  friend bool operator==(SqlStatement const& a, SqlStatement const& b) {
+    return a.statement_ == b.statement_ && a.params_ == b.params_;
+  }
+  friend bool operator!=(SqlStatement const& a, SqlStatement const& b) {
+    return !(a == b);
+  };
+
  private:
   friend std::ostream& operator<<(std::ostream& os, SqlStatement const& stmt);
 
--- a/google/cloud/spanner/sql_statement.h
+++ b/google/cloud/spanner/sql_statement.h
@@ -72,7 +72,7 @@ class SqlStatement {
   }
   friend bool operator!=(SqlStatement const& a, SqlStatement const& b) {
     return !(a == b);
-  };
+  }
 
  private:
   friend std::ostream& operator<<(std::ostream& os, SqlStatement const& stmt);
--- a/google/cloud/spanner/sql_statement.h
+++ b/google/cloud/spanner/sql_statement.h
@@ -31,6 +31,7 @@ inline namespace SPANNER_CLIENT_NS {
  * found here: <a
  * href="https://cloud.google.com/spanner/docs/lexical#case-sensitivity">Case
  * Sensitivity</a>.
+ *
  * @note `SqlStatement` supports case-sensitive equality comparisons.
  *
  * Parameter placeholders are specified by `@<param name>` in the SQL string.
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/sql_statement.h
```diff
@@ -67,6 +67,13 @@ class SqlStatement {
   google::cloud::StatusOr<Value> GetParameter(
       std::string const& parameter_name) const;
 
+  friend bool operator==(SqlStatement const& a, SqlStatement const& b) {
+    return a.statement_ == b.statement_ && a.params_ == b.params_;
+  }
+  friend bool operator!=(SqlStatement const& a, SqlStatement const& b) {
+    return !(a == b);
+  };
```
coryan: nit: extra `;` at the end of the function definition.

tmatsuo: Thanks! Removed

google/cloud/spanner/sql_statement.h
```diff
@@ -31,6 +31,7 @@ inline namespace SPANNER_CLIENT_NS {
  * found here: <a
  * href="https://cloud.google.com/spanner/docs/lexical#case-sensitivity">Case
  * Sensitivity</a>.
+ * @note `SqlStatement` supports case-sensitive equality comparisons.
```
devjgm: Nit: I'd put a blank comment line before this `@note`

tmatsuo: Thanks, done. I'll merge once the tests pass

devjgm:
We talked with Scott offline about this PR. But I added him to this PR anyway to make sure he saw it.

tmatsuo:
Alright, I'll wait for LGTM from @scotthart then!

tmatsuo:
@scotthart Thanks! Merging

PR #221, opened 2019-07-16
feat: implement `ResultSet`.

### LINKED ISSUE #65: Implement spanner::ResultSet skeleton
Implement a skeleton of the spanner::ResultSet class. The initial outline should look somewhat like the following:

```cc
namespace google {
namespace cloud {
namespace spanner {

class ResultSet {
 public:
  ResultSet() = default;
  // Regular type with copy, move, equality, etc.
 
  template <typename... Ts>
  RowRange<Ts...> rows() const

  // Can only be called after consuming the whole stream.
  google::cloud::optional<ResultStats> stats() const;

  // TODO: Add getter for the optional transaction timestamp.
};

}  // namespace spanner
}  // namespace cloud
}  // namespace google
```

## Design notes

This class represents the results of a spanner::Client::Read and spanner::Client::ExecuteSql function call. It gives callers access to the result's columns and rows, as well as some ancillary optional data, such as stats. 

Callers can iterate the rows in the results by getting an iterable object after calling the templated `rows<Ts...>()` member function. This will return a `spanner::RowRange<Ts..>` that will parse each row into the specified types. It is the caller's responsibility to make sure that the specified types match the results that will come back from the server. Otherwise, iteration will fail with a `Status` that the user can print.

mr-salty:
Initial implementation for `ResultSet`.

Fixes #65.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/221)

### DIFF
--- a/google/cloud/spanner/result_set.h
+++ b/google/cloud/spanner/result_set.h
@@ -45,7 +45,7 @@ class ResultSetValueSource {
  * DML operations, i.e., `UPDATE` and `DELETE` also return a `ResultSet`.
  *
  * Note that a `ResultSet` returns both the data for the operation, as a
- * single-pass, input range returned by `rows()`, as well as the metadata for
+ * single-pass, input range returned by `Rows()`, as well as the metadata for
  * the results, and execution statistics (if requested).
  */
 class ResultSet {
@@ -59,7 +59,12 @@ class ResultSet {
   ResultSet& operator=(ResultSet&&) = default;
 
   /**
-   * Return a `RowParser` which can be used to iterate through the results.
+   * Returns a `RowParser` which can be used to iterate the returned `Row`s.
+   *
+   * It is possible to call this function more than once on the same object,
+   * but since the underlying data stream can only be iterated over once, this
+   * must be done with extreme caution. Using the returned `RowParser`s from
+   * multiple threads may result in errors or data corruption.
    */
   template <typename... Ts>
   RowParser<Ts...> Rows() {
@@ -86,7 +91,7 @@ class ResultSet {
    *
    * Statistics are only available for SQL requests with `query_mode` set to
    * `PROFILE`, and only after consuming the entire result stream (i.e.
-   * iterating through `rows()` until the end).
+   * iterating through `Rows()` until the end).
    *
    * TODO(#217) Determine what type we should return from this method.
    */
--- a/google/cloud/spanner/result_set.h
+++ b/google/cloud/spanner/result_set.h
@@ -27,9 +27,9 @@ namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
 
 namespace internal {
-class ResultSetValueSource {
+class ResultSetSource {
  public:
-  virtual ~ResultSetValueSource() = default;
+  virtual ~ResultSetSource() = default;
   // Returns OK Status with no Value to indicate end-of-stream.
   virtual StatusOr<optional<Value>> NextValue() = 0;
   virtual optional<google::spanner::v1::ResultSetMetadata> Metadata() = 0;
@@ -51,7 +51,7 @@ class ResultSetValueSource {
 class ResultSet {
  public:
   ResultSet() = default;
-  explicit ResultSet(std::unique_ptr<internal::ResultSetValueSource> source)
+  explicit ResultSet(std::unique_ptr<internal::ResultSetSource> source)
       : source_(std::move(source)) {}
 
   // This class is moveable but not copyable.
@@ -61,10 +61,10 @@ class ResultSet {
   /**
    * Returns a `RowParser` which can be used to iterate the returned `Row`s.
    *
-   * It is possible to call this function more than once on the same object,
-   * but since the underlying data stream can only be iterated over once, this
-   * must be done with extreme caution. Using the returned `RowParser`s from
-   * multiple threads may result in errors or data corruption.
+   * Since there is a single result stream for each `ResultSet` instance, users
+   * should not use multiple `RowParser`s from the same `ResultSet` at the same
+   * time. Doing so is not thread safe, and may result in errors or data
+   * corruption.
    */
   template <typename... Ts>
   RowParser<Ts...> Rows() {
@@ -100,7 +100,7 @@ class ResultSet {
   }
 
  private:
-  std::unique_ptr<internal::ResultSetValueSource> source_;
+  std::unique_ptr<internal::ResultSetSource> source_;
 };
 
 }  // namespace SPANNER_CLIENT_NS
--- a/google/cloud/spanner/result_set.h
+++ b/google/cloud/spanner/result_set.h
@@ -1,0 +1,99 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/result_set_test.cc
```diff
@@ -0,0 +1,185 @@
...
+        break;
+
+      case 1:
+        EXPECT_EQ(row->get<0>(), 10);
+        EXPECT_EQ(row->get<1>(), false);
+        EXPECT_EQ(row->get<2>(), "bar");
+        break;
+
+      default:
+        EXPECT_TRUE(false);
```
coryan: `FAIL()` maybe?

google/cloud/spanner/result_set_test.cc
```diff
@@ -0,0 +1,185 @@
...
+    switch (num_rows++) {
+      case 0:
+        EXPECT_TRUE(row.ok());
+        EXPECT_EQ(row->get<0>(), 5);
+        EXPECT_EQ(row->get<1>(), true);
+        EXPECT_EQ(row->get<2>(), "foo");
+        break;
+
+      case 1:
+        EXPECT_FALSE(row.ok());
```
coryan: Do you want to make sure the error type is preserved?

google/cloud/spanner/result_set.h
```diff
@@ -0,0 +1,99 @@
...
+  optional<Timestamp> read_timestamp() const {
+    auto metadata = source_->GetMetadata();
+    if (metadata.has_value() && metadata->has_transaction() &&
+        metadata->transaction().has_read_timestamp()) {
+      return internal::FromProto(metadata->transaction().read_timestamp());
+    }
+    return optional<Timestamp>();
+  }
+
+  /**
```
coryan: An alternative would have been to *not* implement the function at all.

google/cloud/spanner/result_set.h
```diff
@@ -0,0 +1,99 @@
...
+        [this]() mutable { return source_->GetNextValue(); });
+  }
+
+  /**
+   * Retrieve the timestamp at which the read occurred.
+   *
+   * Only available if a read-only transaction was used, and the timestamp
+   * was requested by setting `return_read_timestamp` true.
+   */
+  optional<Timestamp> read_timestamp() const {
```
coryan: I think the lovely naming conventions from the Google Style Guide suggest this should be `ReadTimestamp()`.

google/cloud/spanner/result_set.h
```diff
@@ -0,0 +1,99 @@
...
+
+  // This class is moveable but not copyable.
+  ResultSet(ResultSet&&) = default;
+  ResultSet& operator=(ResultSet&&) = default;
+
+  /**
+   * Return a `RowParser` which can be used to iterate through the results.
+   */
+  template <typename... Ts>
+  RowParser<Ts...> rows() {
```
coryan: Can we have multiple `rows()` iterating over the same `ResultSet`?  If not, maybe this should be a `&&` overload?

We should also consider using a `shared_ptr<>` for the `source_` member (without changing the constructor) so the `RowParser` functor and the `ResultSet` can share the source and have independent lifetimes.

google/cloud/spanner/result_set.h
```diff
@@ -0,0 +1,99 @@
...
+  virtual ~ResultSetValueSource() = default;
+  // Returns OK Status with no Value to indicate end-of-stream.
+  virtual StatusOr<optional<Value>> GetNextValue() = 0;
+  virtual optional<google::spanner::v1::ResultSetMetadata> GetMetadata() = 0;
+  virtual optional<google::spanner::v1::ResultSetStats> GetStats() = 0;
+};
+}  // namespace internal
+
+/**
+ * ResultSet is returned from "query" operations (`Read`, `ExecuteSql`).
```
coryan: Maybe:

```C++
/**
 * Represents the result of a query operation, such as `SpannerClient::Read()` and `SpannerClient::ExecuteSql()`.
 *
 * This class encapsulates the result of a Cloud Spanner query, including all DML operations, i.e., `UPDATE`
 *  and `DELETE` also return a `ResultSet`. Note that a `ResultSet` returns both the data for the operation,
 * as a single-pass, input range returned by `rows()`, as well as the metadata for the results.
 */
```

google/cloud/spanner/result_set.h
devjgm: After thinking about this some, I think I'm now of the opinion that a clear comment/warning on this function may be the  best way forward. Here's my thinking.

1. There's nothing inherently wrong with calling `rows()` multiple times. A caller could call `rs.rows<...>()`, iterate and consume the first, say, 10 rows, then do something else, then call `rs.rows<...>()` again, and iterate the next 10, and so on. Of course, the caller *could* also do this by saving the first `RowParser` that was returned, but I'm trying to point out that there's nothing fundamentally wrong with calling `rows()` multiple times. I think this would naturally work.

2. The *problem* would arise if a user got multiple `RowParsers` (either from multiple calls to `rows()` or by copying a `RowParser`), and expected that iterating both would produce consistent results. Clearly, that would not work because they're all consuming a single stream of incoming values. But I don't think it's really a problem with `RowParser` nor `ResultSet`, not `ResultSet::rows()`. I think it's just an "invalid usage" problem.

3. Of course, we want to try to make our APIs easy to use correctly and hard to use incorrectly, but I think this might be a case where clear documentation about what users *should* and *should not* do is reasonable.

But I could easily be convinced of some other option :-)

So, uh, thoughts?

google/cloud/spanner/result_set.h
```diff
@@ -0,0 +1,111 @@
...
+#include "google/cloud/optional.h"
+#include <google/spanner/v1/spanner.pb.h>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+namespace internal {
+class ResultSetValueSource {
```
devjgm: (optional) This is more than a "ValueSource", so maybe removing ValueSource from the name would be good. Other names that might make sense "ResultSetImpl" (boring, but a classic), "ResultSetSource" (kinda meaningful).

google/cloud/spanner/result_set.h
```diff
@@ -0,0 +1,111 @@
...
+      : source_(std::move(source)) {}
+
+  // This class is moveable but not copyable.
+  ResultSet(ResultSet&&) = default;
+  ResultSet& operator=(ResultSet&&) = default;
+
+  /**
+   * Returns a `RowParser` which can be used to iterate the returned `Row`s.
+   *
+   * It is possible to call this function more than once on the same object,
```
devjgm: Nit: I wonder if we could make it clearer that using multiple RowParsers on the same ResultSet is not a good idea.

What do you think about saying something like:

"Since there is a single result stream for each `ResultSet` instance, users should not use multiple `RowParsers` from the same `ResultSet` at the same time. Doing so is not thread safe, and may result in errors or data corruption."

google/cloud/spanner/result_set_test.cc
```diff
@@ -0,0 +1,190 @@
...
+  // NOLINTNEXTLINE(misc-non-private-member-variables-in-classes)
+  MOCK_METHOD0(Stats, optional<spanner_proto::ResultSetStats>());
+};
+
+TEST(ResultSet, IterateNoRows) {
+  auto mock_source_unique = make_unique<MockResultSetValueSource>();
+  auto* mock_source = mock_source_unique.get();
+  ResultSet result_set(std::move(mock_source_unique));
+
+  EXPECT_CALL(*mock_source, NextValue()).WillOnce(Return(optional<Value>()));
```
devjgm: What about structuring this like:

```cc
auto mock_source = make_unique<MockResultSetValueSource>();
EXPECT_CALL(*mock_source, NextValue()).WillOnce(Return(optional<Value>()));

ResultSet result_set(std::move(mock_source));
...
```

That makes the separate raw pointer unnecessary.

PR #223, opened 2019-07-16
feat: Implement InsertOrUpdate and Replace mutations.

coryan:
This is part for the work for #198.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/223)

### DIFF
--- a/google/cloud/spanner/mutations.h
+++ b/google/cloud/spanner/mutations.h
@@ -173,7 +173,7 @@ struct ReplaceOp {
 using InsertMutationBuilder =
     internal::WriteMutationBuilder<internal::InsertOp>;
 
-/// Creates a simple insert mutation for the values in @p row.
+/// Creates a simple insert mutation for the values in @p values.
 template <typename... Ts>
 Mutation MakeInsertMutation(Ts&&... values) {
   return InsertMutationBuilder().AddRow(std::forward<Ts>(values)...).Build();
@@ -191,7 +191,7 @@ Mutation MakeInsertMutation(Ts&&... values) {
 using UpdateMutationBuilder =
     internal::WriteMutationBuilder<internal::UpdateOp>;
 
-/// Creates a simple update mutation for the values in @p row.
+/// Creates a simple update mutation for the values in @p values.
 template <typename... Ts>
 Mutation MakeUpdateMutation(Ts&&... values) {
   return UpdateMutationBuilder().AddRow(std::forward<Ts>(values)...).Build();
@@ -209,7 +209,7 @@ Mutation MakeUpdateMutation(Ts&&... values) {
 using InsertOrUpdateMutationBuilder =
     internal::WriteMutationBuilder<internal::InsertOrUpdateOp>;
 
-/// Creates a simple "insert or update" mutation for the values in @p row.
+/// Creates a simple "insert or update" mutation for the values in @p values.
 template <typename... Ts>
 Mutation MakeInsertOrUpdateMutation(Ts&&... values) {
   return InsertOrUpdateMutationBuilder()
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/mutations.h
```diff
@@ -193,12 +191,50 @@ Mutation MakeInsertMutation(Ts&&... values) {
...
+ * @see The Mutation class documentation for an overview of the Cloud Spanner
+ *   mutation API
+ *
+ * @see https://cloud.google.com/spanner/docs/modify-mutation-api
+ *   for more information about the Cloud Spanner mutation API.
+ */
+using InsertOrUpdateMutationBuilder =
+    internal::WriteMutationBuilder<internal::InsertOrUpdateOp>;
+
+/// Creates a simple "insert or update" mutation for the values in @p row.
```
devjgm: `@p row` is not the function's param. Maybe `@p values` ?

google/cloud/spanner/mutations.h
```diff
@@ -193,12 +191,50 @@ Mutation MakeInsertMutation(Ts&&... values) {
...
+template <typename... Ts>
+Mutation MakeInsertOrUpdateMutation(Ts&&... values) {
```
devjgm: Do the variadic arguments together represent something here? That is, is the intent that the values are taken together to represent a "row"? If so, I'm not sure if it's clearer to allow callers to specify separate arguments with the intent of "please create a row for me", or if we should just have callers explicitly pass a `Row<Ts...>`.

I think we're talking about the difference between the following:

```cc
Mutation m1 = MakeInsertOrUpdateMutation(42, "hello", true);
// vs...
Mutation m2 = MakeInsertOrUpdateMutation(MakeRow(42, "hello", true));
``` 

Yes, the second one is more characters, but might its explicitness be a benefit? Other than fewer characters, does the first form (m1) offer any advantages that m2 doesn't offer?

google/cloud/spanner/mutations.h
```diff
@@ -193,12 +191,50 @@ Mutation MakeInsertMutation(Ts&&... values) {
...
+ * @see The Mutation class documentation for an overview of the Cloud Spanner
+ *   mutation API
+ *
+ * @see https://cloud.google.com/spanner/docs/modify-mutation-api
+ *   for more information about the Cloud Spanner mutation API.
+ */
+using ReplaceMutationBuilder =
+    internal::WriteMutationBuilder<internal::ReplaceOp>;
+
+/// Creates a simple "replace" mutation for the values in @p row.
```
devjgm: The argument name is `values`. But maybe I'm confused about doxygen's `@p` directive.

coryan: Fixed.

google/cloud/spanner/mutations.h
coryan: Fixed here and elsewhere.

google/cloud/spanner/mutations.h
coryan: There is an open bug to actually add the `Row<Ts>` overload, so we are basically arguing to remove this overload in favor of *only* supporting `Row<Ts>`.

> Do the variadic arguments together represent something here?

The dumb answer is "the values to be update / inserted / modified", but see below...

> That is, is the intent that the values are taken together to represent a "row"? I

The values may not need to be all the columns in the row. That is, you can update just a few columns, or insert a row with just a subset of the columns.  Though I guess the relational-theory persons will tell us that a tuple *is* a row for some projection of the table :man_shrugging: 


> Other than fewer characters, does the first form (m1) offer any advantages that m2 doesn't offer?

- Ease of use is a big thing for me. If I have `42`, `"hello"`, and `true` which I just got from my fancy computation, why do I need to wrap it into a `Row`?  Why can't the library do that for me?We are writing this library so it is easy to use the service.
- Presumably there would be more calls to move constructors with `m2` also, but that might be approximately trivial.

PR #226, opened 2019-07-16
fix: RowParser accepts shared_ptr<ValueSource>

devjgm:
`RowParser<Ts...>` now accepts the ValueSource functor wrapped in a shared_ptr. This allows the ResultSet to construct a single shared ValueSource that it shares with RowParser. 

A few fallouts from this change:

1. RowParser and ResultSet will now have independent lifetimes.

2. If a RowParser is moved to another RowParser, the second one will continue where the first left off. 

Related to #221

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/226)

### DIFF
--- a/google/cloud/spanner/row_parser.h
+++ b/google/cloud/spanner/row_parser.h
@@ -159,13 +159,11 @@ class RowParser {
   explicit RowParser(std::shared_ptr<ValueSource> vs)
       : value_source_(std::move(vs)) {}
 
-  /// @name Move-only type
-  ///@{
-  RowParser(RowParser const&) = delete;
-  RowParser& operator=(RowParser const&) = delete;
-  RowParser(RowParser&& rp) = default;
+  // Copy and assignable.
+  RowParser(RowParser const&) = default;
+  RowParser& operator=(RowParser const&) = default;
+  RowParser(RowParser&&) = default;
   RowParser& operator=(RowParser&&) = default;
-  ///@}
 
   /// Returns the begin iterator.
   iterator begin() { return iterator(value_source_); }
--- a/google/cloud/spanner/row_parser.h
+++ b/google/cloud/spanner/row_parser.h
@@ -22,6 +22,7 @@
 #include "google/cloud/status_or.h"
 #include <array>
 #include <functional>
+#include <memory>
 #include <tuple>
 #include <utility>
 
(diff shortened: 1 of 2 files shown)


mr-salty:
If we're going to allow people to call ValueSource::rows() multiple times then we might as well also allow copying RowParser.

I'm still not convinced either of those things are useful. Calling rows() multiple times to get two objects that interfere with each other under the covers seems way more surprising and easy to misuse. Disallowing multiple calls to rows() (and copying RowParser) is more straightforward.

The PR description says "This allows the ResultSet to construct a single shared ValueSource that it shares with ValueSource." - should it be "shares with RowParser'"?

devjgm:
> If we're going to allow people to call ValueSource::rows() multiple times then we might as well also allow copying RowParser.

Maybe. I agree that there's no technical reason to that we need to prevent copying RowParser. It's just a case where if a user *did* copy they'd almost certainly end up with a bug and it almost certainly wouldn't be what they wanted. Forcing them to spell `std::move` seemed like it may be helpful.

That said, if you'd rather I keep RowParser's copy semantics, that's fine with me. Just say the word and I'll add it back.

> 
> I'm still not convinced either of those things are useful. Calling rows() multiple times to get two objects that interfere with each other under the covers seems way more surprising and easy to misuse. 

Totally agree. That also would likely be a bug.

> Disallowing multiple calls to rows() (and copying RowParser) is more straightforward.

Disallowing multiple calls to row() how? Rvalue qualify it? That may work. Although, that may have other oddities. Like, after the caller calls `std::move(result_set).rows<....>()`, they may *still* need to hold on to `result_set` and call other methods on it (e.g., `.stats()`). It seems a little strange to ask users to continue using an object that was "moved"[1]. For example:

```cc
ResultSet rs = client.Read(...);
auto rows = std::move(rs).rows<T, U, V>();  // rs "moved" (see [1])
...
auto stats = rs.stats();  // rs used again
```

[1]: I realize nothing was *actually* moved, it was just `std::move`d, which is just a cast to rvalue reference. But it may still look concerning to some users. I think I've only seen the `std::move(x).F()` pattern used when the rvalue-qualified `F()` actually consumes all of `x` and it should not be used after the "move".


> 
> The PR description says "This allows the ResultSet to construct a single shared ValueSource that it shares with ValueSource." - should it be "shares with RowParser'"?

Oops. My mistakes. Yes, I'll fix the description above.

devjgm:
OK, I removed the move-only change from this PR and I updated the description above. Now this PR is only about making the ValueSource a shared_ptr.

We can discuss whether or not to make RowParser move-only later and (maybe) do it in a separate PR.

PTAL @mr-salty

Issue #188, opened 2019-07-11
Sample: Developers can query with STRUCT as bound SQL param


devjgm:
Sample code: https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/a129e6228fb4a6cb660832168b8f02ac0675673f/spanner/api/Spanner/Program.cs#L2610
Context: https://cloud.google.com/spanner/docs/structs

coryan:
This is the `spanner_query_data_with_struct` sample. Note that it might have moved from the location pointed at above.

PR #229, opened 2019-07-16
Add a sample for creating a struct with data

### LINKED ISSUE #187: Sample: Developers can create STRUCT with data
Sample code: https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/a129e6228fb4a6cb660832168b8f02ac0675673f/spanner/api/Spanner/Program.cs#L2602
Context: https://cloud.google.com/spanner/docs/structs

tmatsuo:
Fixes #187 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/229)

### DIFF
--- a/google/cloud/spanner/database_admin_client.h
+++ b/google/cloud/spanner/database_admin_client.h
@@ -126,7 +126,7 @@ class DatabaseAdminClient {
    *   completes on the service. Note that this can take minutes in some cases.
    *
    * @par Example
-   * @snippet database_admin_samples.cc create-database
+   * @snippet samples.cc create-database
    *
    * @see https://cloud.google.com/spanner/docs/data-definition-language for a
    *     full list of the DDL operations
@@ -151,7 +151,7 @@ class DatabaseAdminClient {
    *   the statements.
    *
    * @par Example
-   * @snippet database_admin_samples.cc update-database
+   * @snippet samples.cc update-database
    *
    * @see https://cloud.google.com/spanner/docs/data-definition-language for a
    *     full list of the DDL operations
@@ -169,7 +169,7 @@ class DatabaseAdminClient {
    *   database. This is an unrecoverable operation.
    *
    * @par Example
-   * @snippet database_admin_samples.cc drop-database
+   * @snippet samples.cc drop-database
    */
   Status DropDatabase(std::string const& project_id,
                       std::string const& instance_id,
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -100,13 +100,12 @@ void AddColumn(std::vector<std::string> const& argv) {
   (argv[0], argv[1], argv[2]);
 }
 
-void QueryWithStruct(std::vector<std::string> const& argv) {
-  // TODO(tmatsuo): Add querying part once data client is ready.
+void QueryWithStruct(std::vector<std::string> const&) {
+  // TODO(#188): Add querying part once data client is ready.
   // [START spanner_create_struct_with_data]
   auto singer_info = std::make_tuple(std::make_pair("FirstName", "Elena"),
                                      std::make_pair("LastName", "Campbell"));
   // [END spanner_create_struct_with_data]
-  (void)argv;
   std::cout << "Struct created with the following data:\n"
             << std::get<0>(singer_info).first << ":"
             << std::get<0>(singer_info).second << "\n"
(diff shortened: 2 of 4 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -98,6 +100,20 @@ void AddColumn(std::vector<std::string> const& argv) {
   (argv[0], argv[1], argv[2]);
 }
 
+void QueryWithStruct(std::vector<std::string> const& argv) {
+  // TODO(tmatsuo): Add querying part once data client is ready.
+  // [START spanner_create_struct_with_data]
+  auto singer_info = std::make_tuple(std::make_pair("FirstName", "Elena"),
+                                     std::make_pair("LastName", "Campbell"));
+  // [END spanner_create_struct_with_data]
+  (void)argv;
```
devjgm: You can just omit the name `argv` from the function argument since we don't use the argument in this function. That *should* quiet whatever warning you were seeing.

google/cloud/spanner/samples/samples.cc
```diff
@@ -98,6 +100,20 @@ void AddColumn(std::vector<std::string> const& argv) {
   (argv[0], argv[1], argv[2]);
 }
 
+void QueryWithStruct(std::vector<std::string> const& argv) {
+  // TODO(tmatsuo): Add querying part once data client is ready.
```
devjgm: We try to put github issues in our TODOs rather than usernames. Here, I think there's already an issue for this, so I can you can write `TODO(#191)`

google/cloud/spanner/database_admin_client.h
```diff
@@ -126,7 +126,7 @@ class DatabaseAdminClient {
    *   completes on the service. Note that this can take minutes in some cases.
    *
    * @par Example
-   * @snippet database_admin_samples.cc create-database
+   * @snippet samples.cc create-database
```
coryan: Why the rename?  We are going to have dozens, maybe as many as 100 samples. The files get unwieldy if they have too many of them.  Why not simply create a new sample, say `data_access_samples.cc` or something?

coryan:
Thanks for the PR.  It feels a bit early to be working on this, the remaining work (actually using this in a `ExecuteSql()`) is probably most of the work, right?

google/cloud/spanner/database_admin_client.h
devjgm: I'll explain our rationale, then I'd love to hear if you have an alternate suggestion.

We discussed this locally yesterday. The thinking was that all of our samples will need some of the same infrastructure, like the ability to create a random database name, create the database, delete the database, some machinery to provide a command-line interface for executing the samples, etc. We could split that into a `database_admin_samples.cc` file and some *non-admin-samples.cc* file, or into the admin samples and N other files for different types of samples. Of course, we could also factory out common parts that would be shared by all the different samples flies, too.

We weren't sure about the best way to split up the samples. So we decided to put all the samples in a single file. That's the easiest solution for now, requiring little more than renaming the file to something not admin specific. Then if/when the `samples.cc` file becomes too big or unwieldy we'll be in a better position to split it up logically.

So, that's what we were thinking. Given that you have more experience with this, I'd really like to hear your thoughts.

devjgm:
> Thanks for the PR. It feels a bit early to be working on this, the remaining work (actually using this in a `ExecuteSql()`) is probably most of the work, right?

Yes, using the struct will be most of the work. But just declaring the struct is a sample that we need to provide ([like they do in C# for example](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/a129e6228fb4a6cb660832168b8f02ac0675673f/spanner/api/Spanner/Program.cs#L2602)). While we can't write most of the needed samples yet, it seems OK IMO to start writing the ones that we can.

google/cloud/spanner/database_admin_client.h
coryan: Ack. Splitting later seems fine.

coryan:
> Yes, using the struct will be most of the work. But just declaring the struct is a sample that we need to provide

Oh, I see now, #187 vs. #188. Makes total sense now.

google/cloud/spanner/samples/samples.cc
tmatsuo: Done

google/cloud/spanner/samples/samples.cc
tmatsuo: I did it with #188

tmatsuo:
Thanks! PTAL

PR #231, opened 2019-07-17
feat: Separate AddRow vs. EmplaceRow.

### LINKED ISSUE #222: Consider a `*MutationBuilder::AddRow(Row)` overload.
Maybe we should have overloads that take a `Row` parameter.

coryan:
This proposes two separate functions to add the values vs. add a row object into the mutations.

Part of the work for #198. Fixes #222.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/231)

### DIFF
--- a/google/cloud/spanner/mutations.h
+++ b/google/cloud/spanner/mutations.h
@@ -15,8 +15,8 @@
 #ifndef GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_SPANNER_MUTATIONS_H_
 #define GOOGLE_CLOUD_CPP_GOOGLE_CLOUD_SPANNER_MUTATIONS_H_
 
+#include "google/cloud/spanner/row.h"
 #include "google/cloud/spanner/value.h"
-#include "google/cloud/spanner/version.h"
 #include <google/spanner/v1/mutation.pb.h>
 
 namespace google {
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/mutations.h
```diff
@@ -119,13 +119,23 @@ class WriteMutationBuilder {
   Mutation Build() && { return Mutation(std::move(m_)); }
 
   template <typename... Ts>
-  WriteMutationBuilder& AddRow(Ts&&... values) {
+  WriteMutationBuilder& AddValues(Ts&&... values) {
```
devjgm: I think "AddRow" is the right naming and terminology here. This matches my reading of the [proto docs](https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/mutation.proto) where mutations are described in terms of updating "rows". They use the term ["values" here](https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/mutation.proto#L60) in the proto, but I think that's because they use a single `google.protobuf.ListValue` to represent a "row". But the documentation about that field still suggests that logically it's a list of "rows". I imagine if they had a "Row" proto message they'd have named it "rows".

That's a long winded (sorry) way of saying that I think "row" is the right concept for what mutations operate on, and it'd be good to use this terminology consistently in our APIs.

Therefore, I tend to think that `AddRow()` may be all we want. I don't know if `Addvalues` adds value (*rimshot*). I think it may be more confusing than helpful. For example, since "row" is nowhere in the function's name or signature, it may not be clear if `x.AddValues(a, b, c)` is adding one row with three values, or if it's adding three rows each with one value. Of course, only one interpretation is correct (and the number of columns should be a hint), but it seems like an easy mistake that may not be noticed until runtime. And contrast `x.AddValues(a, b, c)` with `x.AddRow(MakeRow(a, b, c))`. I think the latter is much clearer.

What's the downside of having *only* `AddRow(Row<Ts....>)` ? It seems like a simplifying feature to have a single abstraction for representing rows that's used consistently throughout our APIs. Are there cases where `Row<Ts....>` is an inadequate representation for a row? If so, maybe we can fix it. If it helps, I'm pretty sure it would be easy to make `Row<Value, Value, Value>` work well and efficiently. 

Maybe there's a performance argument to be made that enabling users to add a row's ingredients rather than the row itself. Maybe. If there's a need to do that, maybe we want to go with the "emplace" terminlogy, like "EmplaceRow(Ts&&... values)". That also makes it clear at the call site that the value arguments are the ingredients making up the logical row.

What do you think?

coryan: I think `AddRow()` and `EmplaceRow()` are better names.

As I said before: forcing the user to write `x.AddRow(MakeRow(a, b, c))` seems like we are making life easy for us but hard for them. If that is not convincing I will remove the `EmplaceRow()` function as part of this PR.

devjgm: I'm not sure I agree with "hard" for them. I think the simplicity of having a single function for adding rows, and a single type for representing rows would make things simpler and easier for users as well as for us (much like functions that take `std::string` are rarely overloaded to also take `char const*` + `std::size_t`).

But maybe I'm wrong! :-) And most of my concerns are addressed with the "EmplaceRow" naming, so it's OK with me if want to keep it.

coryan:
Rebased and resolved conflicts.

PR #239, opened 2019-07-18
chore: update google-cloud-cpp to provide conjunction

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/239)


remyabel:
You will probably want to update the Dockerfiles and Bazel files too.

Issue #195, opened 2019-07-11
Create "external" project showing how to use Spanner lib with cmake and bazel


devjgm:
Developers must be able to easily use Cloud Spanner C++ client in their C++ project. We will support both Bazel and CMake projects. Comment from @coryan 

> ...this suggests we need a sample or test that uses the google-cloud-cpp-spanner project.
> 
> - It could be part of the same repository, but that makes it hard to show that the sample is really using the project "as-if" it was completely separate
> - or it could be in a separate repository, but that makes hard to test for every change....
> 
> Seems like a mini-design doc would be nice.

coryan:
For what is worth, @remyabel contributed an "external" CMake project in #197.

The project lives in `ci/test-readme`, it will be compiled by one of CI builds.

I think this means we just need to:

- [ ] Create Bazel files for the same project.
- [ ] Test the Bazel files as part of the CI builds (maybe separate builds, or just one, because why test it everywhere)?
- [ ] Update the landing page for Doxygen to describe how to integrate the project with CMake and Bazel.

remyabel:
You mean `ci/test-install` not `ci/test-readme`?

coryan:
Doh, yes, `ci/test-install`, fixed the comment.

coryan:
I think this is done. There is the issue of running the test-install program as part of the CI, but we have two other issues open for it.

PR #247, opened 2019-07-19
feat: Create landing page for reference docs.

coryan:
This creates a landing page for the reference docs and describes how to
use the library from CMake, Bazel, and Make.

Part of the work for #195.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/247)

### DIFF
--- a/google/cloud/spanner/doc/spanner-main.dox
+++ b/google/cloud/spanner/doc/spanner-main.dox
@@ -39,7 +39,11 @@ should be automatically compiled when you follow said instructions.
 
 ### Configuring authentication for the C++ Client Library
 
-This library uses the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to
+Like all Google Cloud Platform (GCP) services, Cloud Spanner requires that your
+application authenticates with the service before accessing any data.
+If you are not familiar with GCP authentication please take this opportunity to
+review the [Authentication Overview][authentication-quickstart]. This library
+uses the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to
 find the credentials file. For example:
 
 | Shell              | Command                                        |
@@ -102,8 +106,6 @@ you want to use.
 In your [WORKSPACE][workspace-definition-link] file add a dependency to download
 and install the library, for example:
 
-[workspace-definition-link]: https://docs.bazel.build/versions/master/build-ref.html#packages_targets
-
 @code {.py}
 # Change the version and SHA256 hash as needed.
 http_archive(
@@ -130,7 +132,7 @@ switched_rules_by_language(
     grpc = True,
 )
 
-# Have to manually call the corresponding function for google-cloud-cpp and
+# You have to manually call the corresponding function for google-cloud-cpp and
 # gRPC:  https://github.com/bazelbuild/bazel/issues/1550
 load("@com_github_googleapis_google_cloud_cpp//bazel:google_cloud_cpp_deps.bzl", "google_cloud_cpp_deps")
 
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -0,0 +1,220 @@
...
+    $(CXX) $(CXXFLAGS) $(SPANNER_CXXFLAGS) $(SPANNER_CXXLDFLAGS) -o $@ $^ $(SPANNER_LIBS)
+```
+
+### Error Handling
+
+[status-or-header]: https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/status_or.h
+
+This library never throws exceptions to signal error. In general, the library
+returns a [`StatusOr<T>`][status-or-header] if an error is possible. Some
+functions return objects that already have an existing error handling mechanism,
```
devjgm: Does the Spanner library actually return any type that signals errors using something other than StatusOr? I don't think returns any ostreams. It might grow these things, but I don't think we have them yet.

Do we want to keep this blurb in case we grow these new error-signaling types, or remove it and add it back if/when we need to?

Honestly, I don't have a strong opinion here. Mostly just confirming if my understanding of the library's current state is accruate.

google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -0,0 +1,220 @@
...
+SPANNER_LIBS       := $(shell pkg-config spanner_client --libs-only-l)
+
+# A target using the Cloud Spanner C++ client library.
+my_program: my_program.cc
+    $(CXX) $(CXXFLAGS) $(SPANNER_CXXFLAGS) $(SPANNER_CXXLDFLAGS) -o $@ $^ $(SPANNER_LIBS)
+```
+
+### Error Handling
+
+[status-or-header]: https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/status_or.h
```
devjgm: Do you want to collect all of these short link things at the bottom of the file with the others?

coryan: Done.

google/cloud/spanner/doc/spanner-main.dox
coryan: > Does the Spanner library actually return any type that signals errors using something other than StatusOr?

We return `future<StatusOr<T>>`. We are also planning to return `ResultSet` both of which use `StatusOr<>` to signal an error "later", but are not wrapped by a `StatusOr<>`. The stuff about iostream was (obviously) a cut&paste error.

google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -0,0 +1,220 @@
...
+# Configure @com_google_googleapis to only compile C++ and gRPC:
+load("@com_google_googleapis//:repository_rules.bzl", "switched_rules_by_language")
+
+switched_rules_by_language(
+    name = "com_google_googleapis_imports",
+    cc = True,
+    grpc = True,
+)
+
+# Have to manually call the corresponding function for google-cloud-cpp and
```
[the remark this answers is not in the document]
coryan: Done.

google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -0,0 +1,220 @@
...
+should be automatically compiled when you follow said instructions.
+
+### Configuring authentication for the C++ Client Library
+
+This library uses the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to
+find the credentials file. For example:
+
+| Shell              | Command                                        |
+| :----------------- | ---------------------------------------------- |
+| Bash/zsh/ksh/etc.  | `export GOOGLE_APPLICATION_CREDENTIALS=[PATH]` |
```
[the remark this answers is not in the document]
coryan: Included a link to the authentication overview.

PR #251, opened 2019-07-22
feat: add MergeChunk() function for "unchunking" incoming Values

### LINKED ISSUE #161: Implement proto Value "unchunker"
According to Spanner's [ResultSet proto](https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/result_set.proto), the `PartialResultSet` message that is returned from streaming operations may have "chunked" values. These are conceptually single values that may actually be split across several adjacent "value" proto bufs. See the [docs here](https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/result_set.proto#L140) about how the chunking works.

Well, since results may be chunked, we need an "unchunker". That is, we need something that can take a "stream" of proto Values and recombine them into the single logical Value that they represent.

devjgm:
Fixes #161

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/251)


google/cloud/spanner/internal/merge_chunk_test.cc
```diff
@@ -0,0 +1,219 @@
...
+// google::protobuf::Value protos from convenient user-supplied arguments.
+//
+
+google::protobuf::Value MakeProtoValue(Value v) {
+  google::protobuf::Value value;
+  std::tie(std::ignore, value) = internal::ToProto(std::move(v));
+  return value;
+}
+
+google::protobuf::Value MakeProtoValue(std::string s) {
```
devbww: Perhaps `#include <string>`?

google/cloud/spanner/internal/merge_chunk_test.cc
```diff
@@ -0,0 +1,219 @@
...
+}
+
+google::protobuf::Value MakeProtoValue(double d) {
+  google::protobuf::Value value;
+  value.set_number_value(d);
+  return value;
+}
+
+template <typename T>
+google::protobuf::Value MakeProtoValue(std::vector<T> v) {
```
devbww: Perhaps `#include <vector>`?

google/cloud/spanner/internal/merge_chunk_test.cc
```diff
@@ -0,0 +1,219 @@
...
+TEST(MergeChunk, ExampleListsOfListOfString) {
+  auto a = MakeProtoValue(std::vector<Value>{
+      Value("a"), Value(std::vector<std::string>{"b", "c"})});
+  auto b = MakeProtoValue(
+      std::vector<Value>{Value(std::vector<std::string>{"d"}), Value("e")});
+  ASSERT_STATUS_OK(MergeChunk(a, std::move(b)));
+
+  auto expected = MakeProtoValue(std::vector<Value>{
+      Value("a"), Value(std::vector<std::string>{"b", "cd"}), Value("e")});
+
```
devbww: Remove blank line to match previous tests?

google/cloud/spanner/internal/merge_chunk.cc
```diff
@@ -0,0 +1,81 @@
...
+      // Moves all the remaining elements over.
+      while (chunk_it != chunk_list.end()) {
+        *value.mutable_list_value()->add_values() = std::move(*chunk_it++);
+      }
+
+      return Status();
+    }
+
+    default:
+      return Status(StatusCode::kUnknown, "unknown Value type");
```
devbww: Just `break` if we're going to have the final `return`?

devjgm: Done.

PR #257, opened 2019-07-26
feat: Added Value::Bytes support

### LINKED ISSUE #52: Implement spanner::Value support for BYTES
Current `spanner::Value` implementation: https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/value.h

Spanner types to be supported: https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/type.proto

devjgm:
Fixes #52

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/257)

### DIFF
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -202,7 +202,7 @@ google::protobuf::Value Value::MakeValueProto(std::string s) {
   return v;
 }
 
-google::protobuf::Value Value::MakeValueProto(Value::Bytes bytes) {
+google::protobuf::Value Value::MakeValueProto(Value::Bytes const& bytes) {
   google::protobuf::Value v;
   v.set_string_value(internal::Base64Encode(bytes.data));
   return v;
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/spanner/value.h"
+#include "google/cloud/spanner/internal/base64.h"
 #include "google/cloud/spanner/internal/date.h"
 #include "google/cloud/spanner/internal/time.h"
 #include "google/cloud/log.h"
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/value.cc
```diff
@@ -274,6 +288,17 @@ StatusOr<std::string> Value::GetValue(std::string const&,
   return pv.string_value();
 }
 
+StatusOr<Value::Bytes> Value::GetValue(Value::Bytes const&,
+                                       google::protobuf::Value const& pv,
+                                       google::spanner::v1::Type const&) {
+  if (pv.kind_case() != google::protobuf::Value::kStringValue) {
+    return Status(StatusCode::kUnknown, "missing BYTES data");
```
devbww: FYI: Not covered by test.

google/cloud/spanner/value.cc
```diff
@@ -194,6 +202,12 @@ google::protobuf::Value Value::MakeValueProto(std::string s) {
   return v;
 }
 
+google::protobuf::Value Value::MakeValueProto(Value::Bytes bytes) {
```
devbww: Can we avoid copying the bytes here?

devjgm: Yep, good catch. Fixed.

google/cloud/spanner/value.cc
devjgm: Tested now.

devbww: The new test looks like it hits the decode error below, but not this "missing" case.  Perhaps the other "missing"s are not hit either?

devjgm: I think other missing cases are tested, like in the test named GetBadBool. I missed this one for Bytes, but I just added a couple more tests for `Value::Bytes`, which I think brings up to matching the other value-type tests.

Issue #82, opened 2019-06-11
Create a CI build with exceptions disabled.


coryan:
We need at least one CI build that disables exceptions. Probably something based on GCC.

devjgm:
Fixed in #263 #258

PR #258, opened 2019-07-27
chore: Add build that runs with exceptions disabled

devjgm:
This completes part of #82 . After this, we can hook this build up to a CI and presubmit job. This was mostly copied from the equivalent configuration in github.com/googleapis/google-cloud-cpp

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/258)


google/cloud/spanner/internal/transaction_impl_test.cc
```diff
@@ -124,6 +124,9 @@ ResultSet Client::Read(TransactionSelector& selector, std::string const&,
 #if __EXCEPTIONS
         if (fail_with_throw) throw "1202 Program Alarm";
 #endif
+        // Quiets a compiler warning about an unused variable (when exceptions
+        // are disabled).
+        (void)fail_with_throw;
```
devbww: Alternatively, perhaps ...
```
  if (fail_with_throw) {
#if __EXCEPTIONS
    throw "1202 Program Alarm";
#endif
  }
  break;
```

devjgm: LGTM. Done.

PR #269, opened 2019-07-30
Add a CommitResult type.

mr-salty:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/269)

### DIFF
--- a/google/cloud/spanner/commit_result.h
+++ b/google/cloud/spanner/commit_result.h
@@ -1,0 +1,40 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/commit_result.h
```diff
@@ -0,0 +1,40 @@
...
+#include "google/cloud/spanner/timestamp.h"
+#include "google/cloud/spanner/version.h"
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+/**
+ * The result of Committing a Transaction.
```
coryan: nit: s/Committing/committing/ ?

mr-salty:
FYI, The design doc has this as a class, but it seems to me that a struct is sufficient unless we have some philosophical objection to it.

google/cloud/spanner/commit_result.h
```diff
@@ -0,0 +1,40 @@
...
+ */
+struct CommitResult {
+ public:
```
devjgm: We should probably omit the `public:` for structs.

mr-salty: oops - I forgot to delete it when I made it a struct - thanks!

PR #272, opened 2019-07-31
refactor: Sample refactoring

tmatsuo:
Split the admin samples and data client samples. I also added `-j 4` for integration test run so that they can run in parallel.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/272)

### DIFF
--- a/ci/kokoro/docker/build-in-docker-cmake.sh
+++ b/ci/kokoro/docker/build-in-docker-cmake.sh
@@ -98,7 +98,8 @@ if [[ ${RUN_INTEGRATION_TESTS} == "yes" ]]; then
   # Run the integration tests too.
   env -C "${BINARY_DIR}" ctest \
       -L integration-tests \
-      --output-on-failure
+      --output-on-failure \
+      -j 4
   echo "================================================================"
 fi
 
--- a/google/cloud/spanner/database_admin_client.h
+++ b/google/cloud/spanner/database_admin_client.h
@@ -127,7 +127,7 @@ class DatabaseAdminClient {
    *   completes on the service. Note that this can take minutes in some cases.
    *
    * @par Example
-   * @snippet samples.cc create-database
+   * @snippet sample_utils.cc create-database
    *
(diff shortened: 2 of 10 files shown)


coryan:
I am basically Okay with these changes, but I thought you had agreed with the team to keep everything in one file?  Maybe I misunderstood.

devjgm:
I'm similarly OK w/ splitting these in general, but I also thought we were going to keep them in one file until we had need to split them. Did a need arise?

tmatsuo:
I think one of the reasons to keep them in one file was to reduce the time for integration test, by not having multiple integration tests, but luckily we can add a ctest flag to run them in parallel, so it is now faster than before even if we have more tests.

Well, there's no immediate need though, but considering the data client is coming, by having them separate might make it easy to add samples in parallel?

tmatsuo:
Rebased and fixed the include guard. PTAL

ci/kokoro/docker/build-in-docker-cmake.sh
```diff
@@ -98,7 +98,8 @@ if [[ ${RUN_INTEGRATION_TESTS} == "yes" ]]; then
   # Run the integration tests too.
   env -C "${BINARY_DIR}" ctest \
       -L integration-tests \
-      --output-on-failure
+      --output-on-failure \
+      -j 4
```
remyabel: Might as well make this `"$(nproc)"` to be consistent?

tmatsuo: Well, actually we can run test in parallel even on single core machine because I believe they are in separate processes and most of the time is spent waiting for network io.

devbww: There is a mismatch between `-j` (run `N` jobs in parallel) and `nproc` (the number of processing units available).  In general I think you want to scale `nproc` by some "job load factor" to get `N`.  If the jobs are CPU bound, the factor is 1, but for I/O bound jobs it could be significantly higher.

tmatsuo:
closing for now

PR #275, opened 2019-08-01
Add normalization to `google::cloud::spanner::Date`

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/275)

### DIFF
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -35,6 +35,7 @@ add_library(spanner_client
             commit_result.h
             database_admin_client.cc
             database_admin_client.h
+            date.cc
             date.h
             internal/base64.cc
             internal/base64.h
--- a/google/cloud/spanner/date.cc
+++ b/google/cloud/spanner/date.cc
@@ -37,13 +37,13 @@ int DaysPerMonth(std::int64_t y, int m) {
 
 Date::Date(std::int64_t year, int month, int day)
     : year_(year), month_(month), day_(day) {
-  year_ += month_ / 12;
+  year_ += month_ / 12;  // 12 months == 1 year
   month_ %= 12;
   if (month_ <= 0) {
     year_ -= 1;
     month_ += 12;
   }
-  year_ += (day_ / 146097) * 400;
+  year_ += (day_ / 146097) * 400;  // 146097 days == 400 years
   day_ %= 146097;
   if (day_ <= 0) {
     year_ -= 400;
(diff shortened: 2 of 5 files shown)


devbww:
> what's the significance of this?

https://www.google.com/search?q=march+14+1879

I'll happily entertain alternate ideas for the value of default-constructed Date.

devbww:
> Add some comments about what this is computing

Done.

devbww:
> Are we sure we want this functionality?

I would say yes.

> The Spanner library itself does not need a Date class with any smarts, does it?

It depends on what you think something like `internal::DateFromString("2019-06-31")` should return.  I would say it should either return an error, or something equal to `Date(2019, 7, 1)`, both of which require this functionality.

At the moment `internal::DateFromString("2019-06-31")` succeeds and returns something "weird", while `internal::DateFromString("2019-06-32")` fails.  That seems inconsistent at best.

devjgm:
> > Are we sure we want this functionality?
> 
> I would say yes.
> 
> > The Spanner library itself does not need a Date class with any smarts, does it?
> 
> It depends on what you think something like `internal::DateFromString("2019-06-31")` should return. I would say it should either return an error, or something equal to `Date(2019, 7, 1)`, both of which require this functionality.

I'd vote for an error for two reasons:

1. `internal::DateFromString` already returns StatusOr. So if an invalid dates doesn't produce an error status, what does?

2. I think this Spanner library is intended for reading, writing, querying the Spanner database. All the "spanner logic" is in the cloud. "smarts" in the client library that hide bad data (i.e., invalid dates) could be seen as a bug.

> 
> At the moment `internal::DateFromString("2019-06-31")` succeeds and returns something "weird", while `internal::DateFromString("2019-06-32")` fails. That seems inconsistent at best.

I agree. That's weird, and it would be good to fix.

If you say this functionality is needed regardless, then SGTM. Do you think the logic is needed in the Date class, or in the parsing logic?

Either way, I'm happy to approve to unblock you to take this the way you think makes the most sense.

devjgm:
> > what's the significance of this?
> 
> https://www.google.com/search?q=march+14+1879
> 
> I'll happily entertain alternate ideas for the value of default-constructed Date.

As boring as it is, I think most users would expect Jan 1, 1970, so I'm afraid, that'd get my vote.

devbww:
>> I'll happily entertain alternate ideas for the value of default-constructed Date.
> 
> As boring as it is, I think most users would expect Jan 1, 1970, so I'm afraid, that'd get my vote.

Done.

devbww:
> Do you think the logic is needed in the Date class, or in the parsing logic?

I think here is preferable.  Then the user of `Date` can decide whether normalization is desirable or not, be they `DateFromString()` [where we agree it isn't], or anyone else.  Thanks.

PR #276, opened 2019-08-01
feat: Skeleton for spanner::Client

### LINKED ISSUE #29: Implement spanner::Client skeleton
Implement the skeleton of a `spanner::Client` class, which should initially look something like the following:

```cc
namespace google {
namespace cloud {
namespace spanner {

struct ClientOptions {
  std::string project_id;
  std::string instance_id;
  std::string database_id;
  std::map<std::string, std::string> labels;
};

class Client {
 public:
  Client(Client&&) = default;
  Client& operator=(Client&&) = default;
  Client(Client const&) = default;
  Client& operator=(Client const&) = default;

  friend bool operator==(Client a, Client b);
  friend bool operator!=(Client a, Client b);

  ClientOptions options() const;

  // TODO:
  // Read()
  // ExecuteSql()
  // ExecuteBatchDml()
  // ExecutePartitionedDml()
  // RunTransaction()

 private:
  Client(ClientOptions);
};

StatusOr<Client> MakeClient(ClientOptions opts);

}  // namespace spanner
}  // namespace cloud
}  // namespace google
```

mr-salty:
Fixes #29

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/276)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -1,0 +1,126 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/spanner/client.h"
+#include "google/cloud/spanner/value.h"
+#include "google/cloud/grpc_utils/grpc_error_delegate.h"
+#include "google/cloud/log.h"
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+namespace spanner_proto = ::google::spanner::v1;
+
+StatusOr<Client> Client::MakeClient(ClientOptions const& client_options) {
+  return Client(internal::CreateDefaultSpannerStub(client_options));
+}
+
+StatusOr<ResultSet> Client::Read(std::string const& /*table*/,
+                                 KeySet const& /*keys*/,
+                                 std::vector<std::string> const& /*columns*/,
+                                 ReadOptions const& /*read_options*/) {
+  return Status(StatusCode::kUnimplemented, "not implemented");
+}
+
+StatusOr<ResultSet> Client::Read(
+    Transaction::SingleUseOptions const& /*transaction_options*/,
+    std::string const& /*table*/, KeySet const& /*keys*/,
+    std::vector<std::string> const& /*columns*/,
+    ReadOptions const& /*read_options*/) {
+  return Status(StatusCode::kUnimplemented, "not implemented");
+}
+
+StatusOr<ResultSet> Client::Read(Transaction const& /*transaction*/,
+                                 std::string const& /*table*/,
+                                 KeySet const& /*keys*/,
+                                 std::vector<std::string> const& /*columns*/,
+                                 ReadOptions const& /*read_options*/) {
+  return Status(StatusCode::kUnimplemented, "not implemented");
+}
+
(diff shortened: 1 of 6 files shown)


mr-salty:
The doxygen stuff isn't perfect but let's fix that later so we can unblock people.
I'm not sure why the comment only shows up for one overload, maybe I read the manual incorrectly.

google/cloud/spanner/client.h
```diff
@@ -0,0 +1,379 @@
...
+inline namespace SPANNER_CLIENT_NS {
+
+/// Options passed to `Read` or `PartitionRead`.
+struct ReadOptions {
+  /**
+   * If non-empty, the name of an index on a database table. This index is used
+   * instead of the table primary key when interpreting the `KeySet`and sorting
+   * result rows.
+   */
+  std::string index_name;
```
remyabel: Any reason this shouldn't be `optional<std::string>`?

google/cloud/spanner/client.cc
```diff
@@ -0,0 +1,126 @@
...
+
+StatusOr<Client> Client::MakeClient(ClientOptions const& client_options) {
+  return Client(internal::CreateDefaultSpannerStub(client_options));
+}
+
+StatusOr<ResultSet> Client::Read(std::string const& /*table*/,
+                                 KeySet const& /*keys*/,
+                                 std::vector<std::string> const& /*columns*/,
+                                 ReadOptions const& /*read_options*/) {
+  return Status(StatusCode::kUnimplemented, "not implemented");
```
devjgm: I believe this is a 1-liner forwarding to the overload that takes a `Transaction::SingleUseOptions`.

google/cloud/spanner/client.cc
```diff
@@ -0,0 +1,126 @@
...
+}
+
+std::vector<StatusOr<spanner_proto::ResultSetStats>> Client::ExecuteBatchDml(
+    Transaction const& /*transaction*/,
+    std::vector<SqlStatement> const& /*statements*/, int64_t /*seqno*/) {
+  return {Status(StatusCode::kUnimplemented, "not implemented")};
+}
+
+// returns a lower bound on the number of modified rows
+StatusOr<int64_t> Client::ExecutePartitionedDml(
```
devjgm: To keep to "standard c++" as much as possible, we've been trying to consistently use `std::int64_t` et al.

devjgm:
On Thu, Aug 1, 2019 at 10:00 PM Todd Derr <notifications@github.com> wrote:

> *@mr-salty* commented on this pull request.
>
> Reviewed 1 of 4 files at r1, 3 of 5 files at r2, 1 of 1 files at r3.
> *Reviewable
> <https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/276>*
> status: 5 of 6 files reviewed, 10 unresolved discussions (waiting on
> @coryan <https://github.com/coryan>, @devbww <https://github.com/devbww>,
> @devjgm <https://github.com/devjgm>, @mr-salty
> <https://github.com/mr-salty>, @remyabel <https://github.com/remyabel>,
> @return <https://github.com/return>, @returns <https://github.com/returns>,
> and @scotthart <https://github.com/scotthart>)
>
> ------------------------------
>
> *google/cloud/spanner/client.cc, line 97 at r1
> <https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/276#-LlAhsW8swFkh_8Ti5Q9-r1-97:-LlF1uOKCBAdR-hjlWMS:bc37w3h>
> (raw file
> <https://github.com/googleapis/google-cloud-cpp-spanner/blob/4f8ee1352e4cb903f5a946ba42f89ebc2a4ccf39/google/cloud/spanner/client.cc#L97>):*
> *Previously, devjgm (Greg Miller) wrote…*
>
> I vote to remove them, because I don't think we should spend time working
> on them until the higher-priority alpha stuff is done. These are empty
> function bodies, so we wouldn't be removing much.
>
> If you'd rather keep them, please put a comment here reminding future us
> to ignore these until alpha is done.
>
> I added a warning comment... personally I'd rather have an indication that
> something is intentionally missing than having to wonder if that's the case
> or if it was an oversight. As one example from this afternoon, Brad and I
> were discussing Session management and weren't sure why there were no
> labels in ClientOptions - intentional or not? Then it derailed us for a
> while trying to figure that out.
>
OK. Please add the comment to the .cc file too. The comment is for *us*,
not customers.

PR #278, opened 2019-08-01
feat: Add build info

### LINKED ISSUE #81: Fix build metadata for the spanner library.
The library is picking up build metadata, such as the git revision, compiler version, and compiler flags, from the build for `google_cloud_cpp_common`. That is misleading, we should use the build information for the library itself.

devjgm:
Fixes #81 

This PR mostly follows the "build_info" pattern established by https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/internal/build_info.h with the following notable differences:

* Some functions are spelled differently. The naming follows C++ style convention (e.g., `LanguageVersion()` instead of `language_version()`) to avoid the need for NOLINT.
* Some functions have a different name to make their intent and semantics clearer. For example, rather than `LanguageVersion()` returning all of the compiler ID, the compiler version, and the language version, those parts are all returned by 3 clearly named separate functions.
* The bazel build uses scripts in the //bazel/ directory to extract the "compiler ID" and "compiler version" in a format that is compatible with the format provided by CMake.
* Bazel builds always report the git commit has as "unknown", because I don't know of any way to correctly extract this info from within a bazel build rule. I don't think this is actually a change, because I believe the GCS and BigTable bazel builds also always report "unknown" too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/278)


bazel/compiler_id.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# Given an argument that is the path to a compiler, tries to determine an
+# identification string for the compiler according to the CMake-defined IDs:
+# https://cmake.org/cmake/help/v3.5/variable/CMAKE_LANG_COMPILER_ID.html
+
+if [[ ! -e "${1}" ]]; then
```
devbww: What does existence have to do with usage?

google/cloud/spanner/internal/build_info.h
```diff
@@ -0,0 +1,81 @@
...
+ * Returns the compiler version.
+ *
+ * This string will be something like "9.1.1".
+ */
+std::string CompilerVersion();
+
+/**
+ * Returns the compiler flags.
+ *
+ * Examples include "-c fastbuild" or "-O2 -DNDEBUG".
```
devbww: Is `-c fastbuild` really a compiler flag?

bazel/compiler_id.sh
```diff
@@ -0,0 +1,48 @@
...
+  echo "Usage: $0 <path-to-compiler>"
+  exit 1
+fi
+
+version="$("${1}" --version 2> /dev/null \
```
remyabel: If we're going to support multiple compilers, the script will probably grow larger as we detect the executable and use the appropriate switch/flags. Just spitballing, but what do you think about adopting an approach similar to [this](https://github.com/TristanThrush/drake-9660/blob/e8dd628593786a6ddfed6758bc4ab3c2d58ae0c4/tools/workspace/cc/identify_compiler.cc)? TL;DR, compile a program in Bazel that uses compiler-specific version macros then spits out the information.

bazel/compiler_id.sh
devbww: >> Just since we're about to execute it... ?? Let me know if you want me to change it to something else.
> 
> On second though, you're right. Changed to a more normal $# -ne 1 check. Fixed. thanks.

Alright!

google/cloud/spanner/internal/build_info.h
devbww: >> Is -c fastbuild really a compiler flag?
>
> We don't have the actual compiler flags that were used ... so we use the COMPILATION_MODE as a proxy.

OK.  Then I'll only suggest that something like `BuildFlags()` might be a better name.

devjgm:
Fixes #81

This PR mostly follows the "build_info" pattern established by https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/internal/build_info.h with the following notable differences:

Some functions are spelled differently. The naming follows C++ style convention (e.g., LanguageVersion() instead of language_version()) to avoid the need for NOLINT.
Some functions have a different name to make their intent and semantics clearer. For example, rather than LanguageVersion() returning all of the compiler ID, the compiler version, and the language version, those parts are all returned by 3 clearly named separate functions.
The bazel build uses scripts in the //bazel/ directory to extract the "compiler ID" and "compiler version" in a format that is compatible with the format provided by CMake.
Bazel builds always report the git commit has as "unknown", because I don't know of any way to correctly extract this info from within a bazel build rule. I don't think this is actually a change, because I believe the GCS and BigTable bazel builds also always report "unknown" too.

PR #282, opened 2019-08-02
Added warning not to work on BatchDML until after Alpha. [skip ci]

devjgm:
This comment is intended for *us*, not for customers, so I think it would be useful to have in the .cc file as well.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/282)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -88,6 +88,8 @@ StatusOr<std::vector<SqlPartition>> Client::PartitionQuery(
 std::vector<StatusOr<spanner_proto::ResultSetStats>> Client::ExecuteBatchDml(
     Transaction const& /*transaction*/,
     std::vector<SqlStatement> const& /*statements*/) {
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/client.cc
```diff
@@ -88,6 +88,8 @@ StatusOr<std::vector<SqlPartition>> Client::PartitionQuery(
 std::vector<StatusOr<spanner_proto::ResultSetStats>> Client::ExecuteBatchDml(
     Transaction const& /*transaction*/,
     std::vector<SqlStatement> const& /*statements*/) {
+   // This method is NOT part of the Alpha release. Please do not work on this
+   // until all the higher-prioty alpha work is finished.
```
mr-salty: s/prioty/priority/

devjgm: oops. fixed. thanks.

PR #295, opened 2019-08-02
Supply the database name when creating a Client.

mr-salty:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/295)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -115,6 +115,13 @@ StatusOr<Client> MakeClient(std::string database_name,
                 internal::CreateDefaultSpannerStub(client_options));
 }
 
+std::string MakeDatabaseName(std::string const& project,
+                             std::string const& instance,
+                             std::string const& database_id) {
+  return std::string("projects/") + project + "/instances/" + instance +
+         "/databases/" + database_id;
+}
+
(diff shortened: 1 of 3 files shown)


coryan:
___
*[google/cloud/spanner/client.h, line 93 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/295#-LlITJIa6ptJius1YiGb:-LlIWELB3qeG82Dq8JMo:bladtue) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/f4daecfc5763bf86433cb0a568a2bf55be6e58e8/google/cloud/spanner/client.h#L93)):*

I *think* (I am not certain, but the spanner_tool used to do this) that the "database" parameter needs to be fully qualified, i.e.,  `projects/<project-id>/instances/<instance-id>/databases/<database-id>`.  If I am right, I think this example should show that.

https://github.com/googleapis/google-cloud-cpp-spanner/blob/e000532c6b71f0e0d15b41830fa1ef60bd6cf234/google/cloud/spanner/spanner_tool.cc#L181-L205



</blockquote></details>

Also, if I happen to be right about this, we should create a helper function like `MakeDatabaseName(project_id, instance_id, database_id)`

PR #297, opened 2019-08-02
feat: Create configuration for macOS+Bazel build.

coryan:
Create driver scripts for macOS and Bazel. In this initial version
the integration tests are not executed. That will happen in a future
PR.

This commit also includes a draft script to build with CMake, but it
does not work yet. Or more accurately, we need to create at least
one .o file in the `spanner_client_testing` library before it works.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/297)

### DIFF
--- a/ci/kokoro/macos/build-bazel.sh
+++ b/ci/kokoro/macos/build-bazel.sh
@@ -16,7 +16,7 @@
 
 set -eu
 
-if [[ $# != 1 ]]; then
+if [[ $# -ne 1 ]]; then
   echo "Usage: $(basename "$0") <project-root>"
   exit 1
 fi
@@ -55,8 +55,6 @@ if [[ -n "${BAZEL_CONFIG}" ]]; then
     bazel_args+=("--config" "${BAZEL_CONFIG}")
 fi
 
-# The -DGRPC_BAZEL_BUILD is needed because gRPC does not compile on macOS unless
-# it is set.
 echo
 echo "================================================================"
 echo "Build and run unit tests at $(date)."
--- a/ci/kokoro/macos/build-bazel.sh
+++ b/ci/kokoro/macos/build-bazel.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+if [[ $# != 1 ]]; then
+  echo "Usage: $(basename "$0") <project-root>"
+  exit 1
+fi
+
+readonly PROJECT_ROOT="$1"
+
+echo
+echo "================================================================"
+echo "Update or Install Bazel at $(date)."
+echo
+
+# macOS does not have sha256sum by default, but `shasum -a 256` does the same
+# thing:
+function sha256sum() { shasum -a 256 "$@" ; } && export -f sha256sum
+
+"${PROJECT_ROOT}/ci/install-bazel.sh"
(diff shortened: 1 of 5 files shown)


ci/kokoro/macos/build-bazel.sh
```diff
@@ -0,0 +1,71 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+if [[ $# != 1 ]]; then
```
devbww: Should probably use an arithmetic operator (i.e., `-ne`) here.

ci/kokoro/macos/build-cmake.sh
```diff
@@ -0,0 +1,83 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+if [[ $# != 3 ]]; then
```
devbww: Ditto.

ci/kokoro/macos/build-bazel.sh
```diff
@@ -0,0 +1,71 @@
...
+    # unit tests do exactly that.
+    --action_env=GOOGLE_APPLICATION_CREDENTIALS="${GOOGLE_APPLICATION_CREDENTIALS}"
+    "--test_output=errors"
+    "--verbose_failures=true"
+    "--keep_going")
+if [[ -n "${BAZEL_CONFIG}" ]]; then
+    bazel_args+=("--config" "${BAZEL_CONFIG}")
+fi
+
+# The -DGRPC_BAZEL_BUILD is needed because gRPC does not compile on macOS unless
```
tmatsuo: Nit: the commnent seems outdated?

ci/kokoro/macos/build-cmake.sh
```diff
@@ -0,0 +1,83 @@
...
+readonly SOURCE_DIR="$2"
+readonly BINARY_DIR="$3"
+
+NCPU="$(sysctl -n hw.logicalcpu)"
+readonly NCPU
+
+source "${PROJECT_ROOT}/ci/colors.sh"
+
+echo
+echo "${COLOR_YELLOW}Starting docker build $(date) with $(nproc) cores${COLOR_RESET}"
```
tmatsuo: Nit: maybe remove?

ci/kokoro/macos/build-cmake.sh
```diff
@@ -0,0 +1,83 @@
...
+  echo "Running the integration tests $(date)"
+  echo "================================================================"
+  source "${KOKORO_GFILE_DIR}/spanner-integration-tests-config.sh"
+  export GOOGLE_APPLICATION_CREDENTIALS="${KOKORO_GFILE_DIR}/spanner-credentials.json"
+  export GOOGLE_CLOUD_CPP_AUTO_RUN_EXAMPLES=yes
+
+  # Run the integration tests too.
+  env -C "${BINARY_DIR}" ctest \
+      -L integration-tests \
+      --output-on-failure -j "${NCPU}"
```
tmatsuo: Nit: Probably it will never be a problem, but we may want to use multiple jobs even on single processor machine.

coryan: Probably, these tests are likely to be I/O bound and running more than the number of cores is probably a good optimization, but we need to have more than 2 or 3 of them before this is a worthy optimization.

ci/kokoro/macos/build-bazel.sh
coryan: Done.

ci/kokoro/macos/build-cmake.sh
coryan: Done.

ci/kokoro/macos/build-bazel.sh
coryan: Done.

ci/kokoro/macos/build-cmake.sh
coryan: Done.

PR #299, opened 2019-08-03
Reimplement `DateToString()` and `DateFromString()`

devbww:
The new implementation avoids internal/time_format.h, which is slated
for removal, the end goal being to produce `TimestampToString()` and
`TimestampFromString()` functions that can be shared between bigtable,
storage, and spanner libraries.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/299)

### DIFF
--- a/google/cloud/spanner/internal/date.cc
+++ b/google/cloud/spanner/internal/date.cc
@@ -24,7 +24,7 @@ inline namespace SPANNER_CLIENT_NS {
 namespace internal {
 
 std::string DateToString(Date d) {
-  std::array<char, sizeof "-9223372036854775807-01-01"> buf;
+  std::array<char, sizeof "-9223372036854775808-01-01"> buf;
   std::snprintf(buf.data(), buf.size(), "%" PRId64 "-%02d-%02d", d.year(),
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/internal/date.cc
```diff
@@ -13,42 +13,44 @@
...
-
-}  // namespace
-
 std::string DateToString(Date d) {
-  std::tm tm;
-  tm.tm_year = static_cast<int>(d.year() - 1900);
-  tm.tm_mon = d.month() - 1;
-  tm.tm_mday = d.day();
-  return FormatTime(kDateFormat, tm);
+  std::array<char, sizeof "-9223372036854775807-01-01"> buf;
```
remyabel: Not that it matters, but is `-9223372036854775807` supposed to represent `INT64_MIN`? Shouldn't this be `-9223372036854775808` then? Alternatively, put a comment for what this magic number is supposed to mean.

devbww: Yes, it should be.  Fixed.  Thanks!

PR #302, opened 2019-08-05
feat: Implement spanner::Client::Commit.

coryan:
This is an initial (and likely incomplete) implementation for
`spanner::Client::Commit()`.

The main rationale is that we will need a simple function like this
to test any other function that reads or modifies data. It also
introduces the first integration test that manipulates and reads
some data.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/302)


coryan:
The build breakage is unrelated, please take a look.

google/cloud/spanner/client.cc
```diff
@@ -99,16 +99,68 @@ StatusOr<std::int64_t> Client::ExecutePartitionedDml(
...
+  internal::Visit(
+      transaction, F([&request](spanner_proto::TransactionSelector& s) {
+        if (!s.id().empty()) {
+          request.set_transaction_id(s.id());
+        } else if (s.has_begin()) {
+          *request.mutable_single_use_transaction() = s.begin();
+        } else if (s.has_single_use()) {
+          *request.mutable_single_use_transaction() = s.single_use();
+        }
+        // TODO(#306) - Make Visit() work with functions that return `void`.
```
devbww: Typically you would need to do the `Commit()` in here, at least because in the `has_begin()` case you also need to `s.set_id()` with the returned transaction id.  And that means you typically want to return something as well.

coryan: Can you take another look and see if I am using the `Transaction` class as you intended?

google/cloud/spanner/client.cc
```diff
@@ -99,16 +99,68 @@ StatusOr<std::int64_t> Client::ExecutePartitionedDml(
...
+          *request.mutable_single_use_transaction() = s.single_use();
+        }
+        grpc::ClientContext context;
+        auto response = stub_->Commit(context, request);
+        if (!response) {
+          return std::move(response).status();
+        }
+        CommitResult r;
+        r.commit_timestamp = internal::FromProto(response->commit_timestamp());
+        return r;
```
devbww: > Can you take another look and see if I am using the Transaction class as you intended?

Yes.  A simplifying incantation might make `Client::Commit()` just a `Visit()` call forwarding to an implementation function.

```
StatusOr<CommitResult> Client::Commit(Transaction transaction,
                                      std::vector<Mutation> const& mutations) {
  return internal::Visit(std::move(transaction),
                         [this, mutations](spanner_proto::TransactionSelector& selector) {
                           return conn_->Commit(selector, mutations);
                         });
} 
```

I believe this will be more like the pattern for things like `Read()`, where there are three overloads forwarding to a single implementation, and where that implementation is a concrete sub-class of `Connection` (as proposed in #305).

`Commit()` is also a little easier than things like `Read()` because the latter can return a new transaction id that we need to stuff into `selector`, whereas `Commit()` never needs to do that (it doesn't even return a transaction id).

coryan: Ack. The incantation as-written did not work for me (see https://godbolt.org/z/WFB0MY), so I am explicitly casting the lambda to the right `std::function` for now.

coryan:
PTAL.

google/cloud/spanner/client.cc
```diff
@@ -99,16 +99,74 @@ StatusOr<std::int64_t> Client::ExecutePartitionedDml(
...
+  }
+  google::spanner::v1::CommitRequest request;
+  request.set_session(session->session_name());
+  for (auto const& m : mutations) {
+    *request.add_mutations() = m.as_proto();
+  }
+  if (!s.id().empty()) {
+    request.set_transaction_id(s.id());
+  } else if (s.has_begin()) {
+    *request.mutable_single_use_transaction() = s.begin();
```
devbww: I'm not sure why `CommitRequest` has a special `oneof transaction` instead of using `TransactionSelector` like everything else, apart from driving home the fact that "begin" really is "single-use" for commits.. So, I agree that this is the right thing to do.  For any actual read-write transaction (i.e., not just write) we will have already bound an id.

google/cloud/spanner/client_test.cc
```diff
@@ -124,6 +126,93 @@ TEST(SpannerClientTest, ReadTest) {
...
+              google::spanner::v1::CreateSessionRequest const& request) {
+            EXPECT_EQ(database_name, request.database());
+            return Status(StatusCode::kPermissionDenied, "uh-oh in GetSession");
+          }));
+  auto commit = client.Commit(MakeReadWriteTransaction(), {});
+  EXPECT_EQ(StatusCode::kPermissionDenied, commit.status().code());
+  EXPECT_THAT(commit.status().message(), HasSubstr("uh-oh in GetSession"));
+}
+
+/// @test Verify that errors when creating sessions are reported.
```
devbww: s/creating sessions/calling Commit()/?

google/cloud/spanner/client_test.cc
```diff
@@ -124,6 +126,93 @@ TEST(SpannerClientTest, ReadTest) {
...
+          [](grpc::ClientContext&,
+             google::spanner::v1::CommitRequest const& request) {
+            EXPECT_EQ("test-session-name", request.session());
+            EXPECT_EQ("test-txn-id", request.transaction_id());
+            return Status(StatusCode::kPermissionDenied, "uh-oh in Commit");
+          }));
+
+  auto txn = MakeReadWriteTransaction();
+  using F = std::function<int(spanner_proto::TransactionSelector&)>;
+  internal::Visit(txn, F([](spanner_proto::TransactionSelector& s) -> int {
```
devbww: Aside: This is like `client.Begin(txn)`.

google/cloud/spanner/client.cc
coryan: Ack. I take it that no changes are required.

google/cloud/spanner/client_test.cc
coryan: Fixed.

google/cloud/spanner/client_test.cc
coryan: Ack.

coryan:
PTAL.

PR #304, opened 2019-08-05
bug: Fix Bazel presubmit configuration.

coryan:
I missed the configuration file because it was captured by one of
the .gitignore globs. Add the configuration file and fixed .gitignore.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/304)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ cmake-out/
 # Common bazel output directories
 bazel-*
 !bazel-dependency*.cfg
+!bazel-*.cfg
 
 # Backup files for Emacs
 *~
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ cmake-out/
 # Common bazel output directories
 bazel-*
 !bazel-dependency*.cfg
```
devjgm: Do we still need the `!bazel-dependency*.cfg` line?

coryan: Nope, removed.

PR #313, opened 2019-08-05
feat: Generalize internal::Visit(Transaction).

coryan:
The function was using a `std::function<T(...)>` and made it hard to use
with lambdas. Since it was already a template (and its implementation
too), I decided to use the functor type all the way through.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/313)

### DIFF
--- a/google/cloud/spanner/transaction.h
+++ b/google/cloud/spanner/transaction.h
@@ -26,6 +26,7 @@ namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
+
 class Transaction;  // defined below
 
 // Internal forward declarations to befriend.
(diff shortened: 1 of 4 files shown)


google/cloud/spanner/transaction.h
```diff
@@ -20,23 +20,20 @@
...
 #include <google/spanner/v1/transaction.pb.h>
 #include <chrono>
-#include <functional>
 #include <memory>
 
 namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
-
```
devbww: How come your environment likes to delete these lines, but mine doesn't and the clang-tidy build doesn't complain about it?

coryan: Short answer is I do not know.  The longer answer: clang-format does not seem to care if you have 0 or 1 lines there. It cares if you have 2 or more, I suspect my setup at home just wants 0.

coryan:
PTAL.

Issue #285, opened 2019-08-02
Implement Client::ExecuteSql()


devjgm:
Implement the body and unit test for `Client::ExecuteSql`

See also: https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/client.h

Issue #283, opened 2019-08-02
Implement Client::Read()


devjgm:
Implement the body and unit test for `Client::Read`

See also: https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/client.h

PR #314, opened 2019-08-06
feat: Add a PartialResultSetReader class.

mr-salty:
This class serves as the glue between the gRPC PartialResultSet and the
spanner ResultSet (which allows C++-style row iteration).

needed for #283 (Read) and #285 (ExecuteSql) implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/314)


google/cloud/spanner/internal/partial_result_set_reader.h
```diff
@@ -0,0 +1,80 @@
...
+namespace internal {
+
+/**
+ * This class serves as a bridge between the gRPC `PartialResultSet` streaming
+ * reader and the spanner `ResultSet`, which is used to iterate over the rows
+ * returned from a read operation.
+ */
+class PartialResultSetReader : public internal::ResultSetSource {
+ public:
+  using GRpcReader =
```
devjgm: The count of how we spell "grpc" in google-cloud-cpp is:

```
     17 GRPC
     21 Grpc
     74 gRPC
   2151 grpc
```
It looks like `Grpc` is used most in code.

Do with this info as you will :-)

google/cloud/spanner/internal/partial_result_set_reader.cc
```diff
@@ -0,0 +1,135 @@
...
+    // We should only get stats once; if not, use the last one.
+    if (stats_.has_value()) {
+      GCP_LOG(WARNING) << "Unexpectedly received two sets of stats";
+    }
+    stats_ = std::move(*result_set.mutable_stats());
+  }
+
+  // TODO(#271) store and use resume_token.
+  // TODO(#270) handle chunked_value.
+  values_ = std::move(*result_set.mutable_values());
```
devjgm: Since the type of this is `RepeatedPtrField<X>`, does it actually do an efficient move here? I tried to look it up in proto docs, and got lost. Do you know?

devbww: My reading is that (if src and destination are on the same arena) move assignment is implemented as a swap, where the contents of a `RepeatedPtrField` are (after the arena pointer) just a size, a capacity, and a pointer to the data.

devjgm:
> oops, I missed Brad's comment... would you rather I keep it as Swap() or change it back to move assignment?

Whatever you decide is fine with me.

coryan:
Also, it looks like you need to rebase to pick up the changes for the macOS build.

PR #317, opened 2019-08-06
feat: add KeyRange and KeySet classes

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/317)

### DIFF
--- a/google/cloud/spanner/keys.h
+++ b/google/cloud/spanner/keys.h
@@ -35,7 +35,7 @@ template <typename... Ts>
 struct IsRow<Row<Ts...>> : std::true_type {};
 }  // namespace internal
 
-template <typename KeyType>
+template <typename RowType>
 class KeyRange;
 
 /**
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/keys.h
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  /**
+   * Creates a key `Row` from the values provided and adds it to the `KeySet`.
+   *
+   * @tparam ValueTypes
+   * @param values
+   */
+  template <typename... ValueTypes>
+  void AddKey(ValueTypes... values) {
```
devjgm: I think we want to do one of the following:

a. Delete this method. Users can call `MakeRow` themselves
b. Name this `EmplaceKey` to indicate that it's *not* taking a key, it's taking the ingredients of a Key.

I prefer (a) because I don't think the slight savings of typing is worth the slight increase in API complexity.

google/cloud/spanner/keys_test.cc
```diff
@@ -0,0 +1,171 @@
...
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace {
+
+TEST(KeyRangeTest, ConstructorBoundModeUnspecified) {
+  std::string start_value("key0");
+  std::string end_value("key1");
+  KeyRange<Row<std::string>> closed_range =
+      KeyRange<Row<std::string>>(MakeRow(start_value), MakeRow(end_value));
```
devjgm: I think we want a `MakeKeyRange` helper that deduces the RowType. Then this will become:
```cc
auto closed_range = MakeKeyRange(MakeRow(a, b, c), MakeRow(x, y, z));
```

PR #318, opened 2019-08-06
fix: Use 4-digit year for consistency

tmatsuo:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/318)


tmatsuo:
@remyabel Yeah we need a single object file for each repo, so I'm afraid that there's no easy and clean way.

PR #319, opened 2019-08-07
Make the C++ mappings of orderable Spanner types also orderable

devbww:
Add all relational operators to `Date` and `Value::Bytes`, so
that the ordering of each Spanner types is now reflected in the
ordering of its C++ mapping.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/319)

### DIFF
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -185,7 +185,7 @@ class Value {
    *
    * This struct is a thin wrapper around a `std::string` to distinguish BYTES
    * from STRINGs. The `.data` member can be set/get directly. Consructors and
-   * equality operators are provided for convenience and to make `Bytes` a
+   * relational operators are provided for convenience and to make `Bytes` a
    * regular type that is easy to work with.
    */
   struct Bytes {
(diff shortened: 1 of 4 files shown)


google/cloud/spanner/value.h
```diff
@@ -203,12 +203,18 @@ class Value {
...
     ///@}
 
-    /// @name Equality
+    /// @name Relational operators
     ///@{
     friend bool operator==(Bytes const& a, Bytes const& b) {
       return a.data == b.data;
     }
     friend bool operator!=(Bytes const& a, Bytes const& b) { return !(a == b); }
+    friend bool operator<(Bytes const& a, Bytes const& b) {
```
devjgm: I think less-than on a bytes data type is weird and the semantics are not obvious. Can we provide a link in the docs to where Spanner documents what this means?

devbww: As discussed offline, it appears there is no clear specification of Spanner type ordering (although it is acknowledged that there should be).  So, just added comments asserting that std::string comparison correctly models Spanner BYTES comparison, and updated the test case to cover the signed -v- unsigned char issue.

google/cloud/spanner/value.h
```diff
@@ -203,12 +203,20 @@ class Value {
...
+      // Note that std::string::operator<() behaves as if the chars were
+      // unsigned, which is exactly how Spanner BYTES values compare.
```
devjgm: This comment seems like it should be in the `Bytes` class docs for users to see.

devbww: Moved up.

Issue #196, opened 2019-07-11
Sample: Developers can "mock" our Client in order to write their own unit tests.


devjgm:
Developers must be able to easily write unit/integration tests for their code that utilize the Cloud Spanner C++ client. This suggests that they need some way to "mock" our Client library so that users can write their own tests.

PR #320, opened 2019-08-07
feat: Add Client -> Connection layer

### LINKED ISSUE #305: Consider: Factoring the spanner connection out of the Client
There are two potential issues that when considered together may suggest a slight alternative design to how we construct `Client` objects. The issues are:

### grpc::Channel reuse

I believe our plan today is to construct `Client` objects roughly as follows:

```cc
ClientOptions opts;   // [1]
Client c = MakeClient("/database/name", opts);  // [2]
```
The `ClientOptions` created in [1] above include the `grpc::ChannelCredentials` and the endpoint name (e.g., "spanner.googleapis.com"). Both of these fields have good defaults and are optional.

The call to `MakeClient` in [2] above results in the creation of a new `grpc::Channel`, which is expensive and should be reused as much as possible. We could maintain a channel-pool of our own to minimize the number of new channels we create, but there might be a better idea given the next point.

### Users need to be able to test with "mocks"

For the Alpha launch we need to enable developers to write their own unit tests with mocks of our client (See #196). The obvious solution for this is to make our `Client` class an interface with pure-virtual methods that are amenable to mocking. The problem with this is that virtual interfaces have some downsides in that they don't play well with overloading, nor templates, nor traditional "value semantics".

Another solution is to provide a non-virtual regular type named `Client` that users interact with (exactly as we do today), but then allow users to optionally *inject* in an object that implements some virtual interface that can be mocked. Indeed, this is pretty much exactly the plan we are doing today: https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/client.h#L361

FTR, I like this idea

The issues with today's plan are:

1. The injected interface is `internal::SpannerStub`, which as its name implies is "internal" and should not be used, implemented, or mocked by end users.

2. The API of [`SpannerStub`](https://github.com/googleapis/google-cloud-cpp-spanner/blob/master//google/cloud/spanner/internal/spanner_stub.h) is written in the vocabulary of the gRPC API, not the vocabulary of `Client`. This means that if users are two write their own tests, they need to learn the gRPC API in order to correctly mock `Client`. For example, `SpannerStub` talks about sessions, and "BeginTransaction", and gRPC protobufs, all of which we've tried to hide.

### Proposed Solution

Expose the concept of a "Connection", which is a logical connection to Spanner. It would encapsulate both a grpc::Channel and a the service-specific "stub". The Connection would be created with the users credentials and the endpoint name, both of which are optional, because there are good defaults. 

`spanner::Connection` would be an interface class that would be written completely in the vocabulary of `spanner::Client`. Users *could* mock the `spanner::Connection` class and inject it into their `Client` objects in order to write their own unit tests. It's interface would look something like:

```cc
class Connection {
 public:
  struct ReadParams {
    std::string table;
    KeySet keys;
    // ...
  };
  virtual StatusOr<ResultSet> Read(ReadParams const&) = 0;

  struct ExecuteSqlParams {
    SqlStatement statement;
    // ...
  };
  virtual StatusOr<ResultSet> ExecuteSql(ExecuteSqlParams const&) = 0;

  // ...
};
```

That is, there would be *one* pure-virtual method in this interfae for each logical operation provided by Client. For example, all of Client's `Read()` overloads would call through to the one `Read(ReadParams const&)` pure-virtual method in the Connection interface. This would give callers a single method to mock that would work with all of our convenient overloads.

The `Client` class would then be constructed with a "Connection". Users could create many `Client` objects all using the same "Connection" and this would be efficient and encouraged.

Example code:

```cc
// Optionally pass credential and endpoint to MakeConnection()
std::shared_ptr<spanner::Connection> conn = spanner::MakeConnection();
Client c1("/database/name", conn);
Client c2("/other/db", conn);  // OK, also uses conn
```

### FAQ

**What about the admin interface?**

That would be a different type of Connection class. Call it `spanner::AdminConnection`. But the exact same ideas would apply. It would work well and fit in nicely, I think.

**Are other Spanner libraries designed like this?**

Yes. For example, [C#](https://cloud.google.com/spanner/docs/reference/libraries#client-libraries-usage-csharp), [Java](https://cloud.google.com/spanner/docs/reference/libraries#client-libraries-usage-java), and maybe others.

**Why don't we just expose `SpannerStub` to users?**

Because it's an API written with the wrong vocabulary. It's a thin wrapper above gRPC, but that's an API that we're hiding from users. For example, our users don't care about "begin transaction", or "sessions", but we need `SpannerStub` to expose those methods for our *implementation*.

**Will we still need the `SpannerStub`?***

Probably, yes. That stub is a wrapper over the grpc-generated API, and providing that wrapper will help us unit test *our own* code. But it's an internal thing and will not help our users test their code.

### Steps

We've talked about this a fair bit in person in the office, but I'd like to hear others' opinions. (e.g., @coryan ). If we decide to do this, I think the steps would be something like this

- [ ] Define the `Connection` interface class. This should not be very difficult. It will basically be a single pure-virtual method for each overload set we expose in `Client`. To future-proof this interface, I think the methods on `Connection` should all take "params" structs instead of separate arguments, that way we can add more over time without breaking existing mocks. **effort: low**

- [ ] Implement a `internal::ConnectionImpl` subclass of `spanner::Connection` that implements each of the methods using the existing `internal::SpannerStub`. Initially, most of these methods will be noops, or TODOs, just like they are in `Client` today. **effort: medium**

- [ ] Expose a `spanner::MakeConnection()` method. It should return a `std::shared_ptr<spanner::Connection>` pointing to an `internal::ConnectionImpl` instance. It should optionally accept `grpc::Credential` and an endpoint string. **effort: low**

- [ ] Change `Client` to accept a `std::shared_ptr<spanner::Connection>` argument. Each of Client's methods should populate the params struct and call the pure-virtual method on the Connection object. **effort: med**

- [ ] Do the equivalent changes for the admin interface.

devjgm:
Fixes #305 
Related to #196 

This PR adds an abstract `Connection` interface that gives callers the ability to mock Client's behavior. This functionality is need for Alpha (#196). `Client` now holds a shared ptr to a `Connection` instance. And a `Connection` holds a shared pointer to a `SpannerStub`. See the more detailed design rationale in #305.

The main changes in this PR are that most of the *implementation* from `client.cc` moved to `internal/connection_impl.cc`. And the unit test stuff as well. This mostly affected `Commit`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/320)


google/cloud/spanner/internal/connection_impl.h
```diff
@@ -0,0 +1,77 @@
...
+    ConnectionImpl* conn_;
+  };
+  friend class SessionHolder;
+  StatusOr<SessionHolder> GetSession();
+
+  /// Implementation details for Commit.
+  StatusOr<CommitResult> Commit(google::spanner::v1::TransactionSelector& s,
+                                std::vector<Mutation> mutations);
+  std::string database_;
+  std::shared_ptr<internal::SpannerStub> stub_;
```
devbww: Why do we choose to make this shared?  gRPC gave it to us as a `unique_ptr`, which we converted into a `shared_ptr`, but (as far as I can see) its refcount remains at 1.

google/cloud/spanner/internal/connection_impl.h
```diff
@@ -0,0 +1,77 @@
...
+// and returns instances of this class.
+class ConnectionImpl : public Connection {
+ public:
+  // Creates a ConnectionImpl that will talk to the specified `database` using
+  // the given `stub`. We can test this class by injecting in a mock `stub`.
+  explicit ConnectionImpl(std::string database,
+                          std::shared_ptr<internal::SpannerStub> stub)
+      : database_(std::move(database)), stub_(std::move(stub)) {}
+
+  StatusOr<CommitResult> Commit(CommitParams) override;
```
devbww: Should we use `final`?

google/cloud/spanner/connection.h
```diff
@@ -0,0 +1,67 @@
...
+  virtual ~Connection() = default;
+
+  /**
+   * Commits a transaction.
+   *
+   * The commit might return an `ABORTED` error. This can occur at any time;
+   * commonly, the cause is conflicts with concurrent transactions. However, it
+   * can also happen for a variety of other reasons. If `Commit` returns
+   * `ABORTED`, the caller should re-attempt the transaction from the beginning,
+   * re-using the same session.
```
devbww: Is the "same session" part important or even necessary?

devbww: > It is important. Re-using the same session is how Spanner guarantees that the every Transaction will eventually "win" and be serviced.

Ack.  We mused about including that information, or an link, here.

google/cloud/spanner/internal/connection_impl.h
devbww: > Don't know. Maybe? I don't have a sense of when to apply `final`. Suggestion?

My only suggestions would be to forget about it, or apply it to every override (unless that contradicts some explicit need).  I'm happy with the former for now.

google/cloud/spanner/internal/connection_impl.h
devbww: > Don't know. I'm just copying existing code. Do you mind if we follow up on this change in a separate PR?

I'd suggest that it should be unique until it needs to be otherwise, but a separate PR is good.

mr-salty:
___
*[google/cloud/spanner/connection.h, line 46 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/320#-LlinKGa81ew-lMn_UWV:-LlinKGa81ew-lMn_UWW:bel7dfl) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/144c45f6c8bbb7df86c36e662ec132ae1ed18234/google/cloud/spanner/connection.h#L46)):*
<details><summary><i alt="summary">Quoted 8 lines of code…</i></summary>

> ```
>   /**
>    * Commits a transaction.
>    *
>    * The commit might return an `ABORTED` error. This can occur at any time;
>    * commonly, the cause is conflicts with concurrent transactions. However, it
>    * can also happen for a variety of other reasons. If `Commit` returns
>    * `ABORTED`, the caller should re-attempt the transaction from the beginning,
>    * re-using the same session.
> ```
</details>


I don't think we should duplicate the documentation from `Client` here.

mr-salty:
___
*[google/cloud/spanner/internal/connection_impl.h, line 64 at r3](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/320#-LlggtS9xHIoDLQEshUx-r3-64:-LlilI9k6EaGfNSlEOrc:by5ulo8) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/a6ba272a5ca7e5e3096db25790dcec1884aaf62b/google/cloud/spanner/internal/connection_impl.h#L64)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

> Don't know. I'm just copying existing code. Do you mind if we follow up on this change in a separate PR?

I'd suggest that it should be unique until it needs to be otherwise, but a separate PR is good.
</blockquote></details>

- This is not the gRPC stub; that is held as a `unique_ptr` inside `SpannerStub`.
- The design doc I was working from specified that `Stub` should be a `shared_ptr`. In general I didn't attempt to re-evaluate those decisions when I wrote the code; we can certainly revisit, but we should understand the rationale (perhaps it is no longer be applicable).

mr-salty:
Sorry for not commenting while the PR was open, I didn't realize it was on the verge of being submitted.

devjgm:
No problem. I'll do a follow-up PR addressing your comments.

PR #331, opened 2019-08-09
Provide an ordering to transaction visitors

### LINKED ISSUE #279: Design and implement sequence number management
DML operations (via `ExecuteSql` or `ExecuteBatchDmlRequest`) require a per-transaction sequence number. Ideally we can manage this for the user.

https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/spanner.proto#L441

> A per-transaction sequence number used to identify each request. This makes each request idempotent such that if the request is received multiple times, at most one will succeed.
>   
> The sequence number must be monotonically increasing within the transaction. If a request arrives for the first time with an out-of-order sequence number, the transaction may be aborted. Replays of previously handled requests will yield the same response as the first execution.
>   
> Required for DML statements. Ignored for queries.

devbww:
Pass a monotonically-increasing sequence number to each transaction
visitor.  This will provide a way to sequence read-write `ExecuteSql()`
calls.  See ExecuteSqlRequest.seqno for more details.

Fixes #279.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/331)

### DIFF
--- a/google/cloud/spanner/internal/transaction_impl.h
+++ b/google/cloud/spanner/internal/transaction_impl.h
@@ -49,7 +49,8 @@ class TransactionImpl {
   // passed TransactionSelector in its Client::Read()/Client::ExecuteSql()
   // call. If initially selector.has_begin(), and the operation successfully
   // allocates a transaction ID, then the functor should selector.set_id(id).
-  // Otherwise the functor should not modify the selector.
+  // Otherwise the functor should not modify the selector. A monotonically-
+  // increasing sequence number is also passed to the functor.
   template <typename Functor>
   VisitInvokeResult<Functor> Visit(Functor&& f) {
     std::int64_t seqno;
--- a/google/cloud/spanner/internal/transaction_impl.h
+++ b/google/cloud/spanner/internal/transaction_impl.h
@@ -22,6 +22,7 @@
 #include <condition_variable>
 #include <cstdint>
 #include <mutex>
+#include <type_traits>
 
 namespace google {
 namespace cloud {
@@ -39,7 +40,7 @@ using VisitInvokeResult = google::cloud::internal::invoke_result_t<
 class TransactionImpl {
  public:
   TransactionImpl(google::spanner::v1::TransactionSelector selector)
-      : selector_(std::move(selector)), seqno_(-1) {
+      : selector_(std::move(selector)), seqno_(0) {
     state_ = selector_.has_begin() ? State::kBegin : State::kDone;
   }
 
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/internal/transaction_impl.h
```diff
@@ -49,7 +49,8 @@ class TransactionImpl {
   // passed TransactionSelector in its Client::Read()/Client::ExecuteSql()
   // call. If initially selector.has_begin(), and the operation successfully
   // allocates a transaction ID, then the functor should selector.set_id(id).
-  // Otherwise the functor should not modify the selector.
+  // Otherwise the functor should not modify the selector. A monotonically-
+  // increasing sequence number is also passed to the functor.
   template <typename Functor>
```
coryan: Would this be a good place to use `std::enable_if<google::cloud::is_invocable<Functor, ...>>` and that acted the documentation for what are the parameters for `Functor`?

google/cloud/spanner/internal/transaction_impl.h
```diff
@@ -49,7 +49,8 @@ class TransactionImpl {
   // passed TransactionSelector in its Client::Read()/Client::ExecuteSql()
   // call. If initially selector.has_begin(), and the operation successfully
```
coryan: Not for this PR, but the API is slightly weird.  What if `Functor` had this prototype: `pair<T,std::string>(TransactionSelector const&, std::int64_t)`?  Or maybe even this `T(TransactionSelector const&, std::int64_t, std::string&)` so the implementation of `Visit` is in charge of calling `set_id`?

devbww:
> I'm not clear on whether 0 is a valid sequence number or if they should start at 1. The java implementation starts at 1.

FWIW, I originally started at 1, and then went to INT64_MIN, and then to 0.  But I'm happy with 1.

Done.

google/cloud/spanner/internal/transaction_impl.h
devbww: Yes.  A new requirement since the move to Functor, but done.  Thanks.

google/cloud/spanner/internal/transaction_impl.h
devbww: The visitor functor often doesn't need to set the transaction id (it may be a single use transaction, or a single-operation commit, or the typical case of the id already being bound).  So it is difficult for the `Visit` implementation to know when to do so, and building a protocol for the functor to tell it when seems much more complicated than just leaving it to the functor.

PR #334, opened 2019-08-09
feat: let clang-tidy build not fail if there are edited files.

devjgm:
This change passes an env var to the docker builds to let them know if
they are running as part of a CI or a human invocation. In the former
case, there is no change. In the latter case, some scripts may want to
be a little more friendly. For example, in this PR, I changed the
clang-tidy build to not bail if there are existing uncommitted files.
Instead, the the clang-tidy build will simply format the local files and
leave them edited in the working dir. The author can then look at the
diffs and commit them. This makes "build.sh clang-tidy" a more useful
tool for developers who want to use it to format their existing code.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/334)

### DIFF
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -35,6 +35,13 @@ elif [[ -n "${KOKORO_JOB_NAME:-}" ]]; then
   # name.
   BUILD_NAME="$(basename "${KOKORO_JOB_NAME}" "-presubmit")"
   export BUILD_NAME
+
+  # This is passed into the environment of the docker build and its scripts to
+  # tell them if they are running as part of a CI build rather than just a
+  # human invocation of "build.sh <build-name>". This allows scripts to be
+  # strict when run in a CI, but a little more friendly when run by a human.
+  RUNNING_CI="yes"
+  export RUNNING_CI
 else
  echo "Aborting build as the build name is not defined."
  echo "If you are invoking this script via the command line use:"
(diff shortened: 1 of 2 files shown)


ci/kokoro/docker/build.sh
```diff
@@ -237,6 +244,9 @@ docker_flags=(
     # Additional flags to enable CMake features.
     "--env" "CMAKE_FLAGS=${CMAKE_FLAGS:-}"
 
+    # Tells scripts whether they are running as part of a CI or not.
+    "--env" "RUNNING_CI=${RUNNING_CI:-no}"
```
remyabel: Wouldn't the existing KOKORO environment variables suffice to say "we are running in CI" since we don't use any other type of CI?

devjgm: For today, it would. But my thinking went as follows:

1. The `KOKORO_JOB_NAME` var isn't passed to the docker env, so we'd still need to pass it (or something else)

2. `KOKORO_JOB_NAME` contains some specifically formatted string, so scripts would either need to know the format of that string.

3.`RUNNING_CI` is very specific, easy for scripts to parse, and would work if we ever had a CI system other than kokoro.

*shrug*

Convincing? :-)

remyabel: Yup, SGTM. I misread the PR by reading too quickly.

PR #335, opened 2019-08-09
Add a ToProto() method for SqlStatement.

mr-salty:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/335)

### DIFF
--- a/google/cloud/spanner/sql_statement_test.cc
+++ b/google/cloud/spanner/sql_statement_test.cc
@@ -13,13 +13,18 @@
 // limitations under the License.
 
 #include "google/cloud/spanner/sql_statement.h"
+#include "google/cloud/spanner/testing/matchers.h"
+#include <google/protobuf/text_format.h>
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/sql_statement_test.cc
```diff
@@ -105,6 +110,45 @@ TEST(SqlStatementTest, Equality) {
...
+}
+
+TEST(SqlStatementTest, ToProtoWithParams) {
+  SqlStatement::ParamType params = {
+      {"last", Value("Blues")},
+      {"first", Value("Elwood")},
+      {"destroyed_cars", Value(std::int64_t{103})},
+  };
+
+  const char* sql =
```
devjgm: Nit: since you don't actually care about the type of `sql`, meaning you don't care if it's an array or if it has been decayed to a pointer, consider just `auto sql = "....";`

PR #340, opened 2019-08-12
feat: add type erased KeySet, KeySetBuilder::Build, and ToProto(KeySet)

scotthart:
Additionally, KeySetBuilder is no longer necessary to build an all keys KeySet. That can be done directly from KeySet::All.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/340)


google/cloud/spanner/keys.h
```diff
@@ -202,8 +246,65 @@ class KeySet {
...
+  template <typename RowType>
+  friend class KeySetBuilder;
+
+  friend ::google::spanner::v1::KeySet internal::ToProto(KeySet const& keyset);
+
   explicit KeySet(bool all) : all_(all) {}
 
+  class ValueRow {
+   public:
+    template <std::size_t SIZE>
```
devjgm: I think non-type template params should be spell like `Size`, just like template params. But in this case, I think I'd just avoid the situation and use `N`.

google/cloud/spanner/keys.h
```diff
@@ -202,8 +246,65 @@ class KeySet {
...
+    explicit ValueRow(std::array<Value, SIZE> const& column_values)
+        : column_values_(
+              std::vector<Value>(column_values.begin(), column_values.end())) {}
```
devjgm: I think the explicit call to vector's constructor here is redundant and can probably be removed?

google/cloud/spanner/keys.h
```diff
@@ -202,8 +246,65 @@ class KeySet {
...
+
+    ValueBound const& start() const { return start_; }
+    ValueBound const& end() const { return end_; }
+
+   private:
+    ValueBound start_;
+    ValueBound end_;
+  };
+
+  KeySet& EmplaceKey(ValueRow key) {
```
devjgm: I think all of the "emplace" things in this file should really be "push_back", or in the vocabulary of the KeySet API, maybe name them `AddRow`.

The reason is that "emplace" methods are mostly useful when constructing an object in place from its ingredients. Like, `persons.emplace_back("Jerry", "Seinfeld", Occupation::kComedian)`. However, in all of the cases in this file, we have an already constructed object and we're simply adding it to the container. `vector::push_back` is already aware of rvalues and is move-efficient, so I don't think there's any performance gain to be had with emplacing.

See also https://abseil.io/tips/112

google/cloud/spanner/internal/keys.h
```diff
@@ -0,0 +1,35 @@
...
+
+::google::spanner::v1::KeySet ToProto(KeySet const& keyset);
+
+}  // namespace internal
+}  // namespace SPANNER_CLIENT_NS
+}  // namespace spanner
+}  // namespace cloud
+}  // namespace google
+
+#endif  // GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_INTERNAL_KEYS_H_
```
devbww: No newline at end of file.

scotthart: Done.

google/cloud/spanner/keys.h
```diff
@@ -202,8 +246,65 @@ class KeySet {
...
+    template <std::size_t SIZE>
+    explicit ValueRow(std::array<Value, SIZE> const& column_values)
+        : column_values_(
+              std::vector<Value>(column_values.begin(), column_values.end())) {}
+
+    std::vector<Value> const& column_values() const { return column_values_; }
+
+   private:
+    std::vector<Value> column_values_;
+  };
```
devjgm: I wonder if this `ValueRow` class can be an alias like `using ValueRow = std::vector<Value>;`, and move the `std::array` -> `std::vector` conversion logic to the "AddRow" function, like:

```cc
using ValueRow = std::vector<Value>;
template <std::size_t N>
void AddRow(std::array<Value, N> a) {
  key_values_.emplace_back(
      std::make_move_iterator(a.begin()),
      std::make_move_iterator(a.end()));
}
```

One benefit of this would be that the `KeySetBuilder` class does not need to mention the `ValueRow` type. Instead, it simply passes its array to `AddRow` and the `KeySet` handles the conversion.

I haven't looked at the other internal classes, but maybe they could be implemented similarly?

What do you think?

google/cloud/spanner/internal/keys.h
```diff
@@ -0,0 +1,35 @@
...
+#include "google/cloud/spanner/keys.h"
+#include <google/spanner/v1/keys.pb.h>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+
+::google::spanner::v1::KeySet ToProto(KeySet const& keyset);
```
devbww: Is there (pass-by) value in having this "consume" the KeySet?

google/cloud/spanner/internal/keys.cc
```diff
@@ -0,0 +1,69 @@
...
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+
+::google::spanner::v1::KeySet ToProto(KeySet const& keyset) {
+  ::google::spanner::v1::KeySet proto;
+  proto.set_all(keyset.IsAll());
```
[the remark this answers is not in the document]
scotthart: Maybe. 
- Assuming that spanner doesn't and never will look at any keys or key ranges present in a proto, we could skip adding them to the proto, even if they are present.
- If both keys and key ranges are empty, the function will do little extra work as written.

google/cloud/spanner/keys.h
```diff
@@ -24,20 +24,28 @@
 #include <vector>
 
 namespace google {
+namespace spanner {
+namespace v1 {
+class KeySet;
```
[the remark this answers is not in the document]
scotthart: Fair enough, was hoping to avoid including the keys.pb.h file just for this.

google/cloud/spanner/keys.h
scotthart: Done.

google/cloud/spanner/keys.h
scotthart: Done.

google/cloud/spanner/keys.h
scotthart: Done.

google/cloud/spanner/internal/keys.h
scotthart: I could see a call like 
```
auto keyset_proto = internal::ToProto(keyset_builder.Build());
```
being useful.
Done.

google/cloud/spanner/keys.h
scotthart: So right now, KeySet has a very small public API and all its nested classes are private. ValueRow does not show up in KeySetBuilder's API at all, only inside the implementation of Build(). I'm not sure I understand what the value proposition is in this case.

google/cloud/spanner/keys.h
```diff
@@ -37,7 +33,7 @@ template <typename RowType>
 class KeyRange;
 
 namespace internal {
-::google::spanner::v1::KeySet ToProto(KeySet const& keyset);
+::google::spanner::v1::KeySet ToProto(KeySet keyset);
```
devbww: Just include "internal/keys.h" instead?

google/cloud/spanner/keys.h
scotthart: Refactored such that KeySetBuilder (as a friend) only uses member functions of KeySet and not its internal types.

google/cloud/spanner/keys.h
scotthart: On general principle, I'd prefer not to include an internal header in a public header if I am able.

google/cloud/spanner/keys.h
scotthart: Extended refactoring to remove KeySetBuilder as a friend and added a templated constructor to KeySet.

google/cloud/spanner/keys.h
devbww: I would proffer that we should prefer to include the internal header over forward declaring its contents.  We're including it for implementation-detail reasons, so the include is an implementation detail too.  For what it worth, there are already nine internal-header includes in g/c/s/*.h.

scotthart: Forward declaration resolved by removing internal header.

PR #344, opened 2019-08-12
Add CMakeLists.txt.tmp to .gitignore

mr-salty:
I see this happen fairly often, and sometimes even accidentally add it:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	CMakeLists.txt.tmp
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/344)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ google/cloud/storage/testbench/*.pyc
 
 # Ignore Visual Studio Code files.
 .vscode/
+
+# Temporary cmake file.
+CMakeLists.txt.tmp


mr-salty:
alternatively, is there a way to prevent it, or have the file end up in somewhere that's already ignored like cmake-out?  I have no idea what I do to cause it, it may be related to stash/unstash?  or hitting ctrl-c at the wrong time?

remyabel:
> alternatively, is there a way to prevent it, or have the file end up in somewhere that's already ignored like cmake-out? I have no idea what I do to cause it, it may be related to stash/unstash? or hitting ctrl-c at the wrong time?

`tmp` files are created by `check-style`. It's probable that you interrupted it during that stage. I don't think ignoring them is a good idea because then it won't show up in `git status` and they should probably be deleted before running `check-style` again.

remyabel:
Building on your suggestion, but what about this idea: we can put the tmp files in some already ignored work directory (like `cmake-out`) then delete all the `tmp` files at the start of `check-style.sh`.

mr-salty:
I'd be fine with putting the files in cmake-out, I can work on that.

PR #349, opened 2019-08-13
Enable clang-format proto raw string formatting.

mr-salty:
I also changed the quoting of the existing proto strings, and expanded the repeated fields and maps to use repeated field names rather than the bracket syntax, which clang-format does not seem to handle well. Then I let clang-format reformat them all.

`EnclosingFunctions` is supposed to make this "automatic" based on the function name, i.e. you can just write `R"(` rather than `R"proto(` when calling one of those functions. I decided not to use it since it appears to only look at the function name itself and I thought `ParseFromString` was overly broad. I tried to set it to `TextFormat::ParseFromString` (and the fully qualified name) but those didn't work.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/349)

### DIFF
--- a/google/cloud/spanner/keys_test.cc
+++ b/google/cloud/spanner/keys_test.cc
@@ -184,7 +184,8 @@ TEST(InternalKeySetTest, ToProtoAll) {
   ::google::spanner::v1::KeySet expected;
   EXPECT_TRUE(google::protobuf::TextFormat::ParseFromString(
       R"pb(
-        all: true)pb",
+        all: true
+      )pb",
       &expected));
 
   ::google::spanner::v1::KeySet result = internal::ToProto(ks);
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/keys_test.cc
```diff
@@ -200,27 +199,17 @@ TEST(InternalKeySetTest, BuildToProtoTwoKeys) {
...
+      R"pb(
+        keys: {
+          values: { string_value: "foo0" }
+          values: { string_value: "bar0" }
+        }
+        keys: {
+          values: { string_value: "foo1" }
+          values: { string_value: "bar1" }
+        }
+        all: false)pb",
```
devjgm: It looks weird without a newline before the closing `)pb".`

But if that's what it does, so be it.

PR #350, opened 2019-08-14
feat: add Read and ExecuteSql methods

### LINKED ISSUE #283: Implement Client::Read()

### LINKED ISSUE #285: Implement Client::ExecuteSql()

mr-salty:
I'm still planning to add more tests, but this has basic unit tests for `Read` and `ExecuteSql` and I believe the implementation of both is complete. I think it would also be valuable to get some samples and integration tests written, which is blocked by checking this in.

Fixes #283 
Fixes #285 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/350)

### DIFF
--- a/google/cloud/spanner/connection.h
+++ b/google/cloud/spanner/connection.h
@@ -15,7 +15,7 @@
 #ifndef GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_CONNECTION_H_
 #define GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_CONNECTION_H_
 
-#include "google/cloud/spanner/client_options.h"  // XXX(salty)
+#include "google/cloud/spanner/client_options.h"
 #include "google/cloud/spanner/commit_result.h"
 #include "google/cloud/spanner/keys.h"
 #include "google/cloud/spanner/mutations.h"
--- a/google/cloud/spanner/connection.h
+++ b/google/cloud/spanner/connection.h
@@ -15,11 +15,16 @@
 #ifndef GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_CONNECTION_H_
 #define GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_CONNECTION_H_
 
+#include "google/cloud/spanner/client_options.h"  // XXX(salty)
 #include "google/cloud/spanner/commit_result.h"
+#include "google/cloud/spanner/keys.h"
 #include "google/cloud/spanner/mutations.h"
+#include "google/cloud/spanner/result_set.h"
+#include "google/cloud/spanner/sql_statement.h"
 #include "google/cloud/spanner/transaction.h"
 #include "google/cloud/spanner/version.h"
 #include "google/cloud/status_or.h"
+#include <string>
 #include <vector>
 
 namespace google {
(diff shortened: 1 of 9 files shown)


google/cloud/spanner/connection.h
```diff
@@ -43,6 +48,21 @@ class Connection {
  public:
   virtual ~Connection() = default;
 
+  struct ReadParams {
```
devbww: Perhaps `Read()` and `ExecuteSql()` comments like those on `Commit()` and `Rollback()`?

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -54,15 +71,83 @@ StatusOr<ConnectionImpl::SessionHolder> ConnectionImpl::GetSession() {
...
+  auto reader = internal::PartialResultSetReader::Create(
+      stub_->StreamingRead(context, request));
+  if (!reader.ok()) {
+    return std::move(reader).status();
+  }
+  if (s.has_begin()) {
+    auto metadata = (*reader)->Metadata();
+    if (!metadata || metadata->transaction().id().empty()) {
+      return Status(StatusCode::kInternal,
+                    "begin transaction requested but no transaction returned");
```
coryan: Do you think we should include the operation (maybe `__func__` or `Client::Read`) in the error message?

google/cloud/spanner/internal/connection_impl.h
```diff
@@ -59,11 +62,18 @@ class ConnectionImpl : public Connection {
...
+                           ReadParams rp);
+
+  /// Implementation details for ExecuteSql
+  StatusOr<ResultSet> ExecuteSql(google::spanner::v1::TransactionSelector& s,
+                                 std::int64_t seqno, ExecuteSqlParams esp);
+
   /// Implementation details for Commit.
   StatusOr<CommitResult> Commit(google::spanner::v1::TransactionSelector& s,
-                                std::vector<Mutation> mutations);
+                                CommitParams cp);
```
coryan: This change seems unrelated, other than it is more consistent.  Maybe split to a separate PR?

google/cloud/spanner/connection.h
mr-salty: I was trying to convince @devjgm that we should get rid of (all) the comments here and just defer to `client.h`.  The class-level comment already hints at that, and the existing ones are partial copies of comments in `client.h`.  But I'll go with the will of the people.

devbww: I'm happy to go in that direction (no comments here) too.

mr-salty:
___
*[google/cloud/spanner/client_test.cc, line 55 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/350#-LmFDTOD0Xu1TPle1m4A:-LmG2YYa3tDQycfEIh8s:b3yg1wx) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/fc7c81b02d0760a680f926fdd5d8d949b9e9e823/google/cloud/spanner/client_test.cc#L55)):*

FYI: because `MOCK_METHOD<N>()` does not use `override` adding any other function with `override` produces clang-tidy errors.
</blockquote></details>

hm... although we weren't seeing that (unless the warning in question is turned off?)

devbww:
>> Is this defaulted override of the destructor needed?

> no, I probably copied from Connection (which doesn't need it either)

I would suggest that `Connection` (at least) should have a non-defaulted, non-inline destructor in order to avoid the "-Wweak-vtables" issue.

mr-salty:
To be clear I meant to say `MockConnection` didn't need `~MockConnection() override = default`. Do we want to change `Connection` to do what you suggest (and perhaps other classes?), in another PR?

PR #351, opened 2019-08-14
cleanup: Simplify (future) releases.

coryan:
Like we did in google-cloud-cpp, there is no need to modify the code to
make a release. The `IsRelease()` function and friends is gone, and the
build metadata is added if present.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/351)


google/cloud/spanner/internal/build_info_test.cc
```diff
@@ -30,12 +30,6 @@ TEST(BuildInfo, BuildFlags) {
   EXPECT_THAT(bf, Not(HasSubstr("@")));
 }
 
-TEST(BuildInfo, IsRelease) {
-  bool const b = IsRelease();
-  // We want to test this, but either value is fine.
-  EXPECT_TRUE(b || !b);
```
devjgm: Happy to be removing this test :-)

PR #354, opened 2019-08-15
chore: Small style cleanup to keys.h

devjgm:
A few small cleanups to `keys.h`

* Renamed enum from MACRO_STYLE to kConstant style, per the style guide.

* Use some existing factory functions and removed the need for some friendships.

* Tweak to `KeySetBuilder` docs.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/354)


coryan:
> * Renamed enum from MACRO_STYLE to kConstant style, per the style guide.

Just FYI: this should have been detected by clang-tidy, and if it was not then we have a bug in our clang-tidy config file. Created #355 to track that.

PR #358, opened 2019-08-15
Implement spanner::RunTransaction() skeleton

### LINKED ISSUE #356: Implement spanner::RunTransaction() Skeleton
Read-write transactions can be tricky to use because aborted commits are expected and require that the whole transaction be retried. Therefore, we want to provide a "Transaction Runner" (which is a function named `RunTransaction()`) for use with read-write transactions.

The following is an outline of what this function might look like:
```
namespace google {
namespace cloud {
namespace spanner {

struct TransactionAction {
  enum { kCommit, kRollback } action;
  std::vector<Mutation> mutations;
};
StatusOr<CommitResult> RunTransaction(
    Client c,
    Transaction::ReadWriteOptions opts,
    std::function<TransactionAction(Client, Transaction)> f);

}  // namespace spanner
}  // namespace cloud
}  // namespace google
```

devbww:
Fixes #356.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/358)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -128,6 +128,37 @@ std::shared_ptr<Connection> MakeConnection(
                                                     std::move(stub));
 }
 
+StatusOr<CommitResult> RunTransaction(
+    Client client, Transaction::ReadWriteOptions const& opts,
+    std::function<TransactionAction(Client, Transaction)> const& f) {
+  Transaction txn = MakeReadWriteTransaction(opts);
+  TransactionAction action;
+#if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
+  try {
+#endif
+    action = f(client, txn);
+#if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
+  } catch (...) {
+    client.Rollback(txn);  // ignore error
+    throw;
+  }
+#endif
+  switch (action.action) {
+    case TransactionAction::kCommit: {
+      // TODO(#357): Automatically retry if Commit() aborts. Note: it is
+      // not a good idea to simply cap the number of retries. Instead, it
+      // is better to limit the total amount of wall time spent retrying.
+      return client.Commit(txn, action.mutations);
+    }
+    case TransactionAction::kRollback: {
+      auto status = client.Rollback(txn);
+      if (!status.ok()) return status;
+      return CommitResult{};  // TODO(#357): What should this return?
+    }
+  }
+  return Status(StatusCode::kInvalidArgument, "bad TransactionAction");
+}
+
 }  // namespace SPANNER_CLIENT_NS
 }  // namespace spanner
 }  // namespace cloud
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -29,6 +29,7 @@
 #include "google/cloud/status_or.h"
 #include <google/spanner/v1/spanner.pb.h>
 #include <grpcpp/grpcpp.h>
+#include <functional>
 #include <memory>
 #include <string>
 #include <vector>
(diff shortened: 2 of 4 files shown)


google/cloud/spanner/client.cc
```diff
@@ -128,6 +128,37 @@ std::shared_ptr<Connection> MakeConnection(
...
+    action = f(client, txn);
+#if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
+  } catch (...) {
+    client.Rollback(txn);  // ignore error
+    throw;
+  }
+#endif
+  switch (action.action) {
+    case TransactionAction::kCommit: {
+      // TODO(#357): Automatically retry if Commit() aborts. Note: it is
```
coryan: FYI: we should let a `RetryPolicy` control how many times and/or for how we retry. We already have such policies in the AdminClient.

google/cloud/spanner/client.cc
```diff
@@ -128,6 +128,37 @@ std::shared_ptr<Connection> MakeConnection(
...
+      // not a good idea to simply cap the number of retries. Instead, it
+      // is better to limit the total amount of wall time spent retrying.
+      return client.Commit(txn, action.mutations);
+    }
+    case TransactionAction::kRollback: {
+      auto status = client.Rollback(txn);
```
coryan: Rollbacks can also fail and may need to be retried.  They are easier to retry, of course.

devbww:
> I would like to see at least one integration tests working as part of this PR.

Done.

google/cloud/spanner/client.cc
devbww: Acknowledged.

google/cloud/spanner/client.cc
devbww: Are you referring to "RPC-level" retry that we should apply to all remote calls, or something different?

I don't know that `Rollback()` can ever return a server-initiated "aborted" like `Commit()`, which is what we're dealing with in this particular retry loop.  Indeed, I'm hard pressed to think of any way a server-initiated `Rollback()` error could be retried.  The transaction has probably already been rolled back.

google/cloud/spanner/client.cc
```diff
@@ -128,6 +128,37 @@ std::shared_ptr<Connection> MakeConnection(
...
 
+StatusOr<CommitResult> RunTransaction(
+    Client client, Transaction::ReadWriteOptions const& opts,
+    std::function<TransactionAction(Client, Transaction)> const& f) {
+  Transaction txn = MakeReadWriteTransaction(opts);
+  TransactionAction action;
+#if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
+  try {
+#endif
+    action = f(client, txn);
```
devjgm: As discussed briefly offline, I think the caller may be able to give us a std::function that accepts these args as *non-const* references. And if that's the case, the caller's functor *could* change, or even move-from the `client` and `txn` arg. This would be a problem if we needed to commit or rollback the transaction, or in a retry situation, reusing the `client`.

So, I think an easy fix would be to always pass the copies here like:

```cc
action = f(Client(client), Transaction(txn));
```

google/cloud/spanner/client.cc
coryan: Yes, I was referring to the RPC-level retry.  A `Rollback` that returns `StatusCode::kUnavailable` is a (I think): not an operation that is known to have completed successfully (or with a permanent error), and it can be retried safely.

devbww: Acknowledged.

google/cloud/spanner/client.cc
devbww: As also briefly discussed offline, we're convinced that an argument `std::function` that accepts its args as non-const references would not be convertible to the `RunTransaction()` parameter.

PR #359, opened 2019-08-15
cleanup: Enable clang-tidy check for header files.

### LINKED ISSUE #355: Enable clang-tidy on our headers
Apparently we are not enforcing the `kCamelCase` naming convention for them.

remyabel:
Fixes #355

It appears by default that clang-tidy doesn't check (all?) header files.  This
PR simply enables it. I also run the automated fixer on all files except
`transaction.h`, which generates a "The new replacement overlaps with an
existing replacement." error. Since I simply ran
`/usr/share/clang/run-clang-tidy.py -fix` it is trivial for me to update this
PR if any fixes are undesirable.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/359)


coryan:
I think clang-tidy has too many false positives here.  We need to do do a lot of fine-tuning of the clang-tidy configuration.  I think we also need to decide on the naming convention for functions because as it is, it is unenforceable.

devjgm:
Maybe we can disable the clang-tidy checks for function naming.

And if there's a way to make it east-const friendly, that'd be nice too.

remyabel:
I am closing this for now. Some the checks don't seem possible to fix, particularly the include guards. Maybe somebody can ask on the mailing list.

PR #365, opened 2019-08-15
feat: add KeySet equality operators and constructor from proto

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/365)

### DIFF
--- a/google/cloud/spanner/keys.cc
+++ b/google/cloud/spanner/keys.cc
@@ -35,6 +35,10 @@ namespace internal {
   return std::move(keyset.proto_);
 }
 
+KeySet FromProto(::google::spanner::v1::KeySet keyset) {
+  return KeySet(std::move(keyset));
+}
+
 }  // namespace internal
 }  // namespace SPANNER_CLIENT_NS
 }  // namespace spanner
--- a/google/cloud/spanner/keys.cc
+++ b/google/cloud/spanner/keys.cc
@@ -14,8 +14,8 @@
 
 #include "google/cloud/spanner/keys.h"
 #include "google/cloud/spanner/value.h"
-#include <google/spanner/v1/keys.pb.h>
 #include <google/protobuf/util/message_differencer.h>
+#include <google/spanner/v1/keys.pb.h>
 
 namespace google {
 namespace cloud {
@@ -23,17 +23,14 @@ namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
 
 bool operator==(KeySet const& lhs, KeySet const& rhs) {
-//    std::string delta;
+  //    std::string delta;
   google::protobuf::util::MessageDifferencer differencer;
-//    differencer.ReportDifferencesToString(&delta);
+  //    differencer.ReportDifferencesToString(&delta);
   auto const result = differencer.Compare(lhs.proto_, rhs.proto_);
   return result;
 }
 
-bool operator!=(KeySet const& lhs, KeySet const& rhs) {
-  return !(lhs == rhs);
-}
-
+bool operator!=(KeySet const& lhs, KeySet const& rhs) { return !(lhs == rhs); }
 
 namespace internal {
 
--- a/google/cloud/spanner/keys.cc
+++ b/google/cloud/spanner/keys.cc
@@ -23,11 +23,8 @@ namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
 
 bool operator==(KeySet const& lhs, KeySet const& rhs) {
-  //    std::string delta;
   google::protobuf::util::MessageDifferencer differencer;
-  //    differencer.ReportDifferencesToString(&delta);
-  auto const result = differencer.Compare(lhs.proto_, rhs.proto_);
-  return result;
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/keys.cc
```diff
@@ -14,12 +14,24 @@
...
 
 namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
+
+bool operator==(KeySet const& lhs, KeySet const& rhs) {
+  //    std::string delta;
+  google::protobuf::util::MessageDifferencer differencer;
+  //    differencer.ReportDifferencesToString(&delta);
```
devjgm: I'd remove these two commented lines since their not being used and we don't really have a place to log the `delta` here anyway.

google/cloud/spanner/keys.cc
```diff
@@ -14,12 +14,24 @@
...
+  auto const result = differencer.Compare(lhs.proto_, rhs.proto_);
+  return result;
```
devjgm: consider just `return differencer.Compare(...);`, since we don't need the intermediate `result` variable.

google/cloud/spanner/keys.h
```diff
@@ -258,6 +264,13 @@ class KeySet {
     }
   }
 
+  /// @name Equality
+  /// Order is of keys and key ranges in the `KeySet` is considered.
```
devjgm: The wording on the "ordering..." comment is a little funny. Maybe remove "is"?

google/cloud/spanner/keys.cc
```diff
@@ -14,12 +14,24 @@
...
+#include <google/protobuf/util/message_differencer.h>
 #include <google/spanner/v1/keys.pb.h>
 
 namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
+
+bool operator==(KeySet const& lhs, KeySet const& rhs) {
+  //    std::string delta;
```
coryan: Remove the comment?

google/cloud/spanner/keys.cc
coryan: Same here?

google/cloud/spanner/keys.cc
```diff
@@ -14,12 +14,24 @@
...
+  google::protobuf::util::MessageDifferencer differencer;
+  //    differencer.ReportDifferencesToString(&delta);
+  auto const result = differencer.Compare(lhs.proto_, rhs.proto_);
```
coryan: just `return differencer.Compare(lhs.proto_, rhs.proto_);`

google/cloud/spanner/keys_test.cc
```diff
@@ -48,6 +48,46 @@ TEST(KeySetTest, AllKeys) {
   EXPECT_THAT(result, spanner_testing::IsProtoEqual(expected));
 }
 
+TEST(KeySetTest, EqualityEmpty) {
+  KeySet expected;
+  KeySet actual;
+  EXPECT_TRUE(expected == actual);
```
devjgm: I think it's good to prefer `EXPECT_EQ` and `EXPECT_NE`, because they have the benefit of strinifying the arguments when the expectations fail. We wouldn't get that with a failed `EXPECT_TRUE`

google/cloud/spanner/keys.cc
scotthart: Forgot I had left all that cruft in here. That's what I get for being in a hurry.

google/cloud/spanner/keys.cc
scotthart: Done.

google/cloud/spanner/keys.cc
scotthart: Done.

google/cloud/spanner/keys.cc
scotthart: Done.

google/cloud/spanner/keys.h
scotthart: Done.

google/cloud/spanner/keys_test.cc
scotthart: Done.

google/cloud/spanner/keys.cc
scotthart: Done.

google/cloud/spanner/keys.h
```diff
@@ -222,6 +222,12 @@ class KeySet {
    */
   KeySet() = default;
 
+  /**
+   * Constructs a `KeySet` from a `spanner::v1::KeySet` protobuf.
+   */
+  explicit KeySet(google::spanner::v1::KeySet key_set)
```
devjgm: As we discussed offline, we probably want to remove this public constructor and instead have an internal::FromProto to match the existing ToProto function.

... Or you could remove this constructor from this PR since it's not related to the equality operations being added here. 

Your call.

scotthart: Added FromProto and made constructor private.

PR #366, opened 2019-08-15
feat: Integration tests for Rollback().

coryan:
Wrote an integration test for Client::Rollback(). To do so, I also wrote
a test that performed multiple SQL statements in a single transaction.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/366)

### DIFF
--- a/google/cloud/spanner/integration_tests/mutate_and_read_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/mutate_and_read_integration_test.cc
@@ -196,6 +196,127 @@ TEST_F(MutateAndReadIntegrationTest, DeleteAndCommit) {
                                  2, "test-first-name-2", "test-last-name-2")));
 }
 
+/// @test Verify that read-write transactions with multiple statements work.
+TEST_F(MutateAndReadIntegrationTest, MultipleInserts) {
+  auto commit_result = client_->Commit(
+      MakeReadWriteTransaction(),
+      {InsertMutationBuilder("Singers", {"SingerId", "FirstName", "LastName"})
+           .EmplaceRow(1, "test-fname-1", "test-lname-1")
+           .EmplaceRow(2, "test-fname-2", "test-lname-2")
+           .Build()});
+  EXPECT_STATUS_OK(commit_result);
+
+  auto txn = MakeReadWriteTransaction();
+  auto insert1 = client_->ExecuteSql(
+      txn, SqlStatement("INSERT INTO Singers (SingerId, FirstName, LastName) "
+                        "VALUES (@id, @fname, @lname)",
+                        {{"id", Value(3)},
+                         {"fname", Value("test-fname-3")},
+                         {"lname", Value("test-lname-3")}}));
+  EXPECT_STATUS_OK(insert1);
+  auto insert2 = client_->ExecuteSql(
+      txn, SqlStatement("INSERT INTO Singers (SingerId, FirstName, LastName) "
+                        "VALUES (@id, @fname, @lname)",
+                        {{"id", Value(4)},
+                         {"fname", Value("test-fname-4")},
+                         {"lname", Value("test-lname-4")}}));
+  EXPECT_STATUS_OK(insert2);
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/integration_tests/mutate_and_read_integration_test.cc
```diff
@@ -196,6 +196,127 @@ TEST_F(MutateAndReadIntegrationTest, DeleteAndCommit) {
...
+  }
+
+  EXPECT_THAT(returned_rows,
+              UnorderedElementsAre(RowType(1, "test-fname-1", "test-lname-1"),
+                                   RowType(2, "test-fname-2", "test-lname-2"),
+                                   RowType(3, "test-fname-3", "test-lname-3"),
+                                   RowType(4, "test-fname-4", "test-lname-4")));
+}
+
+/// @test Verify that read-write transactions with multiple statements work.
```
devbww: Un-updated cut-n-paste?

coryan: Indeed, fixed.

google/cloud/spanner/row.h
```diff
@@ -240,6 +240,23 @@ class Row {
...
+   * assertions fail.
+   *
+   * @warning This is intended for debugging and human consumption only, not
+   *   machine consumption as the output format may change without notice.
+   */
+  friend void PrintTo(Row const& r, std::ostream* os) {
+    char const* sep = "{";
+    for (auto const& v : r.values()) {
+      *os << sep;
+      PrintTo(v, os);
```
devjgm: It may be good to use `testing::PrintToString` here, which will work on types that don't have `op<<` or `PrintTo` (which I suspect is the case with like, vector and tuple). See https://github.com/google/googletest/blob/master/googletest/docs/advanced.md

But I'd also like to avoid including a googletest header in this public .h file. Maybe we could put this `PrintTo` body in a .cc file?

What do you think?

coryan: > It may be good to use `testing::PrintToString` here, which will work on types that don't have `op<<` or `PrintTo` (which I suspect is the case with like, vector and tuple).

The `v` are always a `spanner::Value`, right?  So I am actually calling `google::cloud::spanner::PrintTo(Value const&, std::ostream*)` and we know `Value` has a `PrintTo`.

> But I'd also like to avoid including a googletest header in this public .h file. Maybe we could put this `PrintTo` body in a .cc file?

That would be hard because I need a `PrintTo(Row<Ts...> const& r, std::ostream* os)`, it does not look like a template here because it is in the middle of the template class definition.

devjgm: My mistake. You're right. I overlooked the fact that you're calling `.values()`. What you have now LGTM.

google/cloud/spanner/integration_tests/mutate_and_read_integration_test.cc
```diff
@@ -196,6 +196,127 @@ TEST_F(MutateAndReadIntegrationTest, DeleteAndCommit) {
...
+  EXPECT_STATUS_OK(insert1);
+  auto insert2 = client_->ExecuteSql(
+      txn, SqlStatement("INSERT INTO Singers (SingerId, FirstName, LastName) "
+                        "VALUES (@id, @fname, @lname)",
+                        {{"id", Value(4)},
+                         {"fname", Value("test-fname-4")},
+                         {"lname", Value("test-lname-4")}}));
+  EXPECT_STATUS_OK(insert2);
+
+  auto reader = client_->Read(txn, "Singers", KeySet::All(),
```
devjgm: For posterity, I'm confused why this works. The documentation here https://cloud.google.com/spanner/docs/transactions says "**Transaction reads don't see the same transaction's writes.** It's important to note that reads do not see the writes that are buffered in the same transaction....", which makes me think that this should not work.... but it clearly does work.

Anyway, I'm checking w/ the spanner folks to better understand what is *supposed* to happen and maybe highlight to them some misleading documentation.

No change for you, just notes/comments FTR.

PR #367, opened 2019-08-15
feat: Implement logging stub for spanner.

coryan:
Add logging support for the Spanner client. By setting some environment variables we
get each request and (non-streaming) response logged.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/367)


google/cloud/spanner/internal/logging_spanner_stub.cc
```diff
@@ -0,0 +1,240 @@
...
+}
+
+template <typename Functor, typename Request,
+          typename Result = google::cloud::internal::invoke_result_t<
+              Functor, grpc::ClientContext&, Request const&>,
+          typename std::enable_if<IsUniquePtr<Result>::value, int>::type = 0>
+Result LogHelper(Functor&& functor, grpc::ClientContext& context,
+                 Request const& request, char const* where) {
+  GCP_LOG(DEBUG) << where << "() << " << request.DebugString();
+  return functor(context, request);
```
devjgm: Do you think it might still be useful to log whether the returned unique_ptr was nullptr or not?

google/cloud/spanner/internal/spanner_stub.cc
```diff
@@ -235,14 +238,46 @@ StatusOr<spanner_proto::PartitionResponse> DefaultSpannerStub::PartitionRead(
   return response;
 }
 
+bool RpcTracingEnabled() {
+  auto enable_clog =
+      google::cloud::internal::GetEnv("GOOGLE_CLOUD_CPP_ENABLE_CLOG");
+  if (enable_clog.has_value()) {
+    (void)google::cloud::LogSink::EnableStdClog();
```
devjgm: Curious: does the compiler or clang-tidy or something require this cast to void?

coryan: No, that was a mistake. Fixed now.

google/cloud/spanner/internal/logging_spanner_stub.cc
coryan: Don't know.  In every case these are `std::unique_ptr<grpc::ClientReader<T>>`, i.e., the result of a streaming (read) RPC.  As far as I can tell those are never null: gRPC indicates an error by returning a (non-null) reader, whose `Finish()` function tells you what the error was.

But it is easy to add, and may help in timing, and would be less surprising, so why not?

devjgm: I'll leave it to you. I didn't realize the limited cases where unique_ptr was used here, so it may not be that useful. I only noticed that this overload looked different from the others and was wondering why.

Again, your call if you think it's useful.

devjgm: ... oh, and you added it.. OK. Cool.

PR #368, opened 2019-08-16
refactor: Single client for all MutateAndReadIntegrationTest cases

devbww:
Use a single `Client` (which implies a single session) of all the
MutateAndReadIntegrationTest cases. This reduces the test startup/
teardown costs immensely.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/368)


coryan:
Really cool, thanks!

coryan:
Before I forget, I think we should refactor the `::testing::Environment` stuff to `SetUpTestSuite` and `TearDownTestSuite`. I just did not know you could initialize static variables for the whole `TestSuite`, I can do that in a future PR.

PR #372, opened 2019-08-16
Parallelize external builds

### LINKED ISSUE #342: docker cmake-super build compiles with -j1
The docker cmake-super build takes significantly longer than other builds (2-3x).  From the logs we see:

`make[3]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.`

@coryan suggests we could use ninja or CMAKE_BUILD_PARALLEL_LEVEL if we move to cmake-3.12+ https://github.com/googleapis/google-cloud-cpp/issues/2278

mr-salty:
Fixes #342

At least on my machine this dropped the time for `build.sh cmake-super` from 15 min to 4.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/372)


mr-salty:
FYI, the cmake-super build here took 13:45  (and even beat out `msan`).  The 10 most recent runs took 22 to 30 minutes, so this should be a big win.

coryan:
Thank you so much for doing this!  LGTM.

mr-salty:
Thanks to @remyabel for the suggestion in the issue - I wouldn't have figured it out myself!

PR #373, opened 2019-08-16
feat: makes KeySetBuilder<T>::Build() more efficient with rvalue qualifier

devjgm:
Rvalue-ref qualifies the `KeySetBuilder<T>::Build()` method which will enable it to *move* all of its data into the protos in the resulting `KeySet`. I needed to add rvalue-qualified accessors to Bound and KeyRange as well.

I left the `KeySetBuilder<T>::Build() const&` overload, which is inefficient and will result in a copy of all the keys. We could remove this overload if we want to.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/373)

### DIFF
--- a/google/cloud/spanner/keys.h
+++ b/google/cloud/spanner/keys.h
@@ -331,26 +331,31 @@ class KeySetBuilder {
     return key_ranges_;
   }
 
-  /**
-   * Adds a key `spanner::Row` to the `KeySetBuilder`.
-   */
-  KeySetBuilder& Add(RowType key) {
+  /// Adds a key `spanner::Row` to the `KeySetBuilder`.
+  KeySetBuilder&& Add(RowType key) && {
+    keys_.push_back(std::move(key));
+    return std::move(*this);
+  }
+
+  /// Adds a key `spanner::Row` to the `KeySetBuilder`.
+  KeySetBuilder& Add(RowType key) & {
     keys_.push_back(std::move(key));
     return *this;
   }
 
-  /**
-   * Adds a `KeyRange` to the `KeySetBuilder`.
-   */
-  KeySetBuilder& Add(KeyRange<RowType> key_range) {
+  /// Adds a `KeyRange` to the `KeySetBuilder`.
+  KeySetBuilder&& Add(KeyRange<RowType> key_range) && {
+    key_ranges_.push_back(std::move(key_range));
+    return std::move(*this);
+  }
+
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/keys.h
```diff
@@ -153,8 +155,11 @@ class KeyRange {
   KeyRange(KeyRange&& key_range) = default;
   KeyRange& operator=(KeyRange&& rhs) = default;
 
-  Bound<RowType> const& start() const { return start_; }
-  Bound<RowType> const& end() const { return end_; }
+  Bound<RowType> const& start() const& { return start_; }
+  Bound<RowType>&& start() && { return std::move(start_); }
```
coryan: This does not seem to be tested?

google/cloud/spanner/keys.h
```diff
@@ -153,8 +155,11 @@ class KeyRange {
...
+
+  Bound<RowType> const& end() const& { return end_; }
+  Bound<RowType>&& end() && { return std::move(end_); }
```
coryan: Nor this?

google/cloud/spanner/keys.h
```diff
@@ -362,12 +345,40 @@ class KeySetBuilder {
...
    *
    */
-  KeySet Build() const { return KeySet(*this); }
+  KeySet Build() && {
+    google::spanner::v1::KeySet ks;
+    for (auto& key : keys_) {
+      Append(ks.add_keys(), std::move(key).values());
+    }
+    for (auto& range : key_ranges_) {
+      auto* kr = ks.add_ranges();
```
coryan: The code inside the loop does not seem to be tested at all?

coryan:
___
*[google/cloud/spanner/keys.h, line 159 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/373#-LmQlWPI1kV7B8G_Zhst-r1-159:-LmRBplI-RC4sCSnlAfU:b-m11v8n) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/fa59c0c6b7b9f914ebd25b0664c9e6c4fc25778a/google/cloud/spanner/keys.h#L159)):*
<details><summary><i>Previously, devjgm (Greg Miller) wrote…</i></summary><blockquote>

Added some tests, but they're not great. testing moves is hard. Let me know if you have a better idea.
</blockquote></details>

Did you actually `git push` them?

PR #374, opened 2019-08-16
cleanup: Rename integration test.

coryan:
I think we discussed how the current test name is not great.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/374)


coryan:
Rebased. PTAL.

PR #381, opened 2019-08-19
feat: Implement spanner_insert_data sample.

coryan:
This sample populates the tables used in all the other samples, seems
sensible to add it ASAP.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/381)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -146,32 +146,33 @@ void InsertData(std::vector<std::string> const& argv) {
   }
 
   // TODO(#377) - cache the client across sample functions.
-  namespace gcs = google::cloud::spanner;
-  gcs::Client client(
-      gcs::MakeConnection(gcs::MakeDatabaseName(argv[0], argv[1], argv[2])));
+  namespace spanner = google::cloud::spanner;
+  spanner::Client client(spanner::MakeConnection(
+      spanner::MakeDatabaseName(argv[0], argv[1], argv[2])));
 
   //! [START spanner_insert_data]
-  namespace gcs = google::cloud::spanner;
-  [](gcs::Client client) {
-    auto insert_singers = gcs::InsertMutationBuilder(
+  namespace spanner = google::cloud::spanner;
+  [](spanner::Client client) {
+    auto insert_singers = spanner::InsertMutationBuilder(
                               "Singers", {"SingerId", "FirstName", "LastName"})
                               .EmplaceRow(1, "Marc", "Richards")
                               .EmplaceRow(2, "Catalina", "Smith")
                               .EmplaceRow(3, "Alice", "Trentor")
                               .EmplaceRow(4, "Lea", "Martin")
-                              .EmplaceRow(5, "David", "Lomond");
+                              .EmplaceRow(5, "David", "Lomond")
+                              .Build();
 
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -139,6 +139,48 @@ void DropDatabase(std::vector<std::string> const& argv) {
...
+  //! [START spanner_insert_data]
+  namespace gcs = google::cloud::spanner;
+  [](gcs::Client client) {
+    auto insert_singers = gcs::InsertMutationBuilder(
+                              "Singers", {"SingerId", "FirstName", "LastName"})
+                              .EmplaceRow(1, "Marc", "Richards")
+                              .EmplaceRow(2, "Catalina", "Smith")
+                              .EmplaceRow(3, "Alice", "Trentor")
+                              .EmplaceRow(4, "Lea", "Martin")
+                              .EmplaceRow(5, "David", "Lomond");
```
devjgm: What do you think about ending this fluent call sequence with a call to `.Build();`? That would remove the need for `std::move(x).Build()` below.

google/cloud/spanner/samples/samples.cc
```diff
@@ -139,6 +139,48 @@ void DropDatabase(std::vector<std::string> const& argv) {
...
+        "insert-data <project-id> <instance-id> <database-id>");
+  }
+
+  // TODO(#377) - cache the client across sample functions.
+  namespace gcs = google::cloud::spanner;
+  gcs::Client client(
+      gcs::MakeConnection(gcs::MakeDatabaseName(argv[0], argv[1], argv[2])));
+
+  //! [START spanner_insert_data]
+  namespace gcs = google::cloud::spanner;
```
devjgm: I didn't notice this before, but `gcs::Foo` looks to me like `Foo` within the `google::cloud::storage` namespace. I realize the 3-letter abbreviation works here too, but I wonder if it will be confusing.

Suggestion: Maybe we should consider using a `spanner` namespace alias in this case instead of "gcs"?

Thoughts?

google/cloud/spanner/samples/samples.cc

google/cloud/spanner/samples/samples.cc
coryan: I was not liking `gcs::` for the same reasons, changed to `spanner::` in all the samples.

coryan:
PTAL.

google/cloud/spanner/samples/samples.cc
```diff
@@ -139,30 +139,74 @@ void DropDatabase(std::vector<std::string> const& argv) {
...
+                             "Albums", {"SingerId", "AlbumId", "AlbumTitle"})
+                             .EmplaceRow(1, 1, "Total Junk")
+                             .EmplaceRow(1, 2, "Go, Go, Go")
+                             .EmplaceRow(2, 1, "Green")
+                             .EmplaceRow(2, 2, "Forever Hold Your Peace")
+                             .EmplaceRow(2, 3, "Terrified")
+                             .Build();
+
+    auto commit_result = client.Commit(spanner::MakeReadWriteTransaction(),
+                                       {insert_singers, insert_albums});
```
devjgm: Consider moving the `insert_singers` and `insert_albums` since it looks like we're done w/ them.

PR #382, opened 2019-08-19
feat: Add `Database` class to encapsulate project_id, instance_id, database_id

### LINKED ISSUE #369: Consider: a Database class (rather than a string)
Related to closed issue #301 

A Spanner database name includes a `project-id`, `instance-id`, and `database-id`. In RPCs it's represented as a string of the form `projects/<project-id>/instances/<instance-id>/databases/<database-id>`. Today in our APIs we generally pass around a database name as a `std::string`, and we provide a `MakeDatabaseName()` (https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/client.h#L371) helper function to turn the required 3 pieces into a string of the valid form.

My concerns with the current approach:

1. APIs which have a param like `std::string database_name` do not make it clear that the database name must have some specific string format. The documentation *might* have this information, but it will not appear in the code at the call site.

2. Every valid `database_name` string argument should have been constructed from our `MakeDatabaseName()` function, but nothing in the code enforces this because there's no strong type for this argument.

## Proposal

I think we should consider a small helper class like the following:

```cc
// File: database.h
class Database {
 public:
  Database(std::string project_id, std::string instance_id, std::string database_id);
  // Regular type: copy, assign, move, all `=default`

  // Accessors
  std::string const& project_id() const;
  // Also for instance_id and database_id

  // Returns the fully qualified database name of the form:
  // "projects/<project-id>/instances/<instance-id>/databases/<database-id>"
  std::string QualifiedName() const;

  // Maybe, because convenient, and there's a clear stringable format?
  std::ostream& operator<<(std::ostream& out, Database const& dn);

  friend bool operator==(Database const& a, Database const& b);
  friend bool operator!=(Database const& a, Database const& b);

 private:
  // 3 fields for project_id, instance_id, database_id
};
```

If we had this class, I'd propose that this should be the *only* place that project_id, instance_id, database_id are specified. That is **no** other functions should take those arguments. Instead, they should all accept an argument of type `Database`.

devjgm:
Fixes #369 

This PR adds the `Database` class, which identifies a Spanner database. It is constructed with a `project_id`, `instance_id`, and `database_id`.

All functions that need these fields now accept a `Database` argument instead of a `std::string` or 3-separate `std::string`s.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/382)

### DIFF
--- a/google/cloud/spanner/database.cc
+++ b/google/cloud/spanner/database.cc
@@ -1,0 +1,61 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/spanner/database.h"
+#include <array>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+Database::Database(std::string project_id, std::string instance_id,
+                   std::string database_id)
+    : project_id_(std::move(project_id)),
+      instance_id_(std::move(instance_id)),
+      database_id_(std::move(database_id)) {}
+
+std::string Database::FullName() const {
+  // "projects/<project-id>/instances/<instance-id>/databases/<database-id>"
+  constexpr char kProjects[] = "projects/";
+  constexpr char kInstances[] = "/instances/";
+  constexpr char kDatabases[] = "/databases/";
+  std::string name;
+  name.reserve(sizeof(kProjects) + sizeof(kInstances) + sizeof(kDatabases) - 3 +
+               project_id().size() + instance_id().size() +
+               database_id().size());
+  name += kProjects;
+  name += project_id();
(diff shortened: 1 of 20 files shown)


coryan:
___
*[google/cloud/spanner/database.h, line 37 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/382#-LmfIEWkEdtZm0mqsub0:-Lmf_sGJ2ITKJSwl4_45:b313vie) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/84fcbc64ee8a38e6400b00b5a5b055019d758cc8/google/cloud/spanner/database.h#L37)):*

Is it worthwhile to do validation on the client, or just let the server return errors? I think I'd vote for the latter
</blockquote></details>

Server side, please: https://github.com/googleapis/google-cloud-cpp/blob/master/doc/adr/2018-06-19-do-not-duplicate-server-side-validation.md

google/cloud/spanner/database.cc
```diff
@@ -0,0 +1,54 @@
...
+Database::Database(std::string const& project_id,
+                   std::string const& instance_id,
+                   std::string const& database_id)
+    : full_name_("projects/" + project_id + "/instances/" + instance_id +
+                 "/databases/" + database_id) {}
+
+std::string Database::FullName() const { return full_name_; }
+
+std::string Database::DatabaseId() const {
+  auto pos = full_name_.rfind('/');
```
devbww: Should we prefer to find "/databases/", like in `ParentName()`?

devjgm: We took a vote and 3 to 1 prefer Brad's suggestion, so changed :-)

PR #384, opened 2019-08-19
feat: add ReadPartition files

scotthart:
- (add KeySet equality operators and constructor from proto) <-- extra commit from a previously merged PR from a branch of this branch
- implement remainder of ReadPartition and some clang tidy
- refactor ReadPartition to use proto internally

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/384)


google/cloud/spanner/keys_test.cc
```diff
@@ -78,6 +78,46 @@ TEST(KeyrangeTest, Accessors) {
       std::is_same<EndType&&, decltype(std::move(range).start())>::value, "");
 }
 
+TEST(KeySetTest, EqualityEmpty) {
```
devjgm: Were these tests moved? I think their old location was preferable, because then they'll come after KeyRangeTest, which matches the order they are defined in the header.

google/cloud/spanner/keys_test.cc
```diff
@@ -160,65 +200,19 @@ TEST(KeySetTest, AllKeys) {
...
-          .Build(),                                                        //
-      KeySetBuilder<Row<std::int64_t, std::string>>()                      //
-          .Add(MakeRow(42, "hi"))                                          //
-          .Add(MakeRow(123, "bye"))                                        //
-          .Build(),                                                        //
-      KeySetBuilder<Row<std::int64_t, std::string>>()                      //
-          .Add(MakeKeyRangeClosed(MakeRow(42, "hi"), MakeRow(43, "bye")))  //
-          .Build(),                                                        //
+      KeySetBuilder<Row<>>().Build(),
+      KeySetBuilder<Row<std::int64_t>>().Add(MakeRow(42)).Build(),
```
devjgm: FWIW, the empty end-of-line comments where intentionally there to make clang-format wrap the lines in a way that made each call to `KeySetBuilder<T>()` look the same, i.e., have `.Add()` on a separate line, followed by `.Build()` on a separate line. 

Now it's subjective if that was an *improvement* :-) If you prefer the new formatting, that's fine w/ me.

google/cloud/spanner/read_partition.cc
```diff
@@ -0,0 +1,92 @@
...
+  proto_.mutable_transaction()->set_id(std::move(transaction_id));
+  proto_.set_table(std::move(table_name));
+  if (!read_options.index_name.empty()) {
+    proto_.set_index(std::move(read_options.index_name));
+  }
+  for (auto& column : column_names) {
+    *proto_.mutable_columns()->Add() = std::move(column);
+  }
+  *proto_.mutable_key_set() = internal::ToProto(std::move(key_set));
+  if (read_options.limit != 0) {
```
devjgm: It may be better to omit this check and always call `proto_.set_limit()` with whatever `read_options.limit` is set to. That way this class isn't applying any semantic meaning to these fields, it's just copying them through.

I suggest the same thing with `read_options.index_name` above.

google/cloud/spanner/read_partition.h
```diff
@@ -0,0 +1,158 @@
...
+ * Returned `Status` should be checked to determine if deserialization was
+ * successful.
+ *
+ * @param serialized_read_partition - string representation to be deserialized.
+ *
+ * @par Example
+ *
+ * @code
+ * std::string serialized_partition = ReceiveFromRemoteMachine();
+ * spanner::ReadPartition partition =
```
devjgm: This returns a StatusOr

google/cloud/spanner/read_partition.h
```diff
@@ -0,0 +1,158 @@
...
+  ///@{
+  ReadPartition(ReadPartition const&) = default;
+  ReadPartition(ReadPartition&&) = default;
+  ReadPartition& operator=(ReadPartition const&) = default;
+  ReadPartition& operator=(ReadPartition&&) = default;
+  ///@}
+
+  std::string table_name() const { return proto_.table(); }
+  std::vector<std::string> column_names() const {
+    auto columns = proto_.columns();
```
devjgm: Will this make a copy of all the columns here? If so, you may want to avoid that extra copy w/ `auto const&`.

google/cloud/spanner/read_partition.h
```diff
@@ -0,0 +1,158 @@
...
+    return std::vector<std::string>(columns.begin(), columns.end());
+  }
+  ReadOptions read_options() const {
```
devjgm: I think all these accessors should be named like `ReadOptions()` and `ColumnNames()`, etc. Is clang-tidy happy w/ that?

google/cloud/spanner/client_options.h
```diff
@@ -96,6 +96,14 @@ struct ReadOptions {
    * A limit cannot be specified when calling`PartitionRead`.
    */
   std::int64_t limit = 0;
+
+  friend bool operator==(ReadOptions const& lhs, ReadOptions const& rhs) {
```
devjgm: I don't see why these operators are needed. Are these used?

PR #391, opened 2019-08-20
feat: implement spanner_read_{only,write}_transaction samples.

### LINKED ISSUE #180: Sample: Developers can use read-only transactions
Sample code: spanner_read_only_transaction [C#](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/a129e6228fb4a6cb660832168b8f02ac0675673f/spanner/api/Spanner/Program.cs#L863)

This should involve creating a read-only transaction, and executing several reads (or other read-only operations) using the same transaction.

### LINKED ISSUE #181: Sample: Developers can use read-write transactions
Sample code: spanner_read_write_transaction [C#](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/a129e6228fb4a6cb660832168b8f02ac0675673f/spanner/api/Spanner/Program.cs#L1120)

This should involve creating a read-write transaction, and executing a read, creating some mutations, and committing the RW transaction using the `RunTransaction()` function.

coryan:
This fixes #180 and fixes #181.

I also implemented the `spanner_update_data` sample, there was no
issue for it, but we need to prepare the database state for these samples.

I refactored some of the support code because it was getting tedious.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/391)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -144,10 +144,11 @@ void DropDatabase(std::vector<std::string> const& argv) {
 }
 
 google::cloud::spanner::Client MakeSampleClient(
-    std::vector<std::string> const& argv) {
+    std::string const& project_id, std::string const& instance_id,
+    std::string const& database_id) {
   namespace spanner = google::cloud::spanner;
-  return spanner::Client(
-      spanner::MakeConnection(spanner::Database(argv[0], argv[1], argv[2])));
+  return spanner::Client(spanner::MakeConnection(
+      spanner::Database(project_id, instance_id, database_id)));
 }
 
 //! [START spanner_insert_data]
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -143,6 +143,13 @@ void DropDatabase(std::vector<std::string> const& argv) {
   (argv[0], argv[1], argv[2]);
 }
 
+google::cloud::spanner::Client MakeSampleClient(
+    std::vector<std::string> const& argv) {
```
devbww: Given this requires an `argv.size()` of (at least) three, and it is the caller who actually checked that, perhaps `MakeSampleClient()` should be taking the three arguments separately?

coryan:
PTAL.

PR #394, opened 2019-08-20
Support chunked_value reassembly.

### LINKED ISSUE #270: Implement chunked_value support in spanner::Client
Use the unchunker (#161) to handle chunked_values in the spanner Client

mr-salty:
Fixes #270

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/394)

### DIFF
--- a/google/cloud/spanner/internal/partial_result_set_reader.cc
+++ b/google/cloud/spanner/internal/partial_result_set_reader.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "google/cloud/spanner/internal/partial_result_set_reader.h"
+#include "google/cloud/spanner/internal/merge_chunk.h"
 #include "google/cloud/grpc_utils/grpc_error_delegate.h"
 #include "google/cloud/log.h"
 
@@ -54,6 +55,10 @@ StatusOr<optional<Value>> PartialResultSetReader::NextValue() {
       return status;
     }
     if (finished_) {
+      if (partial_chunked_value_.has_value()) {
+        return Status(StatusCode::kInternal,
+                      "incomplete chunked_value at end of stream");
+      }
       return optional<Value>();
     }
     // If the response contained any values, the loop will exit, otherwise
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/internal/partial_result_set_reader.cc
```diff
@@ -115,8 +120,36 @@ Status PartialResultSetReader::ReadFromStream() {
...
+  // If the last response had `chunked_value` set, partial_chunked_value_
+  // contains the partial value; merge it with the first value in the
+  // response.
+  if (partial_chunked_value_.has_value()) {
+    if (values_.empty()) {
+      return Status(StatusCode::kInternal,
+                    "PartialResultSet contained no values to merge with prior "
+                    "chunked_value");
+    }
+    auto merge_status =
```
devbww: Why do we choose to merge `values_[0]` onto `partial_chunked_value_` and then move `partial_chunked_value_` to `values_[0]`, rather than just merge `partial_chunked_value_` into `values_[0]`?  That is, should `MergeChunk()` be "onto front" rather than "onto end"?

google/cloud/spanner/internal/partial_result_set_reader_test.cc
```diff
@@ -442,6 +442,247 @@ TEST(PartialResultSetReaderTest, ResponseWithNoValues) {
...
+  EXPECT_CALL(*grpc_reader, Finish()).WillOnce(Return(grpc::Status()));
+
+  auto context = make_unique<grpc::ClientContext>();
+  auto reader = PartialResultSetReader::Create(std::move(context),
+                                               std::move(grpc_reader));
+  EXPECT_STATUS_OK(reader.status());
+
+  // Trying to read the next value should fail.
+  auto value = (*reader)->NextValue();
+  EXPECT_EQ(value.status().code(), StatusCode::kInvalidArgument);
```
devbww: Look for "invalid type" in message?

google/cloud/spanner/internal/partial_result_set_reader.cc
mr-salty: Maybe, what @devjgm implemented was what I told him I needed, and doing the opposite seems weird as an API (the current one is like 'operator+=' , is there an analog for the other one?  I guess it would need to be named something with "Prepend" to make sense). Under the covers we don't have much choice but to append since we're dealing with strings and vectors. Being being able to modify both arguments saves us half a copy. I'll let Greg weigh in, or we can discuss after lunch.

google/cloud/spanner/internal/partial_result_set_reader.cc
```diff
@@ -115,8 +120,36 @@ Status PartialResultSetReader::ReadFromStream() {
...
+        MergeChunk(*partial_chunked_value_, std::move(values_[0]));
```
devjgm: Values can be chunked into *multiple* subsequent Values ([example](https://github.com/googleapis/googleapis/blob/11592a15391951348a64f5c303399733b1c5b3b2/google/spanner/v1/result_set.proto#L117-L135)). So I think we need to keep checking the `chunked_value` bool and keep merging in a loop until we get one that is not chunked.

google/cloud/spanner/internal/partial_result_set_reader.cc
devjgm: We can certainly change `MergeChunk` if a different signature would be more convenient to use here, though I believe it would still need to maintain the append-like semantics.

google/cloud/spanner/internal/partial_result_set_reader.cc
coryan: Just curious, in that example the intermediate [value](https://github.com/googleapis/googleapis/blob/11592a15391951348a64f5c303399733b1c5b3b2/google/spanner/v1/result_set.proto#L127-L131) has only one element, is it possible for it to have multiple elements?

In other words, how does one interpret something like (not that I split the value in the second entry):

```
  //     {
  //       "metadata": { ... }
  //       "values": ["Hello", "W"]
  //       "chunked_value": true
  //       "resume_token": "Af65..."
  //     }
  //     {
  //       "values": ["o", "r", "l"]
  //       "chunked_value": true
  //       "resume_token": "Bqp2..."
  //     }
  //     {
  //       "values": ["d"]
  //       "resume_token": "Zx1B..."
  //     }
```

PR #395, opened 2019-08-21
Add `std::string` overloads for `Client::ExecuteSql()`

### LINKED ISSUE #389: Consider `Client::ExecuteSql()` overloads with `std::string`.
It is fairly common to write simple queries like: `client.ExecuteSql("SELECT a, b FROM tbl")`, currently all the overloads require a `SqlStatement` which means the application developer (specially folks just learning about the library) need to wrap strings.

I think we should be doing this for them, it makes it easier to teach about the library.

remyabel:
Fixes #389

These are simple wrappers for their `SqlStatement` counterparts.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/395)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -67,6 +67,10 @@ StatusOr<ResultSet> Client::ExecuteSql(SqlStatement statement) {
        std::move(statement)});
 }
 
+StatusOr<ResultSet> Client::ExecuteSql(std::string const& statement) {
+  return ExecuteSql(SqlStatement{statement});
+}
+
 StatusOr<ResultSet> Client::ExecuteSql(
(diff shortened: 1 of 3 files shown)


remyabel:
> Please note that we have not discussed inside the team if this is a good idea or not, I filed the issue as Consider ... to do precisely that, talk to the team and decide if it was a good idea.

Fair enough. I spent more time trying to figure out how to re-use the `ResultSet`/`source` mock objects (if that is even possible) than the rest of the PR, so no foul.

> thoughts? Is this overkill / overtesting?

Probably. I think that is already tested in `sql_statement_test.cc`. Since this is in `client_test.cc`, the tests are better suited for external behavior rather than individual methods of the class.

Issue #390, opened 2019-08-20
Add factory functions for spanner::TransactionAction


coryan:
It would be nice to have factory functions to return in the simple cases, maybe:

```C++
struct TransactionAction {
  static TransactionAction Commit();
  static TransactionAction Rollback();
 ...
};
```

PR #397, opened 2019-08-21
Simplify Client::RunTransaction() functor interface

devbww:
DO NOT SUBMIT.  FOR DISCUSSION ONLY FOR NOW.

Replace `TransactionAction` with `StatusOr<Mutations>` as the return
type for the `Client::RunTransaction()` functor. This makes it easier
to write a functor (addressing #390), and makes it very clear what
`RunTransaction()` should return when the transaction will be rolled
back.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/397)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/spanner/client.h"
 #include "google/cloud/spanner/internal/connection_impl.h"
 #include "google/cloud/spanner/internal/spanner_stub.h"
+#include "google/cloud/log.h"
 #include <grpcpp/grpcpp.h>
 
 namespace google {
@@ -106,7 +107,7 @@ StatusOr<std::int64_t> Client::ExecutePartitionedDml(
 }
 
 StatusOr<CommitResult> Client::Commit(Transaction transaction,
-                                      std::vector<Mutation> mutations) {
+                                      Mutations mutations) {
   return conn_->Commit({std::move(transaction), std::move(mutations)});
 }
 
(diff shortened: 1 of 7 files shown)


coryan:
When implementing the function called by `RunTransaction` this seems much better. Thanks.

One question I have is: as a caller to `RunTransaction`, how do I know if the error was in the function or in a RPC called by the function?

devbww:
> I'm unsure about the `Mutations` alias. I somewhat lean toward not using that alias because users need to know that it's a `vector<Mutation>` in order to use objects of that type, so it's not a real abstraction, it's simply a shorthand. And since it's a shorthand, I think users spelling `vector<Mutation>` may be clearer than `Mutations`, which hides the type of the object. Plus, I think this might be our first "convenience alias" in the code?
> 
> ... but I'm maybe 60/40 in favor of dropping the alias. I'm happy to go w/ the consensus of what others think

Indeed.  Originally I didn't have it, but found the continued use of `std::vector<Mutation>` to be irritating enough (particularly in the lengthy `std::function` instantiations and the trailing return types and return expressions for the corresponding lambdas).  I imagine external users of `RunTransaction()` may well feel the same way.

So, I agree that it is not an abstraction, and simply a shorthand, but I would still vote for it.  I don't think there is any practical loss in clarity.  Happy to go with the consensus too, though.

devbww:
> One question I have is: as a caller to `RunTransaction`, how do I know if the error was in the function or in a RPC called by the function?

Would I be correct in rephrasing this as: "How do I know if the error was from the `Commit()` or from the functor?"

I think the short answer is, you don't.  Do you have a use case in mind where you might need to make the distinction?  Perhaps interrogating the `Status` might be enough?.

devbww:
> Sure, supposed my code makes assumptions about the possible values in the database:

For my money, the "right" error code in that circumstance is `kAborted`.

> > Perhaps interrogating the `Status` might be enough?.
> 
> I would need to check the message, right? Because there are no status codes reserved for the application. That feels icky.

If you need to distinguish errors, and the status code isn't sufficient, then I guess so.  But the same would be true for any other error indicator that doesn't have per-origin-point codes.

coryan:
Okay, sounds better than what we have now.  Would be good to gain some experience using it.

devjgm:
> So, I agree that it is not an abstraction, and simply a shorthand, but I would still vote for it. I don't think there is any practical loss in clarity. Happy to go with the consensus too, though.

If the consensus is to keep the alias, then LGTM.

PR #403, opened 2019-08-22
cleanup: Refactor common code in samples.

coryan:
Also improve logging to troubleshoot build failures.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/403)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -165,15 +165,6 @@ void InsertData(google::cloud::spanner::Client client) {
 }
 //! [END spanner_insert_data]
 
-void InsertDataCommand(std::vector<std::string> const& argv) {
-  if (argv.size() != 3) {
-    throw std::runtime_error(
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -476,31 +396,41 @@ void QueryDataWithStruct(google::cloud::spanner::Client client) {
...
+      };
+    };
+    return CommandMap::value_type(sample_name,
+                                  make_command(sample_name, sample));
+  };
+
+  CommandMap commands = {
       {"create-database", &CreateDatabase},
       {"add-column", &AddColumn},
       {"drop-database", &DropDatabase},
```
devjgm: So I'm clear, these first three are different because they connect to the "admin" interface and don't use a `spanner::Client`?

coryan: Yes, that is why.

PR #406, opened 2019-08-22
bug: Make internal::ConnectionImpl thread-safe.

### LINKED ISSUE #423: ConnectionImpl is not thread safe.
The `sessions_` variable needs some kind of locking around it.

coryan:
Fixes #423.

This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/406)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -68,11 +68,17 @@ Status ConnectionImpl::Rollback(RollbackParams rp) {
 }
 
 StatusOr<ConnectionImpl::SessionHolder> ConnectionImpl::GetSession() {
+  std::unique_lock<std::mutex> lk(mu_);
   if (!sessions_.empty()) {
-    std::string session = sessions_.back();
+    std::string session = std::move(sessions_.back());
     sessions_.pop_back();
+    lk.unlock();
     return SessionHolder(std::move(session), this);
   }
+  // Release the mutex because we won't be doing any more changes to `sessions_`
+  // in this function and holding mutexes while making RPCs is (generally) a bad
+  // practice.
+  lk.unlock();
   grpc::ClientContext context;
   spanner_proto::CreateSessionRequest request;
   request.set_database(db_.FullName());
@@ -81,6 +87,14 @@ StatusOr<ConnectionImpl::SessionHolder> ConnectionImpl::GetSession() {
     return response.status();
   }
   return SessionHolder(std::move(*response->mutable_name()), this);
+}
+
+void ConnectionImpl::ReleaseSession(std::string session) {
+  if (session.empty()) {
+    return;
+  }
+  std::lock_guard<std::mutex> lk(mu_);
+  sessions_.emplace_back(std::move(session));
 }
 
 StatusOr<ResultSet> ConnectionImpl::Read(spanner_proto::TransactionSelector& s,
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -68,17 +68,18 @@ Status ConnectionImpl::Rollback(RollbackParams rp) {
 }
 
 StatusOr<ConnectionImpl::SessionHolder> ConnectionImpl::GetSession() {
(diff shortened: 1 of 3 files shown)


coryan:
For what is worth: both the tsan build and helgrind detect problems when I remove the code using the `mu_` mutex. Both pass with the current code.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -72,6 +78,14 @@ StatusOr<ConnectionImpl::SessionHolder> ConnectionImpl::GetSession() {
   return SessionHolder(std::move(*response->mutable_name()), this);
 }
 
+void ConnectionImpl::ReleaseSession(std::string session) {
+  if (session.empty()) {
```
devbww: Why should we make this check?

google/cloud/spanner/internal/connection_impl.h
```diff
@@ -54,7 +55,10 @@ class ConnectionImpl : public Connection {
    public:
     SessionHolder(std::string session, ConnectionImpl* conn) noexcept
         : session_(std::move(session)), conn_(conn) {}
-    ~SessionHolder() { conn_->sessions_.emplace_back(std::move(session_)); }
+    ~SessionHolder() {
+      conn_->ReleaseSession(std::move(session_));
+      session_.clear();
```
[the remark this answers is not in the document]
coryan: Done.

devbww:
> What we want is to return the session exactly once. I think (but check my math) that to achieve this we need to disallow copy on SessionHolder, and make the move copy and assignment clear something to make sure the destructor only takes action once.

Do we even need move copy and assignment?  Can we just make `session_` and `conn_` const to stop anything bad from happening to a `SessionHolder`?

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -497,6 +500,69 @@ TEST(ConnectionImplTest, RollbackSuccess) {
...
+  int const per_thread_iterations = 1000;
+  auto const thread_count = []() -> unsigned {
+    if (std::thread::hardware_concurrency() == 0) {
+      return 16;
+    }
+    return std::thread::hardware_concurrency();
+  }();
+
+  auto runner = [](int thread_id, int iterations,
+                   std::shared_ptr<Connection> const& conn) {
```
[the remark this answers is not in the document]
coryan: Sure, but it has to be `T&` because we are calling non-const member functions.  Note that this requires a `std::ref` below because `std::async` copies every argument.

coryan:
> > What we want is to return the session exactly once. I think (but check my math) that to achieve this we need to disallow copy on SessionHolder, and make the move copy and assignment clear something to make sure the destructor only takes action once.
> 
> Do we even need move copy and assignment? Can we just make `session_` and `conn_` const to stop anything bad from happening to a `SessionHolder`?

I think we need at least a move constructor because we put `SessionHolder` in a `StatusOr<SessionHolder>`.

coryan:
@devbww PTAL.

coryan:
Rebased to resolve conflicts.

devbww:
Have you looked at the TSan failure (I haven't :-).

coryan:
I noticed a few minutes before you asked, so far nothing obvious, and I needed to change the TSAN build to include line numbers.

coryan:
Good news, it was a bug in the test. Fixed now, ran 1000 iterations of the test without error.  Rebased for good measure and will update the `.bazelrc` file in a separate PR.

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -579,7 +579,7 @@ TEST(ConnectionImplTest, MultipleThreads) {
   auto db = Database("project", "instance", "database");
   std::string const session_prefix = "test-session-prefix-";
   std::string const transaction_id = "test-txn-id";
-  int session_counter = 0;
+  std::atomic<int> session_counter(0);
```
devbww: `#include <atomic>`

coryan: Done.

Issue #400, opened 2019-08-21
protect against SQL injection attacks


mr-salty:
`SqlStatement` can be constructed from a std::string of arbitrary origin, which allows the possibility for SQL injection attacks, especially if any part of the query string is dependent on user input. We should prevent unsafe use of this API to the largest possible extent.

Ideally we should be only accept query strings from a trusted sources like string literals, with all untrusted values relegated to parameters as described in https://cloud.google.com/spanner/docs/sql-best-practices

remyabel:
It is my understanding that SQL injection attacks are from the user to the application, not the application to the library. For example, an application that doesn't use parametrized queries is vulnerable to SQL injection. The library needs to trust input from the application (otherwise it wouldn't be very usable), but the application shouldn't trust input from the user as you should always assume the user is malicious.

> Ideally we should be only accept query strings from a trusted sources like string literals

Do you mean only allow string literals and not `std::string` or force the usage of only parametrized queries? I don't think either option would make the library very usable and the onus still falls upon the developer to minimize passing untrusted input to the library.

mr-salty:
Correct - the idea is to make it difficult for the application developer to inadvertently pass untrusted input to the API. This does place some additional requirements on developers using the library, but the general feeling (inside Google, at least - this topic came up in an internal discussion) is that providing APIs that cannot be used unsafely is the best way to prevent these types of issues, and it's standard practice for a lot of APIs internally.

https://github.com/google/safe-html-types/blob/master/doc/safehtml-types.md is one example that's analogous to what I'm describing.

remyabel:
That makes sense and thanks for the example. The `FromConstant` function is very interesting.

devjgm:
FTR, one incomplete way to do this in C++ is with a class like this (https://godbolt.org/z/xr1aUx):

```cc
struct StringLiteral {
  template <std::size_t N>
  explicit StringLiteral(char const (&a)[N]) : s(a, N) {}
  std::string s;
};
...
StringLiteral s("hello");
```

It's incomplete because it only requires a constant character array, not specifically a string literal. Though incomplete, it's portable, and may still be useful or helpful in our library.

Additionally, there are some compiler extensions that would enforce compile-time string literals for us. The problem here is that portability may suffer. But we could look into this.

Also, there might be some way to build a compile-time string class out of something like

```cc
template <char ...>
struct StringLiteral { ... };
```

But this is just a guess and someone would need to work out the details (if it's possible).

[...]

PR #407, opened 2019-08-22
fix: Harden SqlStatement against injection attacks

devjgm:
# DO NOT APPROVE - JUST FOR DISCUSSION 

This change makes `SqlStatement` only be constructible by compile-time
string literals. In C++11, there's no 100% guaranteed way to ensure this
(that I know of), so this approximates it using a fairly common
technique of requiring the argument to be a statically sized array of
char. There's a TODO to add some compiler-specific hooks (hidden behind
a macro) to verify this "better" if we can.

There will be cases where users need to construct their SQL query as a
std::string at runtime. For these cases there is a
`MakeUntrustedSqlStatement()` factory function, which callers can use.
When this function is used it is the caller's responsibility to verify
that the string they are passing is safe.

Fixes: #400

### DIFF
--- a/google/cloud/spanner/sql_partition.cc
+++ b/google/cloud/spanner/sql_partition.cc
@@ -80,9 +80,9 @@ StatusOr<SqlPartition> DeserializeSqlPartition(
     }
   }
 
-  SqlPartition sql_partition(proto.transaction().id(), proto.session(),
-                             proto.partition_token(),
-                             SqlStatement(proto.sql(), sql_parameters));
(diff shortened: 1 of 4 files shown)


devbww:
> JUST FOR DISCUSSION

How is ` MakeUntrustedSqlStatement()` different/better than simply applying the "It is the caller's responsibility to ensure that the `statement` argument is safe and free from SQL injection attacks" comment directly to the `SqlStatement(std::string ...)` constructors?

PR #408, opened 2019-08-22
feat: implement PartitionRead and Read(ReadPartition)

### LINKED ISSUE #284: Implement Client::PartitionRead()
Implement the body and unit test for `Client::PartitionRead`

See also: https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/client.h

### LINKED ISSUE #346: PartitionRead() should return an object like "ReadPartition"
Today `Client::PartitionRead` looks like:

```cc
  StatusOr<std::vector<SqlPartition>> PartitionRead(
      Transaction const& transaction, std::string const& table,
      spanner::KeySet const& keys, std::vector<std::string> const& columns,
      ReadOptions const& read_options = {},
      PartitionOptions const& partition_options = {});
```

But I think it's returning the wrong type. Instead of returning an `SqlPartition` it should probably return a `ReadPartition`.

scotthart:
fixes #346 
fixes #284

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/408)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -58,9 +58,9 @@ StatusOr<std::vector<ReadPartition>> Client::PartitionRead(
     std::vector<std::string> columns, ReadOptions read_options,
     PartitionOptions partition_options) {
   return conn_->PartitionRead(
-      {std::move(transaction), std::move(table), std::move(keys),
-       std::move(columns), std::move(read_options)},
-      std::move(partition_options));
+      {{std::move(transaction), std::move(table), std::move(keys),
+        std::move(columns), std::move(read_options)},
+       std::move(partition_options)});
 }
 
 StatusOr<ResultSet> Client::ExecuteSql(SqlStatement statement) {
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -49,16 +49,18 @@ StatusOr<ResultSet> Client::Read(Transaction transaction, std::string table,
                       std::move(columns), std::move(read_options)});
 }
 
-StatusOr<ResultSet> Client::Read(SqlPartition const& /*partition*/) {
-  return Status(StatusCode::kUnimplemented, "not implemented");
+StatusOr<ResultSet> Client::Read(ReadPartition const& read_partition) {
+  return conn_->Read(internal::MakeReadParams(read_partition));
 }
 
-StatusOr<std::vector<SqlPartition>> Client::PartitionRead(
-    Transaction const& /*transaction*/, std::string const& /*table*/,
-    KeySet const& /*keys*/, std::vector<std::string> const& /*columns*/,
-    ReadOptions const& /*read_options*/,
-    PartitionOptions const& /*partition_options*/) {
-  return Status(StatusCode::kUnimplemented, "not implemented");
+StatusOr<std::vector<ReadPartition>> Client::PartitionRead(
+    Transaction transaction, std::string table, KeySet keys,
+    std::vector<std::string> columns, ReadOptions read_options,
+    PartitionOptions partition_options) {
+  return conn_->PartitionRead(
+      {std::move(transaction), std::move(table), std::move(keys),
+       std::move(columns), std::move(read_options)},
+      std::move(partition_options));
 }
 
 StatusOr<ResultSet> Client::ExecuteSql(SqlStatement statement) {
--- a/google/cloud/spanner/connection.h
+++ b/google/cloud/spanner/connection.h
@@ -73,8 +73,13 @@ class Connection {
   };
   virtual StatusOr<ResultSet> Read(ReadParams) = 0;
 
+  struct PartitionReadParams {
+    ReadParams read_params;
+    PartitionOptions partition_options;
+  };
+
   virtual StatusOr<std::vector<ReadPartition>> PartitionRead(
-      ReadParams, PartitionOptions partition_options) = 0;
+      PartitionReadParams) = 0;
 
   struct ExecuteSqlParams {
     Transaction transaction;
--- a/google/cloud/spanner/connection.h
+++ b/google/cloud/spanner/connection.h
@@ -31,7 +31,7 @@ namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
-
+class ReadPartition;
 /**
  * A connection to a Spanner database instance.
  *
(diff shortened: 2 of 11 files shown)


google/cloud/spanner/client.cc
```diff
@@ -49,16 +49,18 @@ StatusOr<ResultSet> Client::Read(Transaction transaction, std::string table,
...
+  return conn_->Read(partition);
+}
+
+StatusOr<std::vector<ReadPartition>> Client::PartitionRead(
+    Transaction const& transaction /*transaction*/,
+    std::string const& table /*table*/, KeySet const& keys /*keys*/,
+    std::vector<std::string> const& columns /*columns*/,
+    ReadOptions const& read_options /*read_options*/,
+    PartitionOptions partition_options /*partition_options*/) {
+  return conn_->PartitionRead({transaction, table, keys, columns, read_options},
```
devjgm: I think we want this function's params to all be values and moved into the next call.

google/cloud/spanner/client.cc
```diff
@@ -49,16 +49,18 @@ StatusOr<ResultSet> Client::Read(Transaction transaction, std::string table,
...
-    ReadOptions const& /*read_options*/,
-    PartitionOptions const& /*partition_options*/) {
-  return Status(StatusCode::kUnimplemented, "not implemented");
+StatusOr<ResultSet> Client::Read(ReadPartition const& partition /*partition*/) {
+  return conn_->Read(partition);
+}
+
+StatusOr<std::vector<ReadPartition>> Client::PartitionRead(
+    Transaction const& transaction /*transaction*/,
+    std::string const& table /*table*/, KeySet const& keys /*keys*/,
```
devjgm: You can remove the `/*table*/` argument comments, which are unneeded now that we have actual argument names.

google/cloud/spanner/connection.h
```diff
@@ -54,9 +55,27 @@ class Connection {
     KeySet keys;
     std::vector<std::string> columns;
     ReadOptions read_options;
+
+    // TODO(#307): Refactor once correct location for session implemented.
+    google::cloud::optional<std::string> session_name;
+
+    ReadParams(Transaction transaction, std::string table, KeySet keys,
```
devjgm: Do we need this constructor? It looks like all the callers are using aggregate initialization, so I'm not sure this is needed.

google/cloud/spanner/connection.h
```diff
@@ -54,9 +55,27 @@ class Connection {
...
+        : transaction(std::move(transaction)),
+          table(std::move(table)),
+          keys(std::move(keys)),
+          columns(std::move(columns)),
+          read_options(std::move(read_options)),
+          session_name(std::move(session_name)) {}
   };
   virtual StatusOr<ResultSet> Read(ReadParams) = 0;
 
+  virtual StatusOr<ResultSet> Read(ReadPartition) = 0;
```
devjgm: I don't think we want a Read() overload here. `Client::Read(ReadPartition)` should call `Connection::Read(ReadParams)`. Do you think that'll work?

google/cloud/spanner/connection.h
```diff
@@ -54,9 +55,27 @@ class Connection {
     KeySet keys;
     std::vector<std::string> columns;
     ReadOptions read_options;
+
+    // TODO(#307): Refactor once correct location for session implemented.
+    google::cloud::optional<std::string> session_name;
```
devjgm: This will probably need the partition_token as well.

google/cloud/spanner/connection.h
scotthart: As ReadParams is likely a temporary container for sessions_name, I added the constructor so that the we don't need to change the call sites where we are not specifying session_name. When we remove session_name from ReadParams, we can remove this constructor.

google/cloud/spanner/client.cc
scotthart: Done.

google/cloud/spanner/client.cc
scotthart: Done.

google/cloud/spanner/client.cc
scotthart: Done.

google/cloud/spanner/connection.h
scotthart: Done.

google/cloud/spanner/connection.h
scotthart: I was under the impression we added the Connection interface to allow customers to mock calls. It seemed useful to expose a different Read signature that was intended to be used with the results of PartitionRead.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -193,6 +218,41 @@ Status ConnectionImpl::Rollback(spanner_proto::TransactionSelector& s) {
...
+  }
+  *request.mutable_key_set() = internal::ToProto(rp.keys);
+  *request.mutable_partition_options() = std::move(partition_options);
+
+  auto context = google::cloud::internal::make_unique<grpc::ClientContext>();
+  auto response = stub_->PartitionRead(*context, request);
+  if (!response.ok()) {
+    return std::move(response).status();
+  }
+
```
devbww:
```
if (s.has_begin()) {
  s.set_id(reponse->transaction().id();
}
```

scotthart: Done.

google/cloud/spanner/connection.h
scotthart: Overload removed. client will pass a ReadParams to Connection::Read.

devbww:
> It would be good to test that the Transaction got an ID after the call (for Read and ExecuteSql and probably others as well), but Transaction doesn't have an API we can use for that?

As mentioned offline, the test could use `Visit()` to create a helper for that.  Not direct support admittedly, but perhaps all we need/want.

devbww:
> the test could use Visit() to create a helper for that

```
std::string IdOf(Transaction txn) {
  return Visit(txn, [](spanner_proto::TransactionSelector& s, std::int64_t) {
    return s.id();
  });          
}
```

devbww:
> > the test could use Visit() to create a helper for that
> 
> ```
> std::string IdOf(Transaction txn) {
>   return Visit(txn, [](spanner_proto::TransactionSelector& s, std::int64_t) {
>     return s.id();
>   });          
> }
> ```

For what it's worth, all the `Read()` and `ExecuteSql()` calls in these tests employ single-use transactions, so there will be no returned transaction id to check.

PR #435, opened 2019-08-24
feat: retry a few times in RunTransaction.

coryan:
The error rate in `RunTransaction` was higher than 1%, and as we
add integration tests this almost guarantees that at least one test
will fail.

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/spanner/client.h"
 #include "google/cloud/spanner/internal/connection_impl.h"
 #include "google/cloud/spanner/internal/spanner_stub.h"
+#include "google/cloud/spanner/retry_policy.h"
 #include "google/cloud/log.h"
 #include <grpcpp/grpcpp.h>
 
@@ -122,8 +123,8 @@ std::shared_ptr<Connection> MakeConnection(
   return std::make_shared<internal::ConnectionImpl>(db, std::move(stub));
 }
 
-StatusOr<CommitResult> RunTransaction(
-    Client client, Transaction::ReadWriteOptions const& opts,
+StatusOr<CommitResult> RunTransactionImpl(
+    Client& client, Transaction::ReadWriteOptions const& opts,
     std::function<StatusOr<Mutations>(Client, Transaction)> const& f) {
   Transaction txn = MakeReadWriteTransaction(opts);
   StatusOr<Mutations> mutations;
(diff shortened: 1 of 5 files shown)


coryan:
I made this change because without it so many CI builds fail that I cannot make any progress.

google/cloud/spanner/client.cc
```diff
@@ -155,6 +158,28 @@ StatusOr<CommitResult> RunTransaction(
...
 }
 
+StatusOr<CommitResult> RunTransaction(
+    Client client, Transaction::ReadWriteOptions const& opts,
+    std::function<StatusOr<Mutations>(Client, Transaction)> const& f) {
+  ExponentialBackoffPolicy backoff_policy(std::chrono::milliseconds(100),
+                                          std::chrono::minutes(5), 2.0);
+  LimitedErrorCountRetryPolicy retry_policy(/*maximum_failures=*/2);
+
+  Status last_status;
```
devbww: Perhaps this shouldn't be initialized to `OK` (maybe `IsExhausted()` is initially true).

google/cloud/spanner/integration_tests/rpc_failure_threshold_integration_test.cc
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  void TearDown() override {
+    if (!db_) {
+      return;
+    }
+    std::cout << "Dropping database " << db_->DatabaseId() << std::flush;
+    DatabaseAdminClient admin_client;
+    auto drop_status = admin_client.DropDatabase(*db_);
+    EXPECT_STATUS_OK(drop_status);
```
devbww: Perhaps these two lines should be exchanged?

google/cloud/spanner/integration_tests/rpc_failure_threshold_integration_test.cc
```diff
@@ -0,0 +1,238 @@
...
+ * This program estimates the error rate for ExecuteSql() when the SQL statement
+ * is 'DELETE FROM table WHERE true'. We expect this to be very low for an empty
+ * table, but was high at some point.
+ *
+ * The program assumes that each run is a independent, identically distributed,
+ * Bernoulli experiment, with an underlying probably of success $p$. The program
+ * computes the 95% confidence interval for $p$ using the normal approximation:
+ *
+ *     https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval
+ *
```
devbww: Nit: Delete extra empty line.

google/cloud/spanner/integration_tests/client_integration_test.cc
```diff
@@ -350,40 +350,49 @@ TEST_F(ClientIntegrationTest, RunTransaction) {
 
 /// @test Test various forms of ExecuteSql()
 TEST_F(ClientIntegrationTest, ExecuteSql) {
```
devbww: I have a similar PR that attempts to purge more of the uses of raw `MakeReadWriteTransaction()` from the integration tests in favor of `RunTransaction()` (there are more in the samples that we can deal with later).  For example, from `SetUp()` and `MultipleInserts()`, in addition to `ExecuteSql()`.  [`TransactionRollback()` probably needs to deal with ABORTED by itself.]  Should we attempt to deal with those cases here?

coryan: Fixed all the ones in `client_integration_test.cc` except for Rollback, I think we should deal with that and the examples in a future PR.

google/cloud/spanner/integration_tests/rpc_failure_threshold_integration_test.cc
coryan: Done.

google/cloud/spanner/integration_tests/rpc_failure_threshold_integration_test.cc
coryan: Done.

google/cloud/spanner/client.cc
coryan: Fixed.

PR #437, opened 2019-08-24
chore: run coverage tests sequentially.

coryan:
The gcov .gcda files for library files are shared across tests, and
written when the test finishes. If the tests are executed in parallel
some of the data may become lost (or the file corrupted). The easiest
fix is to run the tests sequentially. We could use a clever value for
GCOV_PREFIX and then aggregate the results, but that can be a future
improvement.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/437)

### DIFF
--- a/google/cloud/spanner/row_test.cc
+++ b/google/cloud/spanner/row_test.cc
@@ -305,14 +305,14 @@ TEST(Row, ValuesMove) {
   EXPECT_EQ(Value("bar"), values[2]);
 }
 
-TEST(Row, PrinTo) {
+TEST(Row, PrintTo) {
   auto row = MakeRow(1234, "foo", 2345, "bar", "baz");
   std::ostringstream os;
   PrintTo(row, &os);
   auto actual = os.str();
   EXPECT_THAT(actual, HasSubstr("1234"));
-  EXPECT_THAT(actual, HasSubstr("2345"));
   EXPECT_THAT(actual, HasSubstr("foo"));
+  EXPECT_THAT(actual, HasSubstr("2345"));
   EXPECT_THAT(actual, HasSubstr("bar"));
   EXPECT_THAT(actual, HasSubstr("baz"));
 }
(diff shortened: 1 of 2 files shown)


coryan:
PTAL: we still have some code that is obviously tested but is reported as not. At least it is consistently reported this way now.

google/cloud/spanner/row_test.cc
```diff
@@ -284,6 +287,36 @@ TEST(Row, ValuesAccessorRvalue) {
...
+TEST(Row, ValuesMove) {
+  auto row = MakeRow("foo", 12345, "bar");
+  auto values = std::move(row).values();
+  EXPECT_EQ(3, values.size());
+  EXPECT_EQ(Value("foo"), values[0]);
+  EXPECT_EQ(Value(12345), values[1]);
+  EXPECT_EQ(Value("bar"), values[2]);
+}
+
+TEST(Row, PrinTo) {
```
devbww: s/PrintTo/PrintTo/

google/cloud/spanner/row_test.cc
```diff
@@ -284,6 +287,36 @@ TEST(Row, ValuesAccessorRvalue) {
...
+  auto row = MakeRow(1234, "foo", 2345, "bar", "baz");
+  std::ostringstream os;
+  PrintTo(row, &os);
+  auto actual = os.str();
+  EXPECT_THAT(actual, HasSubstr("1234"));
+  EXPECT_THAT(actual, HasSubstr("2345"));
```
devbww: Is the non-column ordering of expectations supposed to indicate anything?

google/cloud/spanner/row_test.cc
coryan: Fixed.

google/cloud/spanner/row_test.cc
coryan: Nope, fixed.

coryan:
PTAL.

PR #471, opened 2019-08-28
feat: Implement a simple stress test.

### LINKED ISSUE #455: Write an integration "stress test"
We should have a little bit of a "stress" test to demonstrate that this library doesn't crumble under a somewhat realistic workload. Perhaps something that does some/all of the following:

1. Insert a N million rows of data.
2. Runs queries in transactions from multiple threads against the table

Maybe also optionally runs for a long time (maybe 1 hour) without blowing up. Running for a long time clearly cannot be done on each PR, but maybe a stress-test like this could be run nightly?

coryan:
This change introduces two tests, both work similarly: (1) they run
multiple threads, each thread runs for N seconds, (2) the thread picks
an action, row key, and query size at random, (3) if the action is
"insert" the thread inserts (or updates) the row with the given key, (4)
if the action is "select" the thread reads (or selects) the randomly
selected query size rows starting at the randomly selected key.

The test counts the number of successes and failures and expects less
than 0.1% failures.

This stress test is executed as part of the CI build, but it only runs
each test for 30 seconds. We should run it manually (or nightly) with a
much higher setting, maybe "24 hours" or something similar.

This (I think, but let me know if you disagree) fixes #455

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/471)

### DIFF
--- a/google/cloud/spanner/integration_tests/CMakeLists.txt
+++ b/google/cloud/spanner/integration_tests/CMakeLists.txt
@@ -27,7 +27,9 @@ function (spanner_client_define_integration_tests)
     find_package(google_cloud_cpp_testing CONFIG REQUIRED)
 
     set(spanner_client_integration_tests
-        client_integration_test.cc database_admin_integration_test.cc
+        client_integration_test.cc
+        database_admin_integration_test.cc
+        client_stress_test.cc
         rpc_failure_threshold_integration_test.cc)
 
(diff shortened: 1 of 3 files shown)


coryan:
@devbww PTAL.

coryan:
@devbww PTAL.

PR #474, opened 2019-08-28
Change SessionHolder to return session as optional.

mr-salty:
This avoids assuming "empty" is "missing" as bww noted.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/474)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -124,7 +124,7 @@ void ConnectionImpl::ReleaseSession(std::string session) {
 StatusOr<ResultSet> ConnectionImpl::Read(SessionHolder& session,
                                          spanner_proto::TransactionSelector& s,
                                          ReadParams rp) {
-  if (session.session_name().empty()) {
+  if (!session) {
     auto session_or = GetSession();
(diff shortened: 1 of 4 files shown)


remyabel:
> I think the current SessionHolder has the advantage of handling moves the way we'd like (the moved-from SessionHolder has no value) whereas an optional<SessionHolder> would still have a value which is a SessionHolder in some abnormal state.

Drive-by opinion: if `SessionHolder` is conceptually a `unique_ptr`, can we give it a `unique_ptr` member and invoke Rule of Zero?

mr-salty:
@remyabel thanks - we discussed that some more and it might be the best option.  I put together #507 as an alternative to this PR.  Having to pass `unique_ptr<SessionHolder>` around is a little ugly but otherwise I think it's an improvement.

coryan:
Should we review this PR or #507?

devjgm:
Is this PR being dropped in favor of #507?

devjgm:
> I think so, I was waiting for votes... so do you agree?
> 
> >

Yes. I think I like the unique_ptr change better.

mr-salty:
closed in favor of #507

PR #489, opened 2019-08-29
 feat: implement Partitioned DML transactions.

### LINKED ISSUE #287: Implement Transaction::PartitionedDmlOptions
We also need to add an overload to `Client::ExecuteSql` to accept an argument of type `Transaction::PartitionedDmlOptions` (just like how SingleUseOptions works today).

NOTE: there are no actual options today, but there might be some someday.

### LINKED ISSUE #432: Implement integration test for Client::ExecuteSql() with PartitionedDmlOptions

coryan:
Add `ExecuteSql()` overload that takes `Transaction::PartitionDmlOptions`.
Also add a unit test (only the success case), an integration test, and a couple
of examples. Applications cannot create a transaction object with these new
options because only one DML statement can be used with such transactions.

This fixes #287 and fixes #432.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/489)

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -257,6 +257,25 @@ class Client {
   //@}
 
   /**
+   * Executes a Partitioned DML SQL query.
+   *
+   * @param statement the SQL statement to execute. Please see the
+   *     [spanner documentation][dml-partitioned] for the restrictions on the
+   *     SQL statements supported by this function.
+   *
+   * @par Example
+   * @snippet samples.cc execute-sql-partitioned
+   *
+   * @see [Partitioned DML Transactions][txn-partitioned] for an overview of
+   *     Partitioned DML transactions.
+   * @see [Partitioned DML][dml-partitioned] for a description of which SQL
+   *     statements are supported in Partitioned DML transactions.
+   * [txn-partitioned]: https://cloud.google.com/spanner/docs/transactions#partitioned_dml_transactions
+   * [dml-partitioned]: https://cloud.google.com/spanner/docs/dml-partitioned
+   */
+  StatusOr<PartitionedDmlResult> ExecutePartitionedDml(SqlStatement statement);
+
+  /**
    * Executes a SQL query on a subset of rows in a database. Requires a prior
    * call to `PartitionQuery` to obtain the partition information; see the
    * documentation of that method for full details.
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -270,7 +270,8 @@ class Client {
    *     Partitioned DML transactions.
    * @see [Partitioned DML][dml-partitioned] for a description of which SQL
    *     statements are supported in Partitioned DML transactions.
-   * [txn-partitioned]: https://cloud.google.com/spanner/docs/transactions#partitioned_dml_transactions
+   * [txn-partitioned]:
+   * https://cloud.google.com/spanner/docs/transactions#partitioned_dml_transactions
    * [dml-partitioned]: https://cloud.google.com/spanner/docs/dml-partitioned
    */
   StatusOr<PartitionedDmlResult> ExecutePartitionedDml(SqlStatement statement);
--- a/google/cloud/spanner/integration_tests/client_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/client_integration_test.cc
@@ -351,6 +351,28 @@ TEST_F(ClientIntegrationTest, ExecuteSql) {
   EXPECT_THAT(actual_rows, UnorderedElementsAreArray(expected_rows));
 }
 
+/// @test Test ExecuteSql(PartitionDmlOptions...)
+TEST_F(ClientIntegrationTest, ExecuteSqlPartitionDml) {
+  auto insert_result = RunTransaction(
+      *client_, {}, [](Client client, Transaction txn) -> StatusOr<Mutations> {
+        auto insert = client.ExecuteSql(
+            std::move(txn), SqlStatement(R"sql(
+        INSERT INTO Singers (SingerId, FirstName, LastName)
+        VALUES (@id, @fname, @lname))sql",
+                                         {{"id", Value(1)},
+                                          {"fname", Value("test-fname-1")},
+                                          {"lname", Value("test-lname-1")}}));
+        if (!insert) return std::move(insert).status();
+        return Mutations{};
+      });
+  EXPECT_STATUS_OK(insert_result);
+
+  auto result = client_->ExecutePartitionedDml(
+      SqlStatement("UPDATE Singers SET LastName = 'test-only'"
+                   " WHERE SingerId >= 1"));
+  EXPECT_STATUS_OK(result);
+}
+
 void CheckReadWithOptions(
     Client client,
     std::function<Transaction::SingleUseOptions(CommitResult const&)> const&
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -311,7 +311,7 @@ StatusOr<PartitionedDmlResult> ConnectionImpl::ExecutePartitionedDml(
   auto response = stub_->ExecuteSql(context, request);
   if (!response) return std::move(response).status();
 
-  PartitionedDmlResult result;
+  PartitionedDmlResult result{0};
   if (response->has_stats()) {
     result.row_count_lower_bound = response->stats().row_count_lower_bound();
   }
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -56,6 +56,15 @@ StatusOr<ResultSet> ConnectionImpl::ExecuteSql(ExecuteSqlParams esp) {
       });
 }
 
+StatusOr<PartitionedDmlResult> ConnectionImpl::ExecutePartitionedDml(
+    ExecuteSqlParams esp) {
+  return internal::Visit(
+      std::move(esp.transaction),
+      [this, &esp](SessionHolder& session,
+                   spanner_proto::TransactionSelector& s, std::int64_t seqno) {
+        return ExecutePartitionedDml(session, s, seqno, std::move(esp));
+      });
+}
 StatusOr<std::vector<QueryPartition>> ConnectionImpl::PartitionQuery(
     PartitionQueryParams pqp) {
   return internal::Visit(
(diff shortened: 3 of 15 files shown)


google/cloud/spanner/transaction.h
```diff
@@ -95,6 +97,23 @@ class Transaction {
...
+  /**
+   * Options for Partitioned DML transactions.
+   *
+   * @see
+   * https://cloud.google.com/spanner/docs/transactions#partitioned_dml_transactions
+   *   for an overview of Partitioned DML transactions.
+   */
+  class PartitionDmlOptions {
+   public:
+    // There are currently no read-write options.
```
devbww: s/read-write/partitioned-dml/

google/cloud/spanner/transaction.h
```diff
@@ -95,6 +97,23 @@ class Transaction {
...
   };
 
+  /**
+   * Options for Partitioned DML transactions.
+   *
+   * @see
+   * https://cloud.google.com/spanner/docs/transactions#partitioned_dml_transactions
+   *   for an overview of Partitioned DML transactions.
+   */
+  class PartitionDmlOptions {
```
devbww: > Apart from the shape of comments, the transaction.{h,cc} changes are identical to mine.

Oh ... except that I called this `Partition*ed*DmlOptions`.

google/cloud/spanner/transaction.h
coryan: Fixed.

google/cloud/spanner/transaction.h
coryan: Fixed.

coryan:
PTAL.

google/cloud/spanner/transaction.cc
```diff
@@ -86,6 +95,12 @@ Transaction::Transaction(ReadWriteOptions opts) {
   impl_ = std::make_shared<internal::TransactionImpl>(std::move(selector));
 }
 
+Transaction::Transaction(PartitionedDmlOptions opts) {
+  google::spanner::v1::TransactionSelector selector;
+  *selector.mutable_begin() = MakeOpts(std::move(opts.pd_opts_));
```
devjgm: I think we should consider keeping "partitioned dml" stuff out of the Transaction classess and the `transaction.*` files. Instead, I think we can simply set the TransactionSelector proto directly in the implementation in `connection_impl.cc`. This would make it so the only place "partitioned dml" is uttered in our user-facing code is in the one function `Client::ExecutePartitionedDml`, which does not even take a `Transaction` nor options. I think the rest of the plumbing provided by the `Transaction` class is not needed for PartitionedDml transactions because they are never exposed to users, are not re-usable, and cannot be committed.

coryan: Done.

coryan:
PTAL.

coryan:
Rebased to pick up less flaky builds.

devjgm:
FWIW, this LGTM. Ping @devbww for approval.

google/cloud/spanner/client.h
```diff
@@ -256,6 +256,26 @@ class Client {
...
+   * Executes a Partitioned DML SQL query.
+   *
+   * @param statement the SQL statement to execute. Please see the
+   *     [spanner documentation][dml-partitioned] for the restrictions on the
+   *     SQL statements supported by this function.
+   *
+   * @par Example
+   * @snippet samples.cc execute-sql-partitioned
+   *
+   * @see [Partitioned DML Transactions][txn-partitioned] for an overview of
```
devbww: Do we need to mention "Partitioned DML Transactions" to the user now?

google/cloud/spanner/client.h
```diff
@@ -256,6 +256,26 @@ class Client {
...
+   *     Partitioned DML transactions.
+   * @see [Partitioned DML][dml-partitioned] for a description of which SQL
```
devbww: Is this redundant with `@param statement`?

google/cloud/spanner/integration_tests/client_integration_test.cc
```diff
@@ -341,6 +341,28 @@ TEST_F(ClientIntegrationTest, ExecuteSql) {
...
+                                          {"lname", Value("test-lname-1")}}));
+        if (!insert) return std::move(insert).status();
+        return Mutations{};
+      });
+  EXPECT_STATUS_OK(insert_result);
+
+  auto result = client_->ExecutePartitionedDml(
+      SqlStatement("UPDATE Singers SET LastName = 'test-only'"
+                   " WHERE SingerId >= 1"));
+  EXPECT_STATUS_OK(result);
```
devbww: Should we try to read this back, like the other mutating cases?

google/cloud/spanner/internal/partial_result_set_reader.cc
```diff
@@ -83,10 +83,15 @@ StatusOr<optional<Value>> PartialResultSetReader::NextValue() {
 
 PartialResultSetReader::~PartialResultSetReader() {
   if (!finished_) {
+    // If there is actual data in the streaming RPC Finish() can deadlock, so
```
devbww: Should the edits in this file be in a separate PR?

google/cloud/spanner/partitioned_dml_result.h
```diff
@@ -0,0 +1,39 @@
...
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+/**
+ * The result of executing a Partitioned DML query.
+ */
+struct PartitionedDmlResult {
+  /// A lower bound on the number of rows modified.
+  std::int64_t row_count_lower_bound;
```
devbww: `#include <cstdint>`

google/cloud/spanner/partitioned_dml_result.h
```diff
@@ -0,0 +1,39 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_PARTITIONED_DML_RESULT_H_
+#define GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_PARTITIONED_DML_RESULT_H_
+
+#include "google/cloud/spanner/timestamp.h"
```
devbww: Unused.

google/cloud/spanner/samples/samples.cc
```diff
@@ -391,6 +391,27 @@ void DmlStandardDelete(google::cloud::spanner::Client client) {
...
+  if (!result) throw std::runtime_error(result.status().message());
+  std::cout << "Delete was successful [spanner_dml_partitioned_delete]\n";
+}
+//! [execute-sql-partitioned] [END spanner_dml_partitioned_delete]
+
+//! [START spanner_dml_partitioned_update]
+void DmlPartitionedUpdate(google::cloud::spanner::Client client) {
+  namespace spanner = google::cloud::spanner;
+  auto result = client.ExecutePartitionedDml(
+      spanner::SqlStatement("UPDATE Albums SET MarketingBudget = 100000"
```
devbww: How does this sample know `MarketingBudget` exists?

google/cloud/spanner/samples/samples.cc
```diff
@@ -680,6 +703,16 @@ void RunAll() {
   std::cout << "\nRunning spanner_field_access_on_nested_struct sample\n";
   FieldAccessOnNestedStruct(client);
 
+  namespace s = google::cloud::spanner;
+  s::Client c2(s::MakeConnection(
+      s::Database(project_id, instance_id, database_id),
+      s::ConnectionOptions().enable_clog().enable_tracing("rpc")));
```
devbww: Is this a "submit" thing or a "development testing" thing?

coryan: Ooops, fixed, thanks!

google/cloud/spanner/samples/samples.cc
coryan: The same way all the examples know that the tables exist, or any other column. They need to be executed in a particular order.

google/cloud/spanner/partitioned_dml_result.h
coryan: Removed.

google/cloud/spanner/partitioned_dml_result.h
coryan: Added.

google/cloud/spanner/internal/partial_result_set_reader.cc
coryan: Maybe... It hangs without them, so we need them before or with this PR.

google/cloud/spanner/integration_tests/client_integration_test.cc
coryan: Fixed.

google/cloud/spanner/client.h
coryan: Removed.

google/cloud/spanner/client.h
coryan: Removed.

google/cloud/spanner/samples/samples.cc
```diff
@@ -703,15 +703,11 @@ void RunAll() {
   std::cout << "\nRunning spanner_field_access_on_nested_struct sample\n";
   FieldAccessOnNestedStruct(client);
 
-  namespace s = google::cloud::spanner;
-  s::Client c2(s::MakeConnection(
-      s::Database(project_id, instance_id, database_id),
-      s::ConnectionOptions().enable_clog().enable_tracing("rpc")));
-  std::cout << "\nRunning spanner_dml_partitioned_update sample\n";
-  DmlPartitionedUpdate(c2);
+  std::cout << "\nRunning spanner_dml_partitioned_update sample\n" << std::endl;
```
devbww: `<< "\n" << std::endl` ?

google/cloud/spanner/samples/samples.cc
```diff
@@ -703,15 +703,11 @@ void RunAll() {
...
+  DmlPartitionedUpdate(client);
 
   std::cout << "\nRunning spanner_dml_partitioned_delete sample\n" << std::endl;
```
devbww: Ditto.

google/cloud/spanner/internal/partial_result_set_reader.cc
devbww: OK.  I'm fine with them here, but let's tag @mr-salty so he knows.

google/cloud/spanner/samples/samples.cc
devbww: But `samples.cc` also has `RunOneCommand` support.

google/cloud/spanner/samples/samples.cc
coryan: Fixed.

google/cloud/spanner/samples/samples.cc
coryan: Fixed.

google/cloud/spanner/samples/samples.cc
coryan: Yes. And it was true for other existing examples that use `MarketingBudget`, and it is also true in the samples for other languages (and libraries).  I get the point, it means the samples can fail if you don't setup your environment "just so", but (I think) this is not the PR to fix that (assuming we want to fix it).

devbww: Fair enough, but I wasn't suggesting this was the PR to fix it ... just the PR not to introduce another occurrence.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -265,6 +276,52 @@ StatusOr<ResultSet> ConnectionImpl::ExecuteSql(
   return ResultSet(std::move(*reader));
 }
 
+StatusOr<PartitionedDmlResult> ConnectionImpl::ExecutePartitionedDml(
+    SessionHolder& session, google::spanner::v1::TransactionSelector& s,
+    std::int64_t seqno, ExecutePartitionedDmlParams epdp) {
+  if (session.session_name().empty()) {
+    // Since the session may be sent to other machines, it should not be
+    // returned to the pool when the Transaction is destroyed (release=true).
+    auto session_or = GetSession(/*release=*/true);
```
mr-salty: I think this is a bug - PartitionedDml transactions are not partitioned among clients like the other Partition* cases, right?  @scotthart can you confirm?

mr-salty: FYI I sent #563 - we can continue the discussion there.

PR #503, opened 2019-08-29
chore: add samples for keyset, partition read, and partition query

### LINKED ISSUE #328: chore: migrate example code in KeySet to use snippet
Example code should be put in a separate file and compiled/tested and then included in doxygen via '@snippet'

### LINKED ISSUE #388: chore: convert example code in ReadPartition to use snippet

scotthart:
fixes #328 
fixes #388

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/503)

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -179,6 +179,9 @@ class Client {
    * @return A `StatusOr` containing a `ResultSet` or error status on failure.
    *     No individual row in the `ResultSet` can exceed 100 MiB, and no column
    *     value can exceed 10 MiB.
+   *
+   * @par Example
+   * @snippet samples.cc read-read-partition
    */
   // TODO(#409): possibly change to pass by value when issue resolved.
   StatusOr<ResultSet> Read(ReadPartition const& partition);
@@ -211,6 +214,9 @@ class Client {
    *
    * @return A `StatusOr` containing a vector of `ReadPartition` or error
    *     status on failure.
+   *
+   * @par Example
+   * @snippet samples.cc partition-read
    */
   StatusOr<std::vector<ReadPartition>> PartitionRead(
       Transaction transaction, std::string table, KeySet keys,
@@ -286,6 +292,9 @@ class Client {
    * @return A `StatusOr` containing a `ResultSet` or error status on failure.
    *     No individual row in the `ResultSet` can exceed 100 MiB, and no column
    *     value can exceed 10 MiB.
+   *
+   * @par Example
+   * @snippet samples.cc execute-sql-query-partition
    */
   StatusOr<ResultSet> ExecuteSql(QueryPartition const& partition);
 
@@ -306,6 +315,9 @@ class Client {
    *
    * @return A `StatusOr` containing a vector of `QueryPartition`s or error
    *     status on failure.
+   *
+   * @par Example
+   * @snippet samples.cc partition-query
    */
   StatusOr<std::vector<QueryPartition>> PartitionQuery(
       Transaction transaction, SqlStatement statement,
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -571,6 +571,118 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
 }
 //! [END spanner_field_access_on_nested_struct]
 
+void Send(std::string const&) {}
+
+void SendPartitionToRemoteMachine(
+    google::cloud::spanner::ReadPartition const& partition) {
+  //! [serialize-read-partition]
+  namespace spanner = google::cloud::spanner;
+  google::cloud::StatusOr<std::string> serialized_partition =
+      spanner::SerializeReadPartition(partition);
+  if (!serialized_partition) {
+    throw std::runtime_error(serialized_partition.status().message());
+  }
+  Send(*serialized_partition);
+  //! [serialize-read-partition]
+}
+
+void SendPartitionToRemoteMachine(
+    google::cloud::spanner::QueryPartition const& partition) {
+  //! [serialize-query-partition]
+  namespace spanner = google::cloud::spanner;
+  google::cloud::StatusOr<std::string> serialized_partition =
+      spanner::SerializeQueryPartition(partition);
+  if (!serialized_partition) {
+    throw std::runtime_error(serialized_partition.status().message());
+  }
+  Send(*serialized_partition);
+  //! [serialize-query-partition]
+}
+
+std::string Receive() { return std::string(); }
+
+//! [deserialize-read-partition]
+google::cloud::StatusOr<google::cloud::spanner::ReadPartition>
+ReceiveReadPartitionFromRemoteMachine() {
+  std::string serialized_partition = Receive();
+  return google::cloud::spanner::DeserializeReadPartition(serialized_partition);
+}
+//! [deserialize-read-partition]
+
+//! [deserialize-query-partition]
+google::cloud::StatusOr<google::cloud::spanner::QueryPartition>
+ReceiveQueryPartitionFromRemoteMachine() {
+  std::string serialized_partition = Receive();
+  return google::cloud::spanner::DeserializeQueryPartition(
+      serialized_partition);
+}
+//! [deserialize-query-partition]
+
+void ProcessRow(google::cloud::spanner::Row<std::int64_t, std::string,
+                                            std::string> const&) {}
+
+void PartitionRead(google::cloud::spanner::Client client) {
+  namespace spanner = google::cloud::spanner;
+  //! [key-set-builder]
+  auto key_set = spanner::KeySetBuilder<spanner::Row<int64_t>>()
+                     .Add(spanner::MakeKeyRangeClosed(spanner::MakeRow(1),
+                                                      spanner::MakeRow(10)))
+                     .Build();
+  //! [key-set-builder]
+
+  //! [partition-read]
+  spanner::Transaction ro_transaction = spanner::MakeReadOnlyTransaction();
+  google::cloud::StatusOr<std::vector<spanner::ReadPartition>> partitions =
+      client.PartitionRead(ro_transaction, "Singers", key_set,
+                           {"SingerId", "FirstName", "LastName"});
+  if (!partitions) throw std::runtime_error(partitions.status().message());
+  for (auto& partition : *partitions) {
+    SendPartitionToRemoteMachine(partition);
+  }
+  //! [partition-read]
+
+  //! [read-read-partition]
+  google::cloud::StatusOr<spanner::ReadPartition> partition =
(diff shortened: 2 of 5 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+void Send(std::string) {}  // NOLINT
+
+void SendPartitionToRemoteMachine(
+    google::cloud::spanner::ReadPartition const& partition) {
+  //! [serialize-read-partition]
+  namespace spanner = google::cloud::spanner;
+  google::cloud::StatusOr<std::string> serialized_partition =
+      spanner::SerializeReadPartition(partition);
+  if (serialized_partition.ok()) {
+    Send(*serialized_partition);
```
devjgm: (optional) FWIW, I'd write things like this with a comment instead of a dummy `Send()` function. Maybe something like

```cc
if (serialized_partition) {
  // Use *serialized_partition, write to disk, send to another machine, etc.
}
```
I think that comment would convey more info than `Send()` and it might avoid the need to NOLINT.

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
 }
 //! [END spanner_field_access_on_nested_struct]
 
+void Send(std::string) {}  // NOLINT
```
devjgm: Maybe accept by const-ref here instead of the NOLINT?

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+  namespace spanner = google::cloud::spanner;
+  google::cloud::StatusOr<std::string> serialized_partition =
+      spanner::SerializeQueryPartition(partition);
+  if (serialized_partition.ok()) {
+    Send(*serialized_partition);
+  }
+  //! [serialize-query-partition]
+}
+
+std::string Receive() { return std::string(); }  // NOLINT
```
devjgm: What's the lint complaint? Does `return {}` fix it?

google/cloud/spanner/client.h
```diff
@@ -176,6 +176,10 @@ class Client {
    * @return A `StatusOr` containing a `ResultSet` or error status on failure.
    *     No individual row in the `ResultSet` can exceed 100 MiB, and no column
    *     value can exceed 10 MiB.
+   *
+   * @par Example
+   * @snippet samples.cc read-read-partition
+
```
coryan: nit: remove blank line.

google/cloud/spanner/client.h
```diff
@@ -263,6 +270,10 @@ class Client {
    * @return A `StatusOr` containing a `ResultSet` or error status on failure.
    *     No individual row in the `ResultSet` can exceed 100 MiB, and no column
    *     value can exceed 10 MiB.
+   *
+   * @par Example
+   * @snippet samples.cc execute-sql-query-partition
+
```
coryan: nit: remove blank line.

google/cloud/spanner/samples/samples.cc
coryan: +1

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+    for (auto& partition : *partitions) {
+      SendPartitionToRemoteMachine(partition);
+    }
+  }
+  //! [partition-query]
+
+  //! [execute-sql-query-partition]
+  google::cloud::StatusOr<spanner::QueryPartition> partition =
+      ReceiveQueryPartitionFromRemoteMachine();
+  if (partition) {
```
coryan: I think this is easier to grok if you do:

```C++
if (!partition) throw std::runtime_error(parition.status().message());
```

That reduces nesting and is similar to the error handling in other examples.

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+    auto result_set = client.ExecuteSql(*partition);
+    if (result_set) {
```
coryan: Same here about error handling.

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+      for (auto& row :
+           result_set->Rows<std::int64_t, std::string, std::string>()) {
+        if (row) {
```
coryan: Ditto about error handling.

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+                                                      spanner::MakeRow(10)))
+                     .Build();
+  //! [key-set-builder]
+
+  //! [partition-read]
+  spanner::Transaction ro_transaction = spanner::MakeReadOnlyTransaction();
+  google::cloud::StatusOr<std::vector<spanner::ReadPartition>> partitions =
+      client.PartitionRead(ro_transaction, "Singers", key_set,
+                           {"SingerId", "FirstName", "LastName"});
+  if (partitions) {
```
coryan: See other comments regarding error handling.

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+    for (auto& partition : *partitions) {
+      SendPartitionToRemoteMachine(partition);
+    }
+  }
+  //! [partition-read]
+
+  //! [read-read-partition]
+  google::cloud::StatusOr<spanner::ReadPartition> partition =
+      ReceiveReadPartitionFromRemoteMachine();
+  if (partition) {
```
coryan: See other comments regarding error handling.

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+    auto result_set = client.Read(*partition);
+    if (result_set) {
```
coryan: See other comments regarding error handling.

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+      for (auto& row :
+           result_set->Rows<std::int64_t, std::string, std::string>()) {
+        if (row) {
```
coryan: See other comments regarding error handling.

google/cloud/spanner/samples/samples.cc
```diff
@@ -569,6 +689,9 @@ int RunOneCommand(std::vector<std::string> argv) {
                          &FieldAccessOnStructParameters),
       make_command_entry("field-access-on-nested-struct",
                          &FieldAccessOnNestedStruct),
+      make_command_entry("partition-read", &PartitionRead),
```
coryan: After looking at the code, it seems complex enough that we should run it.

google/cloud/spanner/samples/samples.cc
```diff
@@ -527,6 +527,126 @@ void FieldAccessOnNestedStruct(google::cloud::spanner::Client client) {
...
+
+void PartitionQuery(google::cloud::spanner::Client client) {
+  namespace spanner = google::cloud::spanner;
+  //! [partition-query]
+  spanner::Transaction ro_transaction = spanner::MakeReadOnlyTransaction();
+  google::cloud::StatusOr<std::vector<spanner::QueryPartition>> partitions =
+      client.PartitionQuery(
+          ro_transaction,
+          spanner::SqlStatement(
+              "select SingerId, FirstName, LastName from Singers"));
```
coryan: In other examples we have used `SELECT SingerId, FirstName, LastName FROM Singers`. Do we want to capitalize the SQL keywords?

google/cloud/spanner/samples/samples.cc
coryan: I am going to suggest that we need something more than that.  How do I "save" one of these serialized partitions?  Is the string safe to save and read via iostreams?  Does it contain non-ASCII characters? What about NUL characters?

It would be nice to show how would applications store the string somewhere using standard C++.

coryan:
___
*[google/cloud/spanner/samples/samples.cc, line 539 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/503#-LnTxSV0J5sv-PAN0T4R-r1-539:-LnyAuqT4nmhTVAkp5ut:b-hnxhhk) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/a07dc5612e35aca1acabeaf480b768429d394273/google/cloud/spanner/samples/samples.cc#L539)):*

Under the hood SerializeReadPartition is calling protobuf::MessageLite::SerializeToString which has this disclaimer: "bool SerializeToString(string* output) const;: serializes the message and stores the bytes in the given string. Note that the bytes are binary, not text; we only use the string class as a convenient container."

From that, I'm inferring that its safe to pass around, just don't look inside it.
</blockquote></details>

I am not sure you can conclude that:

- `is >> actual` would stop on the first space character, which is not what we want here.
- The treatment of NUL characters is tricky when you move a string to a iostream, a file or to another representation.
- If you externalize the string to a file then you better open the file with `std::ios::binary` or (at least on Windows) there are automatic conversions that are not easy to reverse.

coryan:
___
*[google/cloud/spanner/samples/samples.cc, line 539 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/503#-LnTxSV0J5sv-PAN0T4R-r1-539:-LnyBZ-l8h0tuD-DEZ1b:bdx164l) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/a07dc5612e35aca1acabeaf480b768429d394273/google/cloud/spanner/samples/samples.cc#L539)):*

I am not sure you can conclude that:

- `is >> actual` would stop on the first space character, which is not what we want here.
- The treatment of NUL characters is tricky when you move a string to a iostream, a file or to another representation.
- If you externalize the string to a file then you better open the file with `std::ios::binary` or (at least on Windows) there are automatic conversions that are not easy to reverse.


</blockquote></details>

BTW, if you want to commit this as-is and then open a bug for "Create an example / documentation on how to persist serialized partitions" then I would be Okay with that.

PR #505, opened 2019-08-29
Fix a reversed comparison that causes Session leaks

mr-salty:
I'm somewhat disturbed no tests failed, so I'll plan to write something
to detect session leaks... but in the meantime I think we should fix
this.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/505)


mr-salty:
heh, actually I was on my way out of the building and realized that the existing code was correct and this PR introduces the bug I thought I was fixing...  apparently my brain is already on a long weekend, and I probably need to rename `release` - if the author of the code can't comprehend it a day later it's probably not a good name ;)

So, this PR is moot - #506 may still be an issue though - certainly no unit tests failed with this change, I'll see if any integration tests fail.

PR #507, opened 2019-08-30
Wrap SessionHolder in a unique_ptr

mr-salty:
This makes "no session" a distinct case (nullptr) instead of being a
special case inside SessionHolder. Now if SessionHolder exists, the
session_name() is always valid.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/507)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -27,13 +27,15 @@ namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
 namespace internal {
 
+using ::google::cloud::internal::make_unique;
+
 namespace spanner_proto = ::google::spanner::v1;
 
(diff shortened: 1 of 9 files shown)


google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -450,11 +462,14 @@ StatusOr<SessionHolder> ConnectionImpl::GetSession(bool release) {
...
   }
 
-  return release ? SessionHolder(std::move(session_name), /*deleter=*/nullptr)
-                 : SessionHolder(std::move(session_name),
-                                 [this](std::string session) {
-                                   ReleaseSession(std::move(session));
-                                 });
+  // TODO(#409) take a (weak) reference to `this` to avoid use-after-free.
+  auto deleter = dissociate_from_pool
+                     ? std::function<void(std::string)>(nullptr)
```
devbww: Just use the default ctor?

PR #510, opened 2019-08-30
bug: RunTransaction uses a time-based retry policy.

### LINKED ISSUE #508: TransactionRunner retry limit should be based on time rather than count
From https://cloud.google.com/spanner/docs/reference/rest/v1/TransactionOptions#transaction-modes

"Under some circumstances (e.g., many transactions attempting to modify the same row(s)), a transaction can abort many times in a short period before successfully committing. Thus, it is not a good idea to cap the number of retries a transaction can attempt; instead, it is better to limit the total amount of wall time spent retrying."

More immediately, this may help reduce test flakiness

coryan:
Apparently errors in quick succession are common, and the recommended
practice is to limit the number of failures based on a time limit rather
than a count. This PR changes `RunTransaction` to use a
`LimitedTimeRetryPolicy` by default. This change also refactors
`RunTransaction` to `internal::RunTransactionWithPolicies` so our tests
can verify the function runs without having to wait for multiple seconds
(or minutes).

This fixes #508

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/510)

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -15,6 +15,7 @@
 #ifndef GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_CLIENT_H_
 #define GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_CLIENT_H_
 
+#include "google/cloud/spanner/backoff_policy.h"
 #include "google/cloud/spanner/batch_dml_result.h"
 #include "google/cloud/spanner/commit_result.h"
 #include "google/cloud/spanner/connection.h"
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/client.h
```diff
@@ -371,6 +373,42 @@ class Client {
...
+ *
+ * The lock priority of the transaction increases after each prior aborted
+ * transaction, meaning that the next attempt has a slightly better chance
+ * of success than before.
+ *
+ * @param client how to contact Cloud Spanner
+ * @param opts options for the transaction created by this function
+ * @param f the function to call in the transaction.
+ * @param retry_policy controls for how long (or how many times) will the
+ *     function retry the operation when there is a retryable failure.
```
devbww: Just to reiterate my opinion ... I don't think `RunTransaction()` should be retrying anything but `ABORTED`.  Everything else is the responsibility of whatever retries normal (non-`RunTransaction()`) calls.

mr-salty: I think I agree with that, although this PR doesn't change that behavior AFAICT?
Can you open an issue?

devbww: Done.  #529

PR #513, opened 2019-08-30
bug: make rollback test less flaky.

coryan:
The integration test for `Client::Rollback()` flakes and cannot use
RunTransaction(). This change runs the part of the test that has been
flaky in a custom loop, and then calls `Rollback()` explicitly.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/513)

### DIFF
--- a/google/cloud/spanner/integration_tests/client_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/client_integration_test.cc
@@ -179,58 +179,62 @@ TEST_F(ClientIntegrationTest, TransactionRollback) {
 
   // Cannot use RunTransaction in this test because we want to call Rollback
   // explicitly.
-  auto txn = MakeReadWriteTransaction();
-  auto insert1 = client_->ExecuteSql(
-      txn, SqlStatement("INSERT INTO Singers (SingerId, FirstName, LastName) "
-                        "VALUES (@id, @fname, @lname)",
-                        {{"id", Value(3)},
-                         {"fname", Value("test-fname-3")},
-                         {"lname", Value("test-lname-3")}}));
-  EXPECT_STATUS_OK(insert1);
-  auto insert2 = client_->ExecuteSql(
-      txn, SqlStatement("INSERT INTO Singers (SingerId, FirstName, LastName) "
-                        "VALUES (@id, @fname, @lname)",
-                        {{"id", Value(4)},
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/integration_tests/client_integration_test.cc
```diff
@@ -179,58 +179,62 @@ TEST_F(ClientIntegrationTest, TransactionRollback) {
...
-              UnorderedElementsAre(RowType(1, "test-fname-1", "test-lname-1"),
+    ASSERT_THAT(returned_rows, UnorderedElementsAre(
+                                   RowType(1, "test-fname-1", "test-lname-1"),
                                    RowType(2, "test-fname-2", "test-lname-2"),
                                    RowType(3, "test-fname-3", "test-lname-3"),
                                    RowType(4, "test-fname-4", "test-lname-4")));
+    break;
+  }
 
   auto insert_rollback_result = client_->Rollback(txn);
```
devbww: I'd probably move this inside the loop (before the final `break`) to make `txn` loop local (and so avoid its strange outside-loop initialization).

google/cloud/spanner/integration_tests/client_integration_test.cc
```diff
@@ -179,58 +179,62 @@ TEST_F(ClientIntegrationTest, TransactionRollback) {
...
-  auto reader = client_->Read(txn, "Singers", KeySet::All(),
-                              {"SingerId", "FirstName", "LastName"});
-  EXPECT_STATUS_OK(reader);
-
   using RowType = Row<std::int64_t, std::string, std::string>;
   std::vector<RowType> returned_rows;
-  if (reader) {
+
+  Transaction txn = MakeReadOnlyTransaction();
+  for (auto start = std::chrono::steady_clock::now(),
```
devbww: My version has a breadcrumb comment here of ...
```
  // We do not use RunTransaction() here as we want to call Rollback()
  // directly. This means we need to retry ABORTED calls ourselves.
```
and (along the lines of previous comments on the matter) my loop only `continue`s on `ABORTED` calls, ala ...
```
  if (!insert1 && insert1.status().code() == StatusCode::kAborted) continue;
  EXPECT_STATUS_OK(insert1);
```

Issue #224, opened 2019-07-16
Fix user-agent string for Spanner.


coryan:
We are using the build information for `google-cloud-cpp` we should capture our own build information.

PR #535, opened 2019-09-04
feat: populate the user-agent string.

coryan:
Populate the user-agent string with information about the library
version. Also allow applications to add their own user-agent string
prefixes.

This is part of the work for #224

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/535)

### DIFF
--- a/google/cloud/spanner/connection_options.h
+++ b/google/cloud/spanner/connection_options.h
@@ -86,6 +86,14 @@ class ConnectionOptions {
     return *this;
   }
 
+  /**
+   * Prepend @p prefix to the user-agent string.
+   *
+   * Libraries or services that use the Cloud Spanner C++ client may want to
+   * set their own user-agent prefix. This can help them develop telemetry
+   * information about number of users running particular versions of their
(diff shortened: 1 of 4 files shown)


google/cloud/spanner/connection_options.h
```diff
@@ -82,12 +86,24 @@ class ConnectionOptions {
     return *this;
   }
 
+  ConnectionOptions& add_user_agent_prefix(std::string prefix) {
```
devjgm: What do users use this for? Is it documented somewhere how a user could/should use the connection's user-agent prefix?

coryan: The typical use-case is a library built on top of the spanner client (think TensorFlow, or something like [SOCI](https://github.com/SOCI/soci) or [sqlpp11](https://github.com/rbock/sqlpp11)), or a service built on top of our library (don't know, maybe Cloud Dataflow or something similar).

They would want to add their own user-agent prefix for the same reasons we do.

PR #568, opened 2019-09-06
doc: document GetElement<I>(Row).

### LINKED ISSUE #554: Document spanner::GetElement<>
The function has no documentation:

https://googleapis.github.io/google-cloud-cpp-spanner/latest/namespacegoogle_1_1cloud_1_1spanner_1_1v0.html#a76a64d3ee14907e69b4f3d415a4c3ab5

https://googleapis.github.io/google-cloud-cpp-spanner/latest/annotated.html

### LINKED ISSUE #559: nit: fix notes with a colon
We have some documents that say `@note:` (note the trailing column), that does not render as one may think it does.

coryan:
Fixes #559 and fixes #554.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/568)

### DIFF
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -186,7 +186,7 @@ class Row {
   /**
    * Returns a std::array of `Value` objects holding all the items in this row.
    *
-   * @note: If the Row's values are large, it may be move efficient to "move"
+   * @note If the Row's values are large, it may be move efficient to "move"
    *     them into the returned array.
    *
    * @par Example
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/row.h
```diff
@@ -272,9 +272,16 @@ class Row {
   std::tuple<Types...> values_;
 };
 
-// Returns the row-element at column `I`. This function will be found via ADL
-// to make `Row<Ts...>` work with `internal::ForEach`. Users should not call
-// this function directly; instead, simply call `row.get<I>()`.
+/**
+ * Returns the row-element at column `I`.
+ *
+ * @note Library users should not need to call this function directly; instead,
```
devjgm: This is really not intended to be part of the public API. It exists in this namespace because ADL requires that. I'm concerned that documenting like this makes it look like a public part, but it's not.

Does doxygen give us any way to *remove* a function from the generated docs?

coryan: > 
> Does doxygen give us any way to _remove_ a function from the generated docs?

Yes. I added `GetElement` to the excluded symbols.  I kept the doxygen comments because I want (someday) to generate a "developer's version" of the Doxygen docs.

> This is really not intended to be part of the public API.

We may not want it to be, but it *is* part of the public API.  It is in the same namespace we put all the public functions.

google/cloud/spanner/row.h
```diff
@@ -186,7 +186,7 @@ class Row {
   /**
    * Returns a std::array of `Value` objects holding all the items in this row.
    *
-   * @note: If the Row's values are large, it may be move efficient to "move"
+   * @note If the Row's values are large, it may be move efficient to "move"
```
devbww: s/move efficient/more efficient/

PR #570, opened 2019-09-06
doc: Document namespaces.

coryan:
A few of the namespaces did not have documentation (or the documentation
was slightly off).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/570)

### DIFF
--- a/google/cloud/spanner/testing/database_environment.h
+++ b/google/cloud/spanner/testing/database_environment.h
@@ -22,7 +22,9 @@
 
 namespace google {
 namespace cloud {
+/// Helper types and functions used in Cloud Spanner C++ client library tests.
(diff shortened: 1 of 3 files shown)


coryan:
*[a discussion](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/570#-Lo6RWW--CIyhANWbvv4:-Lo6XYmm3L4YtX96aNTy:bchzxko) (no related file):*

Are we anticipating documenting all the namespaces, in this fashion, for all external headers?
</blockquote></details>

No, each namespace needs to be documented only once.  If we don't find a convenient header to document them we can create a `.dox` file with the documentation.

---

PR #572, opened 2019-09-06
cleanup: refactor mocks to new namespace and directory

### LINKED ISSUE #569: Separate the mocks from other testing helpers.
We want to export the mocks for use by application developers, while the testing helpers are just for our own use.

We should separate the mocks to `google::cloud::spanner_mocks`, in the `google/cloud/spanner/mocks` directory.

They should be in a new library too, because they get installed, but the testing code should not.

coryan:
Move the mocks to `google::cloud::spanner_mocks::` (and
`google/cloud/spanner/mocks`). Also created a new library for the mocks,
which we will probably want to install.

This fixes #569

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/572)


coryan:
Hold on on any reviews... I am changing my mind about *which* mocks should be refactored.

coryan:
Okay, looks good now. PTAL.

Issue #573, opened 2019-09-06
[Flake]: Windows integration tests fail if run in parallel.


coryan:
I forgot about this:  grpc/grpc#16872

We cannot run the integration tests in parallel on Windows because of that bug, we just one such failure:

https://source.cloud.google.com/results/invocations/990b3e97-c90f-4722-97e2-d8784e6986d4/targets

We should run the tests serially.

coryan:
Fixed in #578.

coryan:
It seems that the fix did not work:

https://source.cloud.google.com/results/invocations/55a6e5d3-7dcb-409f-beaa-9790c55a5069/targets

coryan:
And another one:

https://source.cloud.google.com/results/invocations/6ce70af6-30dc-465f-a3a3-633ef875bc87/targets

tmatsuo:
https://source.cloud.google.com/results/invocations/6ce70af6-30dc-465f-a3a3-633ef875bc87/targets

tmatsuo:
Oh sorry I pasted the same thing

coryan:
And another one:

https://source.cloud.google.com/results/invocations/e5c1d129-e0cb-4ec1-abe6-1ad63f98f7b8

tmatsuo:
Another one:
https://source.cloud.google.com/results/invocations/4eb9bd16-0c8d-4928-9fc0-c4070c972859/targets

tmatsuo:
Another one:
https://source.cloud.google.com/results/invocations/c0e08b88-1fb2-4b0d-8b34-361cd8ad105a/targets

coryan:
Another repeat: 

https://source.cloud.google.com/results/invocations/e727b8c9-c66f-4a50-8342-1adc2e0cfbc5

coryan:
I am not actively working on this.  I think grpc/grpc#20565 may help with this, but it is a few weeks before that PR gets included in a release.

coryan:
FWIW, gRPC-1.26.0 (which includes the fix referenced above) is released and it is (now) the version included with vcpkg, meaning that at least our CMake-based builds may be executed in parallel.

devbww:
@coryan  ... do we think this is fixed now?  (This issue went out of SLO yesterday.)  Thanks.

coryan:
I am going to say yes.

PR #595, opened 2019-09-09
test: Build files for Windows CMake

tmatsuo:
At least it compiles and unit tests are green.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/595)


google/cloud/spanner/value.h
```diff
@@ -523,7 +523,7 @@ class Value {
     google::spanner::v1::Type const& type;
     template <typename T>
     void operator()(T& t) {
-      auto value = GetValue(T{}, list_value.values(i), type);
+      auto value = GetValue(T{}, list_value.values(static_cast<int>(i)), type);
```
devjgm: I think the declaration of `ExtractTupleValues::i` on line 521 should probably declare that member as type `int` rather than `std::size_t`. size_t *seems* like the right type, but since this is a protobuf, the actual right type is `int`.

So I think if you change that declaration, you won't need these casts, and the code should work fine and indeed, probably be slightly more correct (since these are protos).

google/cloud/spanner/value.cc
```diff
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// To avoid secure warnings on Windows
+#ifdef _MSC_VER
+#define _CRT_SECURE_NO_WARNINGS
```
devjgm: Sorry, but I don't know much about this macro. What warnings is this disabling on windows? Is it something we *must* disable, or are there other ways that we should address those warnings?

tmatsuo: It complains as follows:
```
'scanf': This function or variable may be unsafe. Consider using scanf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
```

google/cloud/spanner/value.h
tmatsuo: Thanks! Done

ci/kokoro/windows/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+set(GOOGLE_CLOUD_CPP_SPANNER_ROOT "${PROJECT_SOURCE_DIR}/../../..")
+list(APPEND CMAKE_MODULE_PATH "${GOOGLE_CLOUD_CPP_SPANNER_ROOT}/cmake")
+
+include(external/external-project-helpers)
+google_cloud_cpp_set_prefix_vars()
+
+include(ExternalProject)
+set(
+    GOOGLE_CLOUD_CPP_URL
+    "https://github.com/googleapis/google-cloud-cpp/archive/84772e182e5a82c9f839a9073cef88b9962c48cf.tar.gz"
```
devjgm: This is a different version than we're using in our other cmake builds:
https://github.com/googleapis/google-cloud-cpp-spanner/blob/fd3730b06613aa2a4db43e5845d7334c9c3aca13/cmake/external/google-cloud-cpp.cmake#L29

I'm curious: Can we use that same `cmake/external/google-cloud-cpp.cmake` file from this windows build too?

google/cloud/spanner/value.cc
coryan: Here and above.  I think we can disable the warning for specific files in the CMake code instead:

https://github.com/googleapis/google-cloud-cpp/blob/84772e182e5a82c9f839a9073cef88b9962c48cf/google/cloud/storage/CMakeLists.txt#L243-L251

google/cloud/spanner/internal/date.cc
```diff
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// To avoid secure warnings on Windows
```
coryan: I think we can disable the warning for specific files in the CMake code instead:

https://github.com/googleapis/google-cloud-cpp/blob/84772e182e5a82c9f839a9073cef88b9962c48cf/google/cloud/storage/CMakeLists.txt#L243-L251

ci/kokoro/windows/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+google_cloud_cpp_set_prefix_vars()
+
+ExternalProject_Add(
+    # Use very short names for the project. On Windows the path limit is about
+    # 250 characters. The project name and prefix become part of the path, and
+    # in the case of googleapis (a sub-project of `google-cloud-cpp`) using
+    # full project names creates target files that exceed Windows limit.
+    g-c-cpp
+    EXCLUDE_FROM_ALL ON
+    PREFIX "${CMAKE_BINARY_DIR}/e/g-c-cpp"
```
coryan: Do we need a separate CMake file to use shorter paths?  Any other reason?  If it is just for the shorter paths then I think I prefer to first change the current CMake file in ci/super and then use that file for both.

tmatsuo: I tried to use the existing CMake file for the super build, but it tried to compile everything from the source, and it failed to compile protobuf.

So for now, maybe we can use the separate CMake file and consolidate them later?

coryan: SGTM.

ci/kokoro/windows/CMakeLists.txt
tmatsuo: Yeah I tried that, and getting into trouble. Maybe we can consolidate them later?

google/cloud/spanner/internal/date.cc
tmatsuo: Done

tmatsuo:
https://source.cloud.google.com/results/invocations/6ce70af6-30dc-465f-a3a3-633ef875bc87/log
The Windows Bazel build failed with:

```
E0910 18:01:31.262000000  5868 external/com_github_grpc_grpc/src/core/lib/security/credentials/alts/check_gcp_environment_windows.cc:72] CreateFile failed (3).
Creating database and table unknown file: error: SEH exception with code 0xc0000005 thrown in auxiliary test code (environments or event listeners).
```
I think I need to add `--jobs 1` option to the bazel integration test.

coryan:
> I think I need to add --jobs 1 option to the bazel integration test.

@tmatsuo I thought I had in #578, I am recording these failures in #573 until we figure out how to fix them.

PR #596, opened 2019-09-09
Use a custom deleter for Sessions.

mr-salty:
Rather than having a unique_ptr to a `SessionHolder` that has its own
"deleter", remove one level and just use a custom deleter for
`unique_ptr`.

The old `SessionHolder` becomes `Session` (and is now the object we put
in the pool, although it only has a name for now, I suspect when we
implement proper pool management it will need additional members).
`SessionHolder` is now the unique_ptr with a custom deleter.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/596)


google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -459,22 +452,23 @@ StatusOr<std::unique_ptr<SessionHolder>> ConnectionImpl::GetSession(
...
+  return SessionHolder(session.release(), [this](Session* session) {
+    this->ReleaseSession(session);
+  });
 }
 
-void ConnectionImpl::ReleaseSession(std::string session) {
+void ConnectionImpl::ReleaseSession(Session* session) {
   std::lock_guard<std::mutex> lk(mu_);
-  sessions_.push_back(std::move(session));
+  sessions_.emplace_back(session);
```
devjgm: Maybe still `push_back`? https://abseil.io/tips/112

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -459,22 +452,23 @@ StatusOr<std::unique_ptr<SessionHolder>> ConnectionImpl::GetSession(
...
   // TODO(#409) take a (weak) reference to `this` to avoid use-after-free.
-  auto deleter = dissociate_from_pool
-                     ? std::function<void(std::string)>()
-                     : [this](std::string session) {
-                         this->ReleaseSession(std::move(session));
-                       };
-  return ::google::cloud::internal::make_unique<SessionHolder>(
-      std::move(session_name), std::move(deleter));
+  return SessionHolder(session.release(), [this](Session* session) {
+    this->ReleaseSession(session);
```
devjgm: Does something need to `delete session` somewhere? 

 #IAmNotAUniquePtrDeleterExpert

google/cloud/spanner/internal/transaction_impl_test.cc
```diff
@@ -134,8 +132,8 @@ ResultSet Client::Read(std::unique_ptr<SessionHolder>& session,
     }
     switch (mode_) {
       case Mode::kReadSucceeds:  // `begin` -> `id`, calls now parallelized
-        session = ::google::cloud::internal::make_unique<SessionHolder>(
-            session_id_, /*deleter=*/nullptr);
+        session = ::google::cloud::internal::make_unique<internal::Session>(
```
coryan: This is doing a lot of stuff.

It is creating a `std::unique_ptr<internal::Session, std::default_delete<internal::Session>>` and automatically converting it to `SessionHolder` (which is *not* a type alias, there is an implicit conversion here).

As a side effect, `session` no longer returns the session to the pool, it now simply deallocates.  Is that what we want?

If it is, can we put a comment or something, because it is not super easy to follow.

google/cloud/spanner/internal/connection_impl.cc
coryan: Nah, the `ReleaseSession()` call transfers ownership of the pointer to the `std::vector<std::unique_ptr<Session>>`, which will actually do the delete (eventually).

google/cloud/spanner/internal/connection_impl.cc
coryan: I think `std::unique_ptr<T>::unique_ptr(T*)` is `explicit`, so `push_back()` does not work in this case?

devjgm: I missed that sessions_ was a `vector<unique_ptr<T>>`. So you and Todd are right here. My mistake.

google/cloud/spanner/internal/connection_impl.cc
devjgm: I see now. There are two different unique_ptr's holding `Session`s. One of them uses this custom deleter, the other doesn't. And we transfer ownership between them. Ack.

PR #604, opened 2019-09-10
test: Fix integration tests on Windows CMake builds

tmatsuo:
Also enable integration tests by default

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/604)


tmatsuo:
@coryan I also added some fixes for the build itself. PTAL

PR #607, opened 2019-09-11
Prevent `SessionHolder` from calling into a deleted `ConnectionImpl`.

mr-salty:


mr-salty:
I believe this PR is correct and could be committed (after removing the TODOs I added to illustrate the points below) but I'm marking it "draft" due to a couple issues I wanted to discuss, and thought it would be easier with actual code in front of people:

* It's now essentially impossible to allocate a `ConnectionImpl` on the stack or even on the heap with a normal pointer or `unique_ptr` (you'll get weak pointer exceptions if `this` doesn't actually have an associated `shared_ptr`). I'm a bit torn on whether this is a problem or not (we were only allocating it on the stack in tests, and `ConnectionImpl` is all `internal` regardless).
  * if we go this route maybe we should make the `ConnectionImpl` constructor private to force people to use the factory method (which would become a `friend`; this is a bit odd IMO since the factory is in `client.h`.).
  * Another possible option is to make a separate `SessionPool` class that is owned (via shared_ptr) by the `ConnectionImpl`, although that is a bit tricky because the `SessionPool` will also have to have a pointer to the `ConnectionImpl` so it can make RPCs to create and delete sessions... which is why bww and I have been thinking they should be the same object.
* The problem this PR fixes is a use-after-free bug. So, I wrote a new test to exercise this, but the issue is that with the old (existing) code (note the `#if 1` I left in for now) it only fails under msan. It would be nice if it failed in normal builds too, but I don't know how to accomplish that without mocking out a bunch of other things or adding hooks somewhere. Maybe it's ok as-is?

PR #634, opened 2019-09-13
docs:Upgrade python-protobuf for docuploader

tmatsuo:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/634)

### DIFF
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -341,6 +341,7 @@ fi
 
 if [[ "${BUILD_NAME}" == "publish-refdocs" ]]; then
   "${PROJECT_ROOT}/ci/kokoro/docker/publish-refdocs.sh"
+  exit_status=$?
 else
   "${PROJECT_ROOT}/ci/kokoro/docker/upload-docs.sh"
 fi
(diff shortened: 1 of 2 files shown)


ci/kokoro/docker/publish-refdocs.sh
```diff
@@ -65,6 +65,11 @@ echo "================================================================"
 echo "Installing docuploader package $(date)."
 python3 -m pip install --user gcp-docuploader
 
+
+echo "================================================================"
+echo "Upgrading protobuf package $(date)."
+python3 -m pip install --user protobuf --upgrade
```
coryan: Can we do both of these in a single pip command?

tmatsuo: Yes, done!

ci/kokoro/docker/build.sh
```diff
@@ -341,6 +341,7 @@ fi
 
 if [[ "${BUILD_NAME}" == "publish-refdocs" ]]; then
   "${PROJECT_ROOT}/ci/kokoro/docker/publish-refdocs.sh"
+  exit_status=$?
```
coryan: FYI: this seems like an unrelated change, but I think I understand why we need it

tmatsuo: Yeah, with this the job will correctly report failures

Issue #419, opened 2019-08-23
Create DatabaseAdminConnection to support mocking.


coryan:
For the same reasons we have SpannerConnection to support mocking the `spanner::Client` we need a `DatabaseAdminConnection` to support mocking the `spanner::DatabaseAdminClient`.

PR #638, opened 2019-09-13
feat: implement DatabaseAdminConnection

coryan:
Implement a new class so customers can mock the `DatabaseAdminClient`
behavior. This is the first of a series of PRs, I plan to refactor the
tests, which currently test both DatabaseAdminClient and
DatabaseAdminConnection together, and then move the retry loop to
`DatabaseAdminConnectionImpl`.

Part of the work for #419

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/638)


google/cloud/spanner/database_admin_connection.h
```diff
@@ -0,0 +1,174 @@
...
+  /// Wrap the arguments for `DropDatabase()`.
+  struct DropDatabaseParams {
+    /// The name of the database.
+    Database database_name;
+  };
+
+  /// Wrap the arguments for `ListDatabases()`.
+  struct ListDatabasesParams {
+    /// The full name of the instance in
+    /// `projects/<project-id>/instances/<instance-id>` format.
```
devjgm: Documenting the name format here may not be needed since the field is of type `Instance` and that class manages the format.

Related: FWIW, I might have named these fields just `database` and `instance` (i.e., without the `_name` suffix). But if you think the suffix helps, it's fine with me to keep it.

google/cloud/spanner/mocks/mock_database_admin_connection.h
```diff
@@ -0,0 +1,60 @@
...
+#define GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_MOCKS_MOCK_DATABASE_ADMIN_CONNECTION_H_
+
+#include "google/cloud/spanner/database_admin_connection.h"
+#include <gmock/gmock.h>
+
+namespace google {
+namespace cloud {
+namespace spanner_mocks {
+inline namespace SPANNER_CLIENT_NS {
+/**
```
devjgm: nit: I think most of our files have a single blank line after opening this inline namespace. I think clang-format used to remove this blank line, then I it stopped removing it, not sure what it's doing now :-)

coryan: Fixed.  I think clang-format preserves 1, but no more, blank lines here.

google/cloud/spanner/database_admin_connection.h
coryan: > Documenting the name format here may not be needed since the field is of type `Instance` and that class manages the format.

Done.

> Related: FWIW, I might have named these fields just `database` and `instance` (i.e., without the `_name` suffix). But if you think the suffix helps, it's fine with me to keep it.

Done.

Issue #502, opened 2019-08-29
Consider making results->Rows<Row<Ts...>>() work


devjgm:
Today `ResultSet::Rows()` is a variadic template requiring the user to specify the types of each column in a row. For example: `result_set.Rows<std::int64_t, std::string>()`. Two problems with this are that it can result in verbosity in the call to Rows (which will likely be in a range-for loop and wrapped), and a list of types cannot be aliased.

To address these issues we could make the following work:

```cc
using RowType = Row<std::int64_t, std::string>;
result_set.Row<RowType>();
```
Which addresses both concerns.

Some questions:
1. Do we want to add this new feature?
2. Do we want to support this new way *AND* the old way of separate types? Or only the new way?

PR #653, opened 2019-09-16
feat: Changed RowParser to require a Row<Ts...> template param

devjgm:
BREAKING CHANGE: `ResultSet::Rows` used to be a variadic template that
took the individual C++ types for each row. With this change that
function is now a template with one parameter, which must be a
`Row<...>` type.

Some reasons for this change:

* Our code is simpler when we minimize the functions that need to be
variadic. Instead of making a bunch of functions variadic, we should use
the `Row` class, which handles all the variadic stuff.

* Using a type named "Row" seems more sensible when we're talking about
a "row" than does a sequence of C++ types. That is, bundling C++ types
in a `Row<...>` when those types logically represent a "row" seems like
a readability win.

* A sequence of bare C++ types cannot be aliased, so they need to be
repeated wherever they're needed.

* Spelling all the C++ types (without an alias) in a call to `Rows()` in
a range-for loop causes a bunch of wrapping which hurts readability.

Fixes: #502

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/653)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -566,8 +566,7 @@ void ReadWriteTransaction(google::cloud::spanner::Client client) {
     if (!read) {
       throw std::runtime_error(read.status().message());
     }
-    using RowType = spanner::Row<std::int64_t>;
-    for (auto row : read->Rows<RowType>()) {
+    for (auto row : read->Rows<spanner::Row<std::int64_t>>()) {
       if (!row) {
         throw std::runtime_error(read.status().message());
       }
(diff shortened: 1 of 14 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -564,7 +566,8 @@ void ReadWriteTransaction(google::cloud::spanner::Client client) {
     if (!read) {
       throw std::runtime_error(read.status().message());
     }
-    for (auto row : read->Rows<std::int64_t>()) {
+    using RowType = spanner::Row<std::int64_t>;
```
coryan: nit: here (and a few places below) it seems like `read->Row<spanner::Row<std::int64_t>>()` would be a cleaner example.

devjgm: I wasn't sure which way looked the best. So I updated this (and the places below) to avoid the `RowType` alias if the alias was only used once, and the `Row` only contained one column.

PR #657, opened 2019-09-17
feat: implement retry loop for ConnectionImpl::Rollback

coryan:
Add retry policy and backoff policy member functions to
`internal::ConnectionImpl`, then change the Rollback() member function
to use the retry loop, and change the tests to verify it does retry
a single failure, it stops on the first permanent failure, and that it
stops retrying when there are too many transient failures.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/657)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -14,6 +14,7 @@
 
 #include "google/cloud/spanner/internal/connection_impl.h"
 #include "google/cloud/spanner/internal/partial_result_set_reader.h"
+#include "google/cloud/spanner/internal/retry_loop.h"
 #include "google/cloud/spanner/internal/time.h"
 #include "google/cloud/spanner/query_partition.h"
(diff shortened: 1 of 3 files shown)


mr-salty:
___
*[google/cloud/spanner/internal/connection_impl.cc, line 24 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/657#-Lp0-SwlBE4j4aGeYptK:-Lp4g8vX74SVdUmf-3dj:bpg3qfr) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/12c18d4f3f68ceaa12d940a65baafaf06407d025/google/cloud/spanner/internal/connection_impl.cc#L24)):*

I am going to leave this as-is, several of the includes already here required `<chrono>`.
</blockquote></details>

we should still `#include <chrono>` here according to https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

PR #659, opened 2019-09-17
feat: support install components for the library

coryan:
TIL: we can make it easier for package maintainers to create separate
packages for development vs. runtime if we use named components for
the library install directories. With CMake 3.5, however, we need to
use a secondary `install()` command for the libFoo.so.X.Y.Z link (aka
the NAMELINK_ONLY install)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/659)

### DIFF
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -393,6 +393,8 @@ install(TARGETS spanner_client
                 NAMELINK_SKIP
         ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
                 COMPONENT google_cloud_cpp_spanner_development)
+# With CMake-3.12 and higher we could avoid this separate command (and the
+# duplication).
 install(TARGETS spanner_client
         LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/CMakeLists.txt
```diff
@@ -387,8 +389,16 @@ install(EXPORT spanner-targets
 install(TARGETS spanner_client
         EXPORT spanner-targets
         RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+                COMPONENT google_cloud_cpp_spanner_runtime
         LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+                COMPONENT google_cloud_cpp_spanner_development
```
tmatsuo: Is this for the shared library itself? Then shouldn't it be `runtime`?

I guess the files in the `runtime` goes into the normal libxxx style packages and `development` goes to libxxx-dev.

coryan: Yes, fixed.

coryan:
PTAL.

PR #665, opened 2019-09-19
feat: implement retry loop for Commit()

### LINKED ISSUE #648: Add retry loop for internal::ConnectionImpl::Commit
We need to:

See the discussion on #627 this may require changes to the implementation too.

coryan:
# DO NOT REVIEW: TESTING

Fixes #648.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/665)


coryan:
We need to refactor the retry policies to not handle `kAborted` as a transient error first.

PR #667, opened 2019-09-19
feat: use separate policies for retry vs. rerun

### LINKED ISSUE #529: should RunTransaction only loop on StatusCode::kAborted?
Consider snippets like ...
```
c.ExecuteSql(txn, ...);
```
and ...
```
RunTransaction(c, {}, [](Client c, Transaction txn) {
  c.ExecuteSql(txn, ...);
  return ...;
});
```
If the first one retries transient failures, does/should the `RunTransaction()` loop also retry transient failures?  (Is `StatusCode::kAborted` "transient"?)

Strawman: `kAborted` is a special case, handled only by `RunTransaction()`.  Indeed, let's not even call the `RunTransaction()` loop "retries", but rather "reruns".

### LINKED ISSUE #666: Retry policy should not retry kAborted
The retry policies for most operations should not retry on `kAborted`, recovering from a transaction failure requires a completely different "re-run" loop (aka `RunTransaction` or `Connection::Commit(functor)`.

This also means we cannot use the regular retry policies for `RunTransaction`.

This is related to #529 too.

coryan:
We only want to retry requests on kUnavailable and kResourceExhausted,
and we want to re-run transactions only on kAborted.

Fixes #666 and fixes #529

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/667)

### DIFF
--- a/google/cloud/spanner/retry_policy.h
+++ b/google/cloud/spanner/retry_policy.h
@@ -28,8 +28,24 @@ namespace internal {
 /// Define the gRPC status code semantics for retrying requests.
 struct SafeGrpcRetry {
   static inline bool IsTransientFailure(google::cloud::StatusCode code) {
-    return code == StatusCode::kAborted || code == StatusCode::kUnavailable ||
-           code == StatusCode::kDeadlineExceeded;
+    return code == StatusCode::kUnavailable ||
+           code == StatusCode::kResourceExhausted;
+  }
+  static inline bool IsOk(google::cloud::Status const& status) {
+    return status.ok();
+  }
+  static inline bool IsTransientFailure(google::cloud::Status const& status) {
+    return IsTransientFailure(status.code());
+  }
+  static inline bool IsPermanentFailure(google::cloud::Status const& status) {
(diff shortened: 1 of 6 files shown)


google/cloud/spanner/retry_policy.h
```diff
@@ -28,8 +28,24 @@ namespace internal {
...
+  static inline bool IsTransientFailure(google::cloud::Status const& status) {
+    return IsTransientFailure(status.code());
+  }
+  static inline bool IsPermanentFailure(google::cloud::Status const& status) {
+    return !IsOk(status) && !IsTransientFailure(status);
+  }
+};
+
+/// Define the gRPC status code semantics for rerunning transactions.
+struct SafeTransactionReRun {
```
devbww: Here and throughout, treat "rerun" as a single word, just like "retry" (see Merriam-Webster for example).

coryan: Done.

coryan:
PTAL.

PR #680, opened 2019-09-20
feat: implement retry loop for Commit()

### LINKED ISSUE #648: Add retry loop for internal::ConnectionImpl::Commit

coryan:
Fixes #648

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/680)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl_test.cc
+++ b/google/cloud/spanner/internal/connection_impl_test.cc
@@ -683,6 +683,7 @@ TEST(ConnectionImplTest, CommitCommit_IdempotentTransientSuccess) {
         return response;
       }));
 
+  // Set the id because that makes the commit idempotent.
   auto txn = MakeReadWriteTransaction();
   internal::Visit(txn, [](SessionHolder&, spanner_proto::TransactionSelector& s,
                           std::int64_t) {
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -626,6 +652,78 @@ TEST(ConnectionImplTest, CommitCommitFailure) {
...
+                          spanner_proto::CommitRequest const& request) {
+        EXPECT_EQ("test-session-name", request.session());
+        EXPECT_EQ("test-txn-id", request.transaction_id());
+        spanner_proto::CommitResponse response;
+        *response.mutable_commit_timestamp() =
+            internal::ToProto(Timestamp{std::chrono::seconds(123)});
+        return response;
+      }));
+
+  auto txn = MakeReadWriteTransaction();
```
devbww: Should we add a comment, like below, about why we're binding an id?

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -626,6 +652,78 @@ TEST(ConnectionImplTest, CommitCommitFailure) {
...
+  internal::Visit(txn, [](SessionHolder&, spanner_proto::TransactionSelector& s,
+                          std::int64_t) {
+    s.set_id("test-txn-id");
+    return 0;
+  });
+
+  auto commit = conn->Commit({txn, {}});
+  EXPECT_STATUS_OK(commit);
```
devbww: EXPECT_EQ(commit->commit_timestamp, Timestamp{std::chrono::seconds(123)}); ?

google/cloud/spanner/internal/connection_impl_test.cc
coryan: Done.

google/cloud/spanner/internal/connection_impl_test.cc
coryan: Done.

PR #693, opened 2019-09-25
feat: implement PartialResultSetRead with resumes

### LINKED ISSUE #271: Implement resume support in Spanner Client
See the documentation for `resumeToken` for details: https://cloud.google.com/spanner/docs/reference/rest/v1/PartialResultSet

> Streaming calls might be interrupted for a variety of reasons, such as TCP connection loss. If this occurs, the stream of results can be resumed by re-sending the original request and including resumeToken. Note that executing any other transaction in the same session invalidates the token.

### LINKED ISSUE #674: Add retry/resume to internal::ConnectionImpl::Read
We need to:

coryan:
Use it in ConnectionImpl::Read to convince myself it is actually useful.

This fixes #271 and fixes #674

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/693)


devbww:
> I am pretty sure you cannot capture member variables

This is true, but you can get the effect without the unnecessary incr/decr of the refcount by making `stub` a reference to `stub_`, and then capturing that by value.

coryan:
That is a neat trick, done.

devbww:
The comment is now somewhat misleading/confusing I think.

coryan:
> The comment is now somewhat misleading/confusing I think.

I assume you mean the comment about `stub`.  I took another stab at it, if you still think it needs work please make a suggestion about what to say, or if you prefer we can remove it, I do not think it matters that much.

devbww:
Sorry for not being more specific.  LGTM.  Thanks.

I'll defer the approval to @scotthart.

Issue #521, opened 2019-09-01
Implement spanner::InstanceAdminClient::DeleteInstance


coryan:
At the end of this task we will have:

- [ ] The member function in `spanner::InstanceAdminClient` and related classes.
- [ ] Some unit tests to demonstrate that failures are handled correctly, including retries.
- [ ] An integration test to demonstrate the function works against production.
- [ ] An example to link from the Doxygen documentation.

PR #711, opened 2019-09-30
feat: implement CreateInstance and DeleteInstance on InstanceAdminClient

### LINKED ISSUE #519: Implement spanner::InstanceAdminClient::CreateInstance
At the end of this task we will have:

- [ ] The member function in `spanner::InstanceAdminClient` and related classes.
- [ ] Some unit tests to demonstrate that failures are handled correctly, including retries.
- [ ] An integration test to demonstrate the function works against production.
- [ ] An example to link from the Doxygen documentation.

tmatsuo:
With this PR, I introduced basic integration tests and samples for InstanceAdminClient, which create and delete a new instance with a random name, prefixed with `temporary-instance-{DATE}-`.

While these tests are only executed with an environment variable `RUN_SLOW_INTEGRATION_TESTS=yes`, if a random instance created by these tests is picked up by `google::cloud::spanner_testing::PickRandomInstance`, it's going to be problematic.

So, I made a change in `google::cloud::spanner_testing::PickRandomInstance`, now it only picks up the instance prefixed `test-instance-` for isolating instances between normal tests and CRUD tests for `InstanceAdminClient`.

I feel `InstanceAdminClient::CreateInstance` has too many parameters, but I'm somewhat ok with this because I don't expect many people will use this function anyways.

fixes #519 #521

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/711)

### DIFF
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -162,6 +162,7 @@ elif [[ "${BUILD_NAME}" = "coverage" ]]; then
   export DISTRO=fedora-install
   export DISTRO_VERSION=30
   : "${RUN_INTEGRATION_TESTS:=$DEFAULT_RUN_INTEGRATION_TESTS}"
+  export RUN_SLOW_INTEGRATION_TESTS=yes
   in_docker_script="ci/kokoro/docker/build-in-docker-cmake.sh"
 elif [[ "${BUILD_NAME}" = "bazel-dependency" ]]; then
   export DISTRO=ubuntu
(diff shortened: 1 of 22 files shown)


tmatsuo:
@coryan Thanks for the review! I think I addressed your comments.

* I started to run slow integration tests for coverage builds.
* For instance admin integration tests, use instance name prefixed with "temporary-instance-{DATE}-" for isolation.

There is a potential leak for those instances upon failures, for now we can manually delete instances with old date. Maybe we can add an automatic cleanup code somewhere.

PTAL

tmatsuo:
Oops old gcc doesn't have that function.

tmatsuo:
For the length of the instance id, I found:
https://cloud.google.com/spanner/quotas

stateing

>>> Instance ID length | 2 to 64 characters

coryan:
> For the length of the instance id, I found:
> https://cloud.google.com/spanner/quotas
> 
> stating
> 
> > > > Instance ID length | 2 to 64 characters

Which I guess means either the quota documentation, or the proto comments (or both) have a bug?  I think we should file something.

PR #823, opened 2019-10-01
Instance admin update instance

### LINKED ISSUE #520: Implement spanner::InstanceAdminClient::UpdateInstance
At the end of this task we will have:

- [ ] The member function in `spanner::InstanceAdminClient` and related classes.
- [ ] Some unit tests to demonstrate that failures are handled correctly, including retries.
- [ ] An integration test to demonstrate the function works against production.
- [ ] An example to link from the Doxygen documentation.

tmatsuo:
fixes #520 

I surfaced `gcsa::UpdateInstanceRequest`, and wrote samples and integration tests accordingly. The sample code doesn't look so bad as I worried, so I created this PR without implementing the request builder.

Another question.
Well, then, when I look at `CreateInstance` and `UpdateInstance`, I feel they are inconsistent. I'm thinking to also surface `CreateInstanceRequest` on `CreateInstance` to users. WDYT?

### DIFF
--- a/google/cloud/spanner/instance_admin_client.cc
+++ b/google/cloud/spanner/instance_admin_client.cc
@@ -33,6 +33,13 @@ InstanceAdminClient::CreateInstance(
                                 instance_config, node_count, labels});
 }
 
+future<StatusOr<google::spanner::admin::instance::v1::Instance>>
+InstanceAdminClient::UpdateInstance(
+    google::spanner::admin::instance::v1::UpdateInstanceRequest const&
+        request) {
+  return conn_->UpdateInstance({request});
+}
+
 Status InstanceAdminClient::DeleteInstance(Instance const& in) {
   return conn_->DeleteInstance({in.FullName()});
 }
--- a/google/cloud/spanner/instance_admin_client.h
+++ b/google/cloud/spanner/instance_admin_client.h
@@ -122,6 +122,9 @@ class InstanceAdminClient {
   /**
    * Updates a Cloud Spanner instance.
    *
+   * Note that when updating a field, you have to set the field mask on the
+   * `UpdateInstanceRequest` with the field name.
+   *
    * @par Example
    * @snippet samples.cc update-instance
    *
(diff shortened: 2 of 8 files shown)


coryan:
___
*[google/cloud/spanner/instance_admin_connection.cc, line 157 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/823#-Lq6HvgZ-UhALefDP3ak:-Lq7XfgsAoRrrj6tA-Nu:b-xk351e) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/394897b49fb8dd7757fb35105b9899351ca2f406/google/cloud/spanner/instance_admin_connection.cc#L157)):*
<details><summary><i>Previously, tmatsuo (Takashi Matsuo) wrote…</i></summary><blockquote>

I found an issue with the current code in general. The current sample assigns all the instance fields in the UpdateInstanceRequest. Which might be little bit problematic when two things are trying to modify different fields (the second change will override the field changed in other thread). In such case, the request is not idempotent, and also it's questionable to retry unconditionally.

If we provide a Request builder, I think we can say the request is "retryable", while the request might not be strictly idempotent.

WDYT?
</blockquote></details>

If you have two tasks modifying the same field then all bets are off, the outcome is "one of the possible orderings."

But two (or more) tasks modifying different fields are Okay because `field_mask` will instruct the server to ignore any other fields that you send, so the final result, no matter how many times the two updates are executed.  I think these are idempotent after all.

coryan:
___
*[google/cloud/spanner/instance_admin_connection.cc, line 157 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/823#-Lq6HvgZ-UhALefDP3ak:-Lq87ezWDGLFffc7khQF:bk7olvp) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/394897b49fb8dd7757fb35105b9899351ca2f406/google/cloud/spanner/instance_admin_connection.cc#L157)):*
> How about to enable retry (as the current code does) and add some notes in the function comment?

I agree that we should enable retry, but I disagree that it is enabled, RetryLoop() with a false `is_idempotent` parameter fails at the first error I think .

tmatsuo:
@coryan I added a builder class, and enabled retry. The sample code looks even nicer, thanks!
PTAL

Issue #307, opened 2019-08-05
Implement better session pooling.


coryan:
To unblock the implementation of `Client::Commit()` I implemented a very naive session pool. Here are some of the things it does not do:

- It does not remove sessions that have likely expired (i.e. unused for more than 1 hour).
- It does not automatically refresh the sessions in the background (consider using a `CompletionQueue` to do this with application-provided threads).
- It does not support a Stub that has multiple channels under the hood.
- It does not warm the session pool during the class initialization (consider doing this with background threads).

mr-salty:
Assigning to myself to do some research / design work on this; even if we don't implement it (fully or at all) for alpha, we've run into several case where other decisions are predicated on having some understanding of how we eventually expect it to work.

devjgm:
Any update on this? Looks like there's been lots of progress. Is this done?

mr-salty:
not done, I have one more PR in progress and then I think I can declare it "done enough to close this bug" and open issues for any small things leftover (in retrospect I probably should have opened multiple issues at the start instead of using this for everything)

devjgm:
Can we file a separate issue that clearly describes what remaining feature/functionality we need to add in this next PR you're referring to? I'm thinking then we can probably label *that* issue with GA (not beta), then close this one for Beta.

FTR, this is related to what Todd and I talked about offline.

mr-salty:
The 4 bullets I listed initially probably should have been separate issues.

3 of them are done: utilizing multiple channels, proactively creating `min_sessions`, and removing expired `Sessions` from the pool (which had already been split out to #914).

I opened #1171 for `Session` refresh and #1172 as an additional issue that wasn't identified here (properly maintaining `min_sessions_ and `max_idle_sessions`) - but as we discussed we are comfortable going beta without those implemented, so I'm closing this issue.

PR #916, opened 2019-10-02
Factor SessionPool out of ConnectionImpl.

mr-salty:
Make `SessionPool` its own class to avoid clutter in `ConnectionImpl`
and to make it easier to test.

Since `SessionPool` is a member of `ConnectionImpl`, which guarantees
the former cannot outlive the latter; only `ConnectionImpl` is required
to be reference counted. This simplifies lifetime management, avoiding
some subtle issues and race conditions, at the cost of requiring some
ostensibly pool-related private methods in `ConnectionImpl` for the
benefit of `SessionHolder`

There are no functional changes compared to the current implementation.

Part of #307

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/916)


mr-salty:
___
*[google/cloud/spanner/internal/connection_impl.h, line 61 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/916#-Lq9lsARBztlBV4T17k-:-Lq9lsARBztlBV4T17k0:bvei9lp) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/cd0c3bccb29e88b52b57656bd68851e22318260b/google/cloud/spanner/internal/connection_impl.h#L61)):*
> ```objc
>         DefaultConnectionBackoffPolicy());
> 
> // Interface used by the SessionPool to manage sessions
> ```

@devjgm @devbww I don't really like this especially given it leads to multiple inheritance, perhaps we can discuss alternatives tomorrow?

It will eventually get other methods like DeleteSession and GetSession/RefreshSession or similar

google/cloud/spanner/internal/connection_impl.h
```diff
@@ -120,19 +128,35 @@ class ConnectionImpl : public Connection,
...
-  // The current session pool.
-  // TODO(#307) - improve session refresh and expiration.
-  std::mutex mu_;
-  std::vector<std::unique_ptr<Session>> sessions_;  // GUARDED_BY(mu_)
+  std::shared_ptr<SpannerStub> stub_;
 
   std::unique_ptr<RetryPolicy> retry_policy_;
   std::unique_ptr<BackoffPolicy> backoff_policy_;
+
+  friend class SessionPool;
```
devbww: Why do we need this friendship?

google/cloud/spanner/internal/connection_impl.h
```diff
@@ -56,12 +58,22 @@ std::shared_ptr<ConnectionImpl> MakeConnection(
     std::unique_ptr<BackoffPolicy> backoff_policy =
         DefaultConnectionBackoffPolicy());
 
+// Interface used by the SessionPool to manage sessions
+class SessionManager {
```
devbww: Should we have a separate header for SessionManager?

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -0,0 +1,54 @@
...
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+
+StatusOr<std::unique_ptr<Session>> SessionPool::Allocate() {
+  {
+    std::lock_guard<std::mutex> lk(mu_);
+    if (!sessions_.empty()) {
+      auto session = std::move(sessions_.back());
+      sessions_.pop_back();
+      return {std::move(session)};
```
devbww: Are the `{}`s necessary?  Indeed, is the `move` necessary?

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  auto sessions = manager_->CreateSessions(/*num_sessions=*/1);
+  if (!sessions.ok()) {
+    return std::move(sessions).status();
+  }
+  // TODO(#307) for now, assume CreateSessions() returns exactly one Session on
```
devbww: Nit: Didn't we ask for exactly one?

google/cloud/spanner/internal/session_pool.h
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Allocate a session from the pool, creating a new `Session` if necessary.
+   * @returns an error if session creation fails; always returns a non-null
+   * pointer on success.
+   */
+  StatusOr<std::unique_ptr<Session>> Allocate();
+
+  /// Release `session` back to the pool.
+  void Release(Session* session);
```
devbww: Why the asymmetry on `unique_ptr` -v- raw pointer between `Allocate()` and `Release()`.

google/cloud/spanner/internal/session_pool.h
```diff
@@ -0,0 +1,67 @@
...
+
+ private:
+  std::mutex mu_;
+  std::vector<std::unique_ptr<Session>> sessions_;  // GUARDED_BY(mu_)
```
devbww: Might this be a vector of dissociated `SessionHolder`s instead?  Having two different varieties of `unique_ptr<Session>` might be confusing.

google/cloud/spanner/internal/session_pool_test.cc
```diff
@@ -0,0 +1,126 @@
...
+  EXPECT_EQ((*session)->session_name(), "session1");
+
+  auto session2 = pool.Allocate();
+  ASSERT_STATUS_OK(session2);
+  EXPECT_EQ((*session2)->session_name(), "session2");
+
+  pool.Release((*session).release());
+  pool.Release((*session2).release());
+
+  // We should get the sessions back in the reverse order they were released.
```
devbww: Perhaps expound on why this is (currently?) important.

devbww: > Done.

OK, although I'd still say it is unclear why this is a necessary/desirable property.

google/cloud/spanner/internal/session_pool.cc
devbww: It looks like Clang-3.8 and GCC-4.8 didn't like something there.  Sorry.

PR #920, opened 2019-10-02
BREAKING CHANGE: Change Value::Bytes to g::c::spanner::Bytes

### LINKED ISSUE #413: Change `Value::Bytes` to be `Bytes<T>`
`Value::Bytes`([link](https://github.com/googleapis/google-cloud-cpp-spanner/blob/7586d46bfcd5f5b1ccf3d6f0edc7cbcddebc3ed7/google/cloud/spanner/value.h#L194-L222)) is currently implemented to hold a single data member of type `std::string`. Instead of us choosing `std::string` as The One Preferred data type for string a sequence bytes, we probably want a more general type that can efficiently accept and return bytes stored however the user has them.

Therefore, we probably want a type that looks like:

```cc
template <typename T>
class Bytes {
 public:
  template <typename Container>
  explicit Bytes(Container const& c) {
    // base64-encodes the data in `c` into `data_`
  }

  template <typename Container>
  T get() const {
    // base64-decode's data_ into a container of type `T`
    // Can this decoding fail? Should this return StatusOr<T>?
  }

 private:
  std::string data_;
};
```

A few other thoughts:

* We probably do not want to define `operator<()`, but we *do* want equality
* This class is complicated enough that it should probably live in its own `bytes.h` file
* We can define whatever other templated constructors that we want. For example, we *migth* want a constructor that takes a start and end iterator pair. This could be done in a future PR.

devbww:
Introduce `google::cloud::spanner::Bytes` as the representation for
the Spanner BYTES type ... a sequence of raw bytes (in contrast to the
Spanner STRING type, which is interpreted as Unicode characters).

Like the previous `Value::Bytes`, a `google::cloud::spanner::Bytes`
value can be converted to/from a base64-encoded US-ASCII `std::string`.

But unlike `Value::Bytes`, `google::cloud::spanner::Bytes` can be
converted to/from any ordered sequence of bytes, not just `std::string`.
That is, it can accept/produce data in any compatible representation
(for example, as a `std::vector<std::uint8_t>`).

Fixes #413.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/920)


google/cloud/spanner/bytes.h
```diff
@@ -0,0 +1,157 @@
...
+#include <iterator>
+#include <string>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+/**
+ * A representation of the Spanner BYTES type ... variable-length binary data.
```
devjgm: I'm not sure about the `...`. Maybe `, a` ??

google/cloud/spanner/bytes.h
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * A `Bytes` value can be constructed from, and converted to (1) a sequence of
+ * octets, or (2) a base64-encoded US-ASCII `std::string`. The base64 alphabet
```
devjgm: I don't think we want to expose anything about base64 in this class's interface. I think the base64 encoding should be completely an implementation detail hidden within this class.

If we need the ability to inject/extract the raw base64 string, we can expose internal friend functions that we can use, but users cannot.

Thoughts?

google/cloud/spanner/bytes.h
```diff
@@ -0,0 +1,157 @@
...
+  Bytes(InputIt first, InputIt last) {
+    Encoder encoder(base64_rep_);
+    while (first != last) {
+      encoder.buf_[encoder.len_++] = *first++;
+      if (encoder.len_ == encoder.buf_.size()) encoder.Flush();
+    }
+    if (encoder.len_ != 0) encoder.FlushAndPad();
+  }
+  template <typename Container>
+  explicit Bytes(Container const& c) : Bytes(c.begin(), c.end()) {}
```
devjgm: I think we may want to use `std::begin(c)` and `std::end(c)` here, which will make this code more generic. Indeed, I think this change may then naturally handle the case of a string literal, because `std::begin` has the needed overload for arrays (https://en.cppreference.com/w/cpp/iterator/begin). So you can probably remove the `Bytes(char const*)` constructor??

google/cloud/spanner/bytes.h
```diff
@@ -0,0 +1,157 @@
...
+  ///@}
+
+  /// Conversion to a sequence of octets.
```
devjgm: May want to document that `Container` needs to be a container with a constructor that accepts an input iterator pair??

google/cloud/spanner/bytes.h
devbww: "a variable-length binary data" doesn't scan right.  How about a colon?

[Aside: the description comes from [here](https://cloud.google.com/spanner/docs/data-types).]

google/cloud/spanner/bytes.h
devbww: Done.  Thanks.

google/cloud/spanner/bytes.h
devbww: Done.

google/cloud/spanner/bytes.h
devbww: The idea is that the user, as well as having their binary data in a variety of containers, might also have/want it as a base64-encoded string.  So, rather converting back-and-forth again, we can supply it directly.

coryan:
This is good to go as far as I can tell.

devjgm:
> This is good to go as far as I can tell.

@devbww and I talked offline and we're going to remove the base64 stuff from the public API of this class before submitting.

google/cloud/spanner/bytes.h
devbww: PTAL

devjgm: Now that the base64 functions are gone from the user-facing API, we should probably remove all reference to base64 in the documentation.

google/cloud/spanner/bytes.h
```diff
@@ -0,0 +1,157 @@
...
+std::string BytesToBase64(Bytes const& b);
+}  // namespace internal
+
+/**
+ * A representation of the Spanner BYTES type: variable-length binary data.
+ *
+ * A `Bytes` value can be constructed from, and converted to any sequence of
+ * octets. All functions/operators should interpret the data as raw bytes (in
+ * contrast to the Spanner STRING type, where the data should be interpreted
+ * as Unicode characters).
```
devjgm: NO CHANGE NEEDED (feel free to commit as-is):

FWIW, I'm not sure if this last sentence is needed. It's talking about how the data should be interpreted. But `spanner::Bytes` is pretty much an opaque container, so callers can only even see the bytes either before they add them to `Bytes` or after they extract them from `Bytes`, at which point they can interpret the bytes however they want (they're their bytes).

Related: I'm not sure if mentioning Spanner STRING here clarifies or confuses anything. It's not clear why we are drawing any comparison between bytes and strings at all.

But just a question. Leaving as-is is fine with me. Whatever you think SGTM.

PR #921, opened 2019-10-03
feat: implement overloads on UpdateInstanceRequestBuilder

tmatsuo:
Add && overloads to all the fluent functions so that if you start with an R value, you'll just use the && version of `Build()`.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/921)


tmatsuo:
@coryan Thanks for the review! I inlined all the function and now it's header only. PTAL

Issue #629, opened 2019-09-13
Consider: Replacing ExecuteSql() with ExecuteQuery() and ExecuteDml()


devjgm:
Today we expose `Client::ExecuteSql`[1],  which returns a `StatusOr<ResultSet>`. This makes a lot of sense when a query is issued and the caller wants to consume the resulting rows, but it makes less sense when a DML statement is issued and therefore no rows will be returned. Another difference between a "query" and a dml statement is that the dml statement will return stats about how many rows were modified.

I think we should consider removing `Client::ExecuteSql` and replacing it with two separate functions that are named for "queries" (i.e., `SELECT ...`) and DML (e.g., `INSERT ...`). The functions would exist solely so they can have a different return type that is optimized for each use case.

For *example* (this is just an example, not a formal design spec):

```cc
class Client {
 public:
  struct QueryResult {
    template <typename... Ts>
    RowParser<Ts...> Rows();
  };
  QueryResult ExecuteQuery(SqlStatement);

  struct DmlResult {
    std::int64_t row_count();
  };
  StatusOr<DmlResult> ExecutDml(SqlStatement);
};
```

Notice how `ExecuteQuery` returns `QueryResult` that is *not* wrapped in a `StatusOr`? This is because the `StatusOr` is already exposed via the `RowParser`. However, `ExecuteDml` *does* return its results wrapped in a `StatusOr`, because it needs to.

NOTE: These functions would **not** parse the given `SqlStatement` and attempt to validate that, say, `ExecuteQuery` is only given a `SELECT` statement. It would be up to the caller to use these functions correctly.

Related to #628 
Related to #217 

[1]: https://github.com/googleapis/google-cloud-cpp-spanner/blob/970ce7a559fe1fdbc7fd6435835f0dd6b889aaac/google/cloud/spanner/client.h#L243

scotthart:
Looking at the C#, go, Java, and Python libraries, they all have different return types for DML statements vs. Query statements, so I think we would be following an established paradigm.

I agree that we should replace ExecuteSql with ExecuteQuery and ExecuteDml. If we later need to support some kind of hybrid, we could reintroduce ExecuteSql at that time. Furthermore, I think we need separate return types for each of Read, ExecuteQuery, and ExecuteDml.

PR #927, opened 2019-10-04
feat!: Replace ExecuteSql with ExecuteQuery and ExecuteDml

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/927)


scotthart:
fixes #629

google/cloud/spanner/internal/partial_result_set_source.h
```diff
@@ -51,6 +51,32 @@ class PartialResultSetSource : public internal::ResultSetSource {
     }
     return {};
   }
+
+  std::int64_t RowsModified() const override {
```
devjgm: (Just a comment FTR, not blocking this PR at all.)

Scott and I chatted offline about this. In a follow up PR, the `ExecuteDml` call should be changed to call the non-streaming `ExecuteSql` gRPC method. This will simplify the implementation of our `ExecuteDmlImpl`, and it will eliminate our need to use `PartialResultSetSource` in the DML cases. 

No change in this PR is needed now that I understand the direction we're headed. But maybe you want to drop a comment here that this function will be removed?

google/cloud/spanner/result_set.h
```diff
@@ -33,10 +34,182 @@ class ResultSetSource {
...
+  virtual optional<QueryPlan> QueryExecutionPlan() const = 0;
+
   virtual optional<google::spanner::v1::ResultSetStats> Stats() = 0;
 };
 }  // namespace internal
 
+/**
+ * Represents the result of a read operation using `spanner::Client::Read()`.
+ *
+ * Note that a `ReadResult` returns both the data for the operation, as a
```
devjgm: "both" doesn't sound right in this sentence.

google/cloud/spanner/result_set.h
```diff
@@ -33,10 +34,182 @@ class ResultSetSource {
...
+   */
+  optional<std::unordered_map<std::string, std::string>> QueryStats() const;
+
+  /**
+   * Returns the plan of execution for the query.
+   *
+   * @note Only available when the query is profiled or when the plan is
+   * explicitly requested.
+   */
+  optional<QueryPlan> QueryExecutionPlan() const;
```
devjgm: Since we have no way to request the query plan at this point, I'd recommend removing this. We can add it back if/when we add support for requesting the query plan. WDYT?

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -294,13 +303,43 @@ StatusOr<std::vector<ReadPartition>> ConnectionImpl::PartitionReadImpl(
   return read_partitions;
 }
 
-StatusOr<ResultSet> ConnectionImpl::ExecuteSqlImpl(
+class StatusOnlyResultSetSource : public internal::ResultSetSource {
+ public:
+  /// Factory method to create a PartialResultSetSource.
```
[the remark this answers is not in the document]
scotthart: Done.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -294,13 +303,43 @@ StatusOr<std::vector<ReadPartition>> ConnectionImpl::PartitionReadImpl(
...
+  static StatusOr<std::unique_ptr<StatusOnlyResultSetSource>> Create(
```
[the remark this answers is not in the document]
scotthart: Done.

google/cloud/spanner/internal/partial_result_set_source.h
scotthart: created issue #929

google/cloud/spanner/result_set.h
scotthart: Done.

google/cloud/spanner/result_set.h
scotthart: created issue #930, will document header indicating these functions are not yet implemented, pending resolve the issue.

devbww:
For what it's worth ...
```
$ grep -w ExecuteSql google/cloud/spanner/client.h 
   * key/value style alternative to `ExecuteSql()`.
   * to `ExecuteSql` to specify a subset of the query result to read.
   * `ExecuteSql`.
   * that includes  one or more `Read` or `ExecuteSql` requests and ultimately
 * issue any `Read()`s, `ExecuteSql()`s, etc., and return the `Mutation`s to
 * issue any `Read()`s, `ExecuteSql()`s, etc., and return the `Mutation`s to
```

PR #928, opened 2019-10-04
feat: initial version of throughput benchmark

coryan:
This benchmark just measures how many reads and write operations per
second we can execute using the client library, with both a single
`spanner::Client` and using one per thread.

This is the first version, it only measures the throughput for insert
mutations (but not DML statements), and for `SELECT` queries (but not
`Read()` calls). Those other measurements will be added later.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/928)


coryan:
Rebased to fix conflicts.

PR #942, opened 2019-10-09
feat!: finalize result types

scotthart:
- Renamed result_set.* files to results.*
- Unified return type for Client::ExecuteQuery and Client::Read into QueryResult
- Shortened ExecuteDmlResult to DmlResult
- Added ProfileQueryResult and ProfileDmlResult
- Implemented ExecutionPlan and ExecutionStats methods on appropriate result types
BREAKING CHANGE:

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/942)


google/cloud/spanner/results.cc
```diff
@@ -0,0 +1,106 @@
...
+    std::unique_ptr<internal::ResultSetSource> const& source) {
+  return source->Stats()->row_count_exact();
+}
+
+optional<std::unordered_map<std::string, std::string>> GetExecutionStats(
+    std::unique_ptr<internal::ResultSetSource> const& source) {
+  auto stats = source->Stats();
+  if (stats && stats->has_query_stats()) {
+    std::unordered_map<std::string, std::string> execution_stats;
+    auto map = stats->query_stats().fields();
```
devjgm: `auto const&` maybe? Or maybe just `for (auto const& entry : stats->query_stats().fields())` ?

google/cloud/spanner/results.h
```diff
@@ -12,20 +12,22 @@
...
 
 namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
-using QueryPlan = google::spanner::v1::QueryPlan;
+using ExecutionPlan = google::spanner::v1::QueryPlan;
 
 namespace internal {
 class ResultSetSource {
```
devjgm: Somewhere someone wrote that renaming this to `ResultSource` may be good. I agree that that sounds like a good rename. But it's fine to do in a separate PR.

google/cloud/spanner/results.h
```diff
@@ -12,20 +12,22 @@
...
+#include <memory>
+#include <string>
+#include <unordered_map>
 
 namespace google {
 namespace cloud {
 namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
-using QueryPlan = google::spanner::v1::QueryPlan;
+using ExecutionPlan = google::spanner::v1::QueryPlan;
```
devjgm: Since this is now a public symbol in our API, I think we should drop a little user-facing comment here about what an `ExecutionPlan` is. I don't think we need to duplicate all of QueryPlan's docs. but perhaps a brief sentence, and a link to QueryPlan's docs?

scotthart: Done.

google/cloud/spanner/results.h
scotthart: Done.

google/cloud/spanner/results.cc
scotthart: Done.

PR #947, opened 2019-10-10
fix: fix branch detection in the document publishing script

tmatsuo:
Theoretically it will work, but there's no easy way to test it before the actual release process.

Maybe we can backport this change and try this with the v0.2.x branch?

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/947)


coryan:
> Maybe we can backport this change and try this with the v0.2.x branch?

You can do that.  Applying a patch to the branch should not be hard.  Try not to create too many tags, or put 'test' in their name (e.g. `v0.2.1-test`) so we can remove them.

PR #955, opened 2019-10-14
feat: configurable strategy for background threads

### LINKED ISSUE #606: Consider using a CompletionQueue for background operations.
We have a few background operations to run:

- Polling for long running operations in `CreateInstance`, `CreateDatabase`, and the `Update*` versions.
- Keeping the sessions alive.

Maybe we should consider just having one thread (pool) and scheduling asynchronous requests using a `google::grpc_utils::CompletionQueue`.

coryan:
This change introduces an interface to represent the background
thread(s) used by a connection. I included two implementations in this
PR, one that automatically creates a single background thread, and
another that uses one or more threads provided by the application.

The ConnectionOptions class provides a factory for `BackgroundThreads`,
for those (we expect few) applications that will want to control the
background threads.

This fixes #606

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/955)

### DIFF
--- a/google/cloud/spanner/background_threads.h
+++ b/google/cloud/spanner/background_threads.h
@@ -1,0 +1,45 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_BACKGROUND_THREADS_H_
+#define GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_BACKGROUND_THREADS_H_
+
+#include "google/cloud/spanner/version.h"
+#include "google/cloud/grpc_utils/completion_queue.h"
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+
+/**
+ * A object representing the background threads available to a Client.
+ */
+class BackgroundThreads {
+ public:
+  virtual ~BackgroundThreads() = default;
+
+  /// The completion queue used for the background operations.
+  virtual grpc_utils::CompletionQueue cq() const = 0;
+
+  /// Terminate any automatically created threads.
+  virtual void Shutdown() = 0;
+};
+
+}  // namespace SPANNER_CLIENT_NS
+}  // namespace spanner
+}  // namespace cloud
+}  // namespace google
+
+#endif  // GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_BACKGROUND_THREADS_H_
--- a/google/cloud/spanner/background_threads.h
+++ b/google/cloud/spanner/background_threads.h
@@ -32,9 +32,6 @@ class BackgroundThreads {
 
   /// The completion queue used for the background operations.
   virtual grpc_utils::CompletionQueue cq() const = 0;
-
-  /// Terminate any automatically created threads.
-  virtual void Shutdown() = 0;
 };
 
 }  // namespace SPANNER_CLIENT_NS
--- a/google/cloud/spanner/connection_options.h
+++ b/google/cloud/spanner/connection_options.h
@@ -161,7 +161,9 @@ class ConnectionOptions {
    * Connections need to perform background work on behalf of the application.
    * Normally they just create a background thread and a `CompletionQueue` for
    * this work, but the application may need more fine-grained control of their
-   * threads. In this case the application can provide the
+   * threads. In this case the application can provide the `CompletionQueue` and
+   * it assumes responsibility for creating one or more threads blocked on
+   * `CompletionQueue::Run()`.
    */
   ConnectionOptions& DisableBackgroundThreads(
       google::cloud::grpc_utils::CompletionQueue const& cq);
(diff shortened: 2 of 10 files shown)


google/cloud/spanner/connection_options.h
```diff
@@ -152,13 +155,31 @@ class ConnectionOptions {
...
+   * Normally they just create a background thread and a `CompletionQueue` for
+   * this work, but the application may need more fine-grained control of their
+   * threads. In this case the application can provide the
+   */
+  ConnectionOptions& DisableBackgroundThreads(
+      google::cloud::grpc_utils::CompletionQueue const& cq);
+
+  using BackgroundThreadsFactory =
+      std::function<std::unique_ptr<BackgroundThreads>()>;
+  BackgroundThreadsFactory background_threads_factory() {
```
devjgm: I'm not sure I understand how to use this method. It looks like every call to this getter will return a factory that, when invoked, will create a new "BackgroundThread", and by default, will use a new CompletionQueue. Is that right?

Would it be good to just expose access to a single CompletionQueue instance that everyone can use? But honestly, maybe I'm just not understanding how this works.

google/cloud/spanner/connection_options.h
```diff
@@ -152,13 +155,31 @@ class ConnectionOptions {
    */
   grpc::ChannelArguments CreateChannelArguments() const;
 
+  /**
+   * Configure the connection to use @p cq for all background work.
+   *
+   * Connections need to perform background work on behalf of the application.
+   * Normally they just create a background thread and a `CompletionQueue` for
+   * this work, but the application may need more fine-grained control of their
+   * threads. In this case the application can provide the
```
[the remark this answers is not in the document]
coryan: Fixed.

google/cloud/spanner/background_threads.h
```diff
@@ -0,0 +1,45 @@
...
+ */
+class BackgroundThreads {
+ public:
+  virtual ~BackgroundThreads() = default;
+
+  /// The completion queue used for the background operations.
+  virtual grpc_utils::CompletionQueue cq() const = 0;
+
+  /// Terminate any automatically created threads.
+  virtual void Shutdown() = 0;
```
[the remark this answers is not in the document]
coryan: I am not sure. I *think* we will want to shut it down explicitly. But let's be conservative, I have removed the member function, we can always add it later if we really need it.

google/cloud/spanner/connection_options.h
coryan: > I'm not sure I understand how to use this method. It looks like every call to this getter will return a factory that, when invoked, will create a new "BackgroundThread", and by default, will use a new CompletionQueue. Is that right?

Yes.

> Would it be good to just expose access to a single CompletionQueue instance that everyone can use? But honestly, maybe I'm just not understanding how this works.

I think we decided to tie the lifetime of the automatically created threads to the lifetime of the `spanner::*Connection` object. That means the `*Connection` needs to know if it creates a thread and completion queue, and if it does create them it needs to shutdown the completion queue and join the thread in its destructor.  All those ifs suggested that this was a strategy given to the `*Connection` object.

In other words, we need to return something more than just the `CompletionQueue`, we need either (a) a thing that creates (when needed) the background thread for you, or (b) a `bool` that tells you if you should create a thread.

I went with option (a) because it seems wasteful to have every `spanner::*Connection` implement the "If I was asked to create a thread do this, otherwise do that".

google/cloud/spanner/connection_options_test.cc
```diff
@@ -146,6 +146,47 @@ TEST(ConnectionOptionsTest, CreateChannelArguments_WithChannelPool) {
               StartsWith(options.user_agent_prefix()));
 }
 
+TEST(ConnectionOptionsTest, DefaultBackgroundThreads) {
+  auto options = ConnectionOptions(grpc::InsecureChannelCredentials());
+  auto background = options.background_threads_factory()();
+
+  using ms = std::chrono::milliseconds;
+
+  // Verify the background thread is not the main thread.
```
devbww: s/the main/this/?

google/cloud/spanner/internal/background_threads_impl.cc
```diff
@@ -0,0 +1,40 @@
...
+    : runner_([](grpc_utils::CompletionQueue cq) { cq.Run(); }, cq_) {}
+
+AutomaticallyCreatedBackgroundThreads::
+    ~AutomaticallyCreatedBackgroundThreads() {
+  Shutdown();
+}
+
+void AutomaticallyCreatedBackgroundThreads::Shutdown() {
+  cq_.Shutdown();
+  if (runner_.joinable()) runner_.join();
```
devbww: How might `runner_` have become detached if we don't provide any access to it?

google/cloud/spanner/internal/background_threads_test.cc
```diff
@@ -0,0 +1,101 @@
...
+TEST(AutomaticallyCreatedBackgroundThreads, IsActive) {
+  AutomaticallyCreatedBackgroundThreads actual;
+
+  using ms = std::chrono::milliseconds;
+
+  auto expired = actual.cq().MakeRelativeTimer(ms(0));
+  EXPECT_EQ(std::future_status::ready, expired.wait_for(ms(100)));
+}
+
+/// @test Verify the background thread is not the main thread.
```
devbww: s/the main/this/

google/cloud/spanner/internal/background_threads_impl.cc
coryan: You can call `Shutdown()` twice, or `Shutdown()` and then the destructor.  In both cases the second call would throw:

https://en.cppreference.com/w/cpp/thread/thread/join

coryan: Calling `join()` (via `Shutdown()` for example) leaves `runner_` in a non-joinable state, in which case `std::thread::join` throws.

google/cloud/spanner/internal/background_threads_test.cc
coryan: Fixed.

google/cloud/spanner/connection_options_test.cc
coryan: Fixed.

coryan:
PTAL

google/cloud/spanner/internal/background_threads_impl.cc
devbww: OK.  (I wonder whether this class should do something to ensure it only ever tries to shutdown the CompletionQueue and join the thread once, but I'm happy if this has the same effect.)

Issue #960, opened 2019-10-16
Add profile and analyze samples


scotthart:
Add samples for
- ProfileQuery
- ProfileDml
- AnalyzeSql

devjgm:
I think this is now a dupe of #1112 #1113 #1114 , which are about adding integration tests for these functions. There are no official "samples" that are needed for these, so I think the integration tests are what was meant here.

PR #961, opened 2019-10-16
feat: add Profile and Analyze methods

### LINKED ISSUE #930: Determine how to request and return query_stats and query_plan
Providing methods on the appropriate Result type to return the QueryPlan and QueryStats is straightforward, the bigger questions are a) do we need to support this and b) how will the user request this.

scotthart:
fixes #930 
added ProfileQuery and ProfileDml to handle query_mode == PROFILE
added AnalyzeSql to handle query_mode == PLAN

created issue #960 to track adding sample code (this PR is big enough)

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/961)

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -274,7 +274,7 @@ class Client {
    * Profiling executes the query, provides the resulting rows, `ExecutionPlan`,
    * and execution statistics.
    *
-   * Operations inside read-write transactions might return `ABORTED`. If this
+   * Operations inside read-write transactions might return `kAborted`. If this
    * occurs, the application should restart the transaction from the beginning.
    *
    * Callers can optionally supply a `Transaction` or
@@ -283,12 +283,12 @@ class Client {
    * is used.
    *
    * @note Callers must consume all rows from the result before execution
-   * statistics and `ExecutionPlan` are available.
+   *     statistics and `ExecutionPlan` are available.
    *
    * @param statement The SQL statement to execute.
    *
    * @note No individual row in the `ProfileQueryResult` can exceed 100 MiB, and
-   * no column value can exceed 10 MiB.
+   *     no column value can exceed 10 MiB.
    */
   ProfileQueryResult ProfileQuery(SqlStatement statement);
 
@@ -353,7 +353,7 @@ class Client {
   /**
    * Profiles a SQL DML statement.
    *
-   * Profiling executes the DMl statement, provides the modified row count,
+   * Profiling executes the DML statement, provides the modified row count,
    * `ExecutionPlan`, and execution statistics.
    *
    * Operations inside read-write transactions might return `ABORTED`. If this
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -123,7 +123,7 @@ class Client {
   //@{
   /**
    * Reads rows from the database using key lookups and scans, as a simple
-   * key/value style alternative to `ExecuteSql()`.
+   * key/value style alternative to `ExecuteQuery()`.
    *
    * Callers can optionally supply a `Transaction` or
    * `Transaction::SingleUseOptions` used to create a single-use transaction -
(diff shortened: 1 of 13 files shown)


google/cloud/spanner/client.h
```diff
@@ -267,10 +267,51 @@ class Client {
   QueryResult ExecuteQuery(QueryPartition const& partition);
   //@}
 
+  //@{
+  /**
+   * Profiles a SQL query.
+   *
+   * Profiling executes the query, provides the resulting rows, `ExecutionPlan`,
+   * and execution statistics.
```
devjgm: Should we say something about the caller is responsible to consume all the returned rows form the query, and the profile information is not available until all the rows have been consumed?

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -220,6 +247,12 @@ class StatusOnlyResultSetSource : public internal::ResultSourceInterface {
 
 class DmlResultSetSource : public internal::ResultSourceInterface {
  public:
+  static StatusOr<std::unique_ptr<ResultSourceInterface>> Create(
```
devjgm: I think `DmlResult` and `QueryResult` are quite different types and would likely best be treated separately. In a future PR we may want to remove all connections between ResultSourceInterface and DmlResult, and I think it'll be a simplification.

No change needed in this PR, but we should probably discuss if this is a good idea or not.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -379,90 +416,139 @@ QueryResult ConnectionImpl::ExecuteQueryImpl(
...
   }
   if (s.has_begin()) {
     auto metadata = (*reader)->Metadata();
     if (!metadata || metadata->transaction().id().empty()) {
-      return QueryResult(
-          google::cloud::internal::make_unique<StatusOnlyResultSetSource>(
-              Status(StatusCode::kInternal,
-                     "Begin transaction requested but no transaction returned "
-                     "(in ExecuteQuery).")));
+      return T(google::cloud::internal::make_unique<StatusOnlyResultSetSource>(
```
devjgm: Ideally in the case of a DmlResult, this would just return the `Status`, not a `DmlResult` wrapping the Status. Is there anyway to do that now?

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -379,90 +416,139 @@ QueryResult ConnectionImpl::ExecuteQueryImpl(
...
     std::int64_t seqno, ExecuteSqlParams esp) {
-  if (!session) {
-    auto session_or = AllocateSession();
-    if (!session_or) {
-      return std::move(session_or).status();
+  return CommonQueryImpl<QueryResult>(
+      session, s, seqno, std::move(esp),
+      spanner_proto::ExecuteSqlRequest::NORMAL,
+      "Begin transaction requested but no transaction returned "
+      "(in ExecuteQuery).");
```
devjgm: This "in ExecuteQuery" parenthetical is the only place where this error message differs. Is this difference important? If not, we could avoid passing this error string, which would nice.

google/cloud/spanner/client.h
```diff
@@ -267,10 +267,51 @@ class Client {
...
+   * occurs, the application should restart the transaction from the beginning.
+   *
+   * Callers can optionally supply a `Transaction` or
+   * `Transaction::SingleUseOptions` used to create a single-use transaction -
+   * or neither, in which case a single-use transaction with default options
+   * is used.
+   *
+   * @param statement The SQL statement to execute.
+   *
+   * @note No individual row in the `ProfileQueryResult` can exceed 100 MiB, and
```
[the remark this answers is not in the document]
scotthart: It applies to all Read and Query API calls.

google/cloud/spanner/client.h
```diff
@@ -306,10 +347,43 @@ class Client {
   StatusOr<DmlResult> ExecuteDml(Transaction transaction,
                                  SqlStatement statement);
 
+  /**
+   * Profiles a SQL DML statement.
+   *
+   * Profiling executes the query, provides the modified row count,
```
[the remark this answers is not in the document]
scotthart: Missed that appearance of "query". It has been corrected.

google/cloud/spanner/client.h
scotthart: Done.

google/cloud/spanner/internal/connection_impl.cc
scotthart: ResultSetSource is an interface to reading either of the spanner protos: ResultSet or PartialResultSet. It's not specific to queries or dml operations.

google/cloud/spanner/internal/connection_impl.cc
scotthart: I'm not sure if the difference is important; it does add a small amount of useful diagnostic information.

google/cloud/spanner/internal/connection_impl.cc
scotthart: Yes, good catch.

google/cloud/spanner/internal/connection_impl.cc
devjgm: `ResultSet` is so trivial that using `ResultSetSource` interface to read it is adding complexity, IMO. But we can fix this in another PR.

google/cloud/spanner/internal/connection_impl.cc
devjgm: Asking callers to construct a potential error message for an operation that the caller does not call does not feel right. I think we should simply remove this argument from the function. I suspect the extra information will not be critical. If it ever is needed, the called function can compute the string itself based on the template parameter `T` that's specified.

mr-salty:
___
*[google/cloud/spanner/internal/connection_impl.cc, line 393 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/961#-LrVNECx-fvolMPH91TV:-LrVNECx-fvolMPH91TW:bopxg74) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/bdc76602170f3e57f40d273a18249f4c318581a7/google/cloud/spanner/internal/connection_impl.cc#L393)):*
> ```
> esp
> ```

FYI, you'll probably encounter some of these in rebase conflicts but I sent a PR to just change all of these `<goofy-abbreviation>p` names to just be called `params`...   I think all the new ones you added are `esp` so just `%s/\<esp\>/params/` throughout will probably take care of it.

mr-salty:
___
*[google/cloud/spanner/internal/connection_impl.cc, line 450 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/961#-Lrot8p6CvgnJHH3vOAS:-LssIYt_5GGG70EVThod:bf5xw9p) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/61284d4976b11e9636ff87257f00e7334fdd1ec3/google/cloud/spanner/internal/connection_impl.cc#L450)):*

That sounds like the least objectionable option here.
</blockquote></details>

I believe this change (making the members shared_ptr) is not correct - we want shared_ptrs of clones of the policy as Scott originally had, so each call gets its own state. The way it is now, if we have multiple threads that end up in here simultaneously they will share the same policy and conflict.

mr-salty:
___
*[google/cloud/spanner/internal/connection_impl.cc, line 450 at r5](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/961#-Lrot8p6CvgnJHH3vOAS:-LssdtJ72V4upklJhKXM:b-jg0n85) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/61284d4976b11e9636ff87257f00e7334fdd1ec3/google/cloud/spanner/internal/connection_impl.cc#L450)):*

I believe this change (making the members shared_ptr) is not correct - we want shared_ptrs of clones of the policy as Scott originally had, so each call gets its own state. The way it is now, if we have multiple threads that end up in here simultaneously they will share the same policy and conflict.
</blockquote></details>

My previous statement was incorrect, because we still clone() the shared_ptrs prior to using them.

However, this pattern does seem a bit error-prone.  Making the member variables `const` would prevent `OnFailure` from being called on them and disturbing the initial state - I'll send a PR for that.

PR #963, opened 2019-10-16
refactor!: split `ExecuteSqlParams` into Query/DML

mr-salty:
DML Queries cannot be partitioned; splitting these allows us to remove
the partition token from the DML variant so we don't have to worry about
handling it or not in the method bodies.

BREAKING CHANGE: This changes the `Connection` API which is public for testing.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/963)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -86,7 +86,7 @@ QueryResult Client::ExecuteQuery(Transaction transaction,
 }
 
 QueryResult Client::ExecuteQuery(QueryPartition const& partition) {
-  return conn_->ExecuteQuery(internal::MakeExecuteSqlParams(partition));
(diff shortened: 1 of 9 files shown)


coryan:
> Hm, not sure I've seen this failure before, but I also don't see what's in here that would cause a segfault (this is Windows Cmake)

I have been attributing those (maybe incorrectly) to another manifestation of grpc/grpc#16872.

scotthart:
This change is innocuous in isolation, but may complicate the pending changes I have in #961. Let's go ahead and merge this now, but based on the impact I may want to recombine them later.

mr-salty:
reverting this due to conflicts with work Scott is doing; I'll take another look once he's done.

Issue #387, opened 2019-08-20
Design a good solution for `SELECT *`


devjgm:
Our library should work for queries like `SELECT * FROM table;` As the library stands today, the user could only consume the results if they

a. Know exactly how many columns would be returned, and in what order
b. Know the types of each column so they could specify that as the arguments to `reader->Rows<T1, T2, T3, ...>()`

There may be cases where users do not know exactly how many columns will be returned, nor the order, and maybe not even the types of each column.

So, before we attempt to solve this problem, I think we need to better understand the *actual* use cases that we are solving for. To do this, we should look at what other Cloud Spanner libraries do (in other languages), talk w/ the Spanner team, etc.

We should also produce a small design doc clearly stating which use cases we're supporting and which (if any) we're not supporting. Then it should describe the design of how we plan to solve this.

devjgm:
FYI, we have a somewhat fleshed out design proposal for what to change here. A sketch of this design is shown in this godbolt link: https://godbolt.org/z/mDOClD

To implement new, user-facing API shown in the godbolt link, we will make the following changes to our existing API:

*   Remove class `Row<Ts…>`
*   Remove class `RowParser<Ts…>`
*   Remove method `QueryResult::Rows<Ts…>()`
*   Remove method `ReadResult::Rows<Ts…>()`

And
*  Add class `Row`
*  Add class `RowRange`
*  Add class `RowParser<Tuple>`
*  Add method `QueryResult::Rows()`
*  Add method `ReadResult::Rows()`

Some of the work items to accomplish this are:

- [ ] Remove use of `Row<Ts...>` from `keys.h`
- [ ] Remove use of `Row<Ts...> from `mutations.h`
- [ ] Refactor `internal::ResultSetSource` to communicate a full `vector<Value>` (i.e., a "row") at a time rather than a single `Value` at a time.
- [ ] This final step is probably a fairly big PR, I think this involves the rest of change required to implement this new API and close this issue.

PR #968, opened 2019-10-19
feat: adds Row class, not yet used

devjgm:
Part of #387 

This PR adds the `Row` class that will be used by #387. This class is added and tested in this PR, but it is not used yet. This is an attempt to break up a large PR into smaller PRs. So this PR just adds the class. A subsequent PR will start using this type.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/968)

### DIFF
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -31,10 +31,10 @@ namespace spanner {
 inline namespace SPANNER_CLIENT_NS {
 
 /**
- * A `Row` is a sequence of columns each with a name and an associated `Value.
+ * A `Row` is a sequence of columns each with a name and an associated `Value`.
  *
  * The `Row` class is a regular value type that may be copied, moved, assigned,
- * compared for equality, etc. Instance may be large if they hold lots of
+ * compared for equality, etc. Instances may be large if they hold lots of
  * `Value` data, so copy only when necessary.
  *
  * `Row` instances are typically returned as the result of queries or reads of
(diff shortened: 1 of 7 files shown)


devjgm:
/cc: @aryann I already removed spanner's `Row<T>` class template. This non-template `Row` is what we're moving toward. No action needed on your part; just FYI.

google/cloud/spanner/row.h
```diff
@@ -16,14 +16,186 @@
...
+  Row();
+
+  /**
+   * Constructs a `Row` with the given @p values and @p columns.
+   *
+   * @note columns must not be nullptr
+   * @note columns.size() must equal values.size()
+   */
+  explicit Row(std::vector<Value> values,
+               std::shared_ptr<const std::vector<std::string>> columns);
```
aryann: I spent some time today to think about how we might want to adapt this for the BigQuery client. I'm replying here so you're aware of the current thinking and if there are problems with our approach, you can let us know.

BigQuery returns data in response messages containing one or more rows in serialized Arrow format. We have a desire to not fully materialize every row as that can be quite inefficient. What I'm thinking about is an internal-only constructor like:

```cpp
Row(
  std::shared_ptr<arrow::Buffer> buf /* contains one of more row */,
  std::shared_ptr<arrow::Schema> schema,
  std::size_t offset_in_buf /* i.e., this is the row # in buf */ );
```

This construction defers all parsing until the user wants to fetch a column value (and even then, the parsing that is done is quite efficient).

One problem with this approach that we *may* need to solve later: if the user is reading the data, but sampling rows into a container, then the memory consumption will not be equivalent to the number of rows sampled, but equivalent to the number of arrow:Buffer objects that have to be kept around for those sampled rows. If this becomes a problem, we can add a function to materialize a row such that it doesn't rely on the shared buffer.

Another problem we'll need to figure out is how to map non-primitive types (e.g., lists, unions, structs) into C++ types without materializing too much data. For example, we should be able to have an STL-friendly list that is lazy. I noticed that the Value type in Spanner uses a vector, which is probably not great for us, though we can start with such an implementation to start. Let me know if you have any thoughts on this.

Thanks!

PR #975, opened 2019-10-23
feat!: Replace RunTransaction() with Client::Commit(functor)

### LINKED ISSUE #557: Replace RunTransaction() with Client::Commit(functor)
(See #534 for a description of some problems with the existing `RunTransaction()` function. That issue also presents an alternative design (called `RunCommitBlock`) that was discussed but we decided not to move forward with. This issue captures preferred idea to replace the existing `RunTransaction()`.)

We want to implement a "transaction runner" that is an overload of `Client::Commit`. This overload will take a user-provided functor instead of a `Transaction` and `std::vector<Mutation>`. Its interface should look like:

```cc
class Client {
 public:
  ...
  StatusOr<CommitResult> Commit(
    std::function<StatusOr<Mutations>(Transaction)> mutator);
  ...
};
```
This `Commit` overload will commit the transaction and optional mutations, along with handling rollbacks, exactly as `RunTransaction` does today. The only real changes from `RunTransaction` today are in the function's interface -- this one is much simpler for users to call.

NOTE: This overload will **replace** `RunTransaction`. That is, `RunTransaction` should be deleted.

devbww:
Replace `g:c:s:RunTransaction()` with a new `g:c:s:Client::Commit()`
overload.  This is similar to the existing `Commit()`, but you
specify how to create the mutations, instead of passing them
directly, which is what is needed to create the rerun loop.

The "mutator" callback no longer receives the target `Client` as
an argument because many mutators don't need it.  If you want
a client in your callback, you will need to bind it yourself.

The `Transaction::ReadWriteOptions` have been removed as there
are no options, and being able to preallocate RW transactions
pretty much requires that remains the case.

Session affinity over reruns is still to be addressed.

BREAKING CHANGE

Fixes #557.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/975)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -135,6 +135,74 @@ StatusOr<BatchDmlResult> Client::ExecuteBatchDml(
       {std::move(transaction), std::move(statements)});
 }
 
+StatusOr<CommitResult> Client::Commit(
+    std::function<StatusOr<Mutations>(Transaction)> const& mutator,
+    std::unique_ptr<TransactionRerunPolicy> rerun_policy,
+    std::unique_ptr<BackoffPolicy> backoff_policy) {
+  // The status-code discriminator of TransactionRerunPolicy.
+  using RerunnablePolicy = internal::SafeTransactionRerun;
+
+  for (int rerun = 0;; ++rerun) {
+    // TODO(#472): Make this transaction use the same session each time.
+    Transaction txn = MakeReadWriteTransaction();
+    StatusOr<Mutations> mutations;
+#if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
+    try {
+#endif
+      mutations = mutator(txn);
+#if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
+    } catch (...) {
+      auto rb_status = Rollback(txn);
+      if (!RerunnablePolicy::IsOk(rb_status)) {
+        GCP_LOG(WARNING) << "Rollback() failure in Client::Commit(): "
+                         << rb_status.message();
+      }
+      throw;
+    }
+#endif
+    auto status = mutations.status();
+    if (RerunnablePolicy::IsOk(status)) {
+      auto result = Commit(txn, *mutations);
+      status = result.status();
+      if (!RerunnablePolicy::IsTransientFailure(status)) {
+        return result;
+      }
+    } else {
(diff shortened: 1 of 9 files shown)


devbww:
> nit: Doxygen understand @p mutator as a reference to the parameter called mutator and it formats and links correctly.

Done (et seq).

devbww:
> FYI: I think it would be nice if this used some defaults set in the constructor or something, but this PR is better than what we have today

Yes, agreed.

google/cloud/spanner/client.cc
```diff
@@ -135,6 +135,74 @@ StatusOr<BatchDmlResult> Client::ExecuteBatchDml(
...
+    std::function<StatusOr<Mutations>(Transaction)> const& mutator,
+    std::unique_ptr<TransactionRerunPolicy> rerun_policy,
+    std::unique_ptr<BackoffPolicy> backoff_policy) {
+  // The status-code discriminator of TransactionRerunPolicy.
+  using RerunnablePolicy = internal::SafeTransactionRerun;
+
+  for (int rerun = 0;; ++rerun) {
+    // TODO(#472): Make this transaction use the same session each time.
+    Transaction txn = MakeReadWriteTransaction();
+    StatusOr<Mutations> mutations;
```
devjgm: Does this need to be declared here, or can it be declared on line 152 when it's initialized?

google/cloud/spanner/integration_tests/client_integration_test.cc
```diff
@@ -356,10 +353,8 @@ void CheckReadWithOptions(
         options_generator) {
   using RowValues = std::vector<Value>;
   std::vector<RowValues> expected_rows;
-  auto commit = RunTransaction(
-      client, Transaction::ReadWriteOptions{},
-      [&expected_rows](Client const&,
-                       Transaction const&) -> StatusOr<Mutations> {
+  auto commit = client.Commit(
+      [&expected_rows](Transaction const&) -> StatusOr<Mutations> {
```
devjgm: FYI, all these call site examples in here really show how much nicer this new API is than the old one. Big improvement, IMO.

google/cloud/spanner/client.cc
devbww: I'll say yes, it does need to be declared here, otherwise it would (`#if`-ly) be out-of-scope on lines 163 and beyond.

PR #980, opened 2019-10-25
feat: adds efficient move support to Value::get<string>()

### LINKED ISSUE #422: Add a `Value::get<T>() &&` overload to allow moving out of Value
This may be more efficient and worth allowing code to write something like:

```cc
Value v = ...
std::string s = std::move(v).get<std::string>();
```
This would allow the following line during row parsing to efficiently move values rather than copy them: https://github.com/googleapis/google-cloud-cpp-spanner/blob/e077503c43c0262c3db600fab17640e029bb36fc/google/cloud/spanner/row.h#L301

devjgm:
Fixes #422

Changes `Value::get<T>()` to efficiently move `std::string` values out
of the proto and directly to the caller. `std::string` is the only type
that can do this because it's the only type which is exactly the same in
the proto and outside of it. All other types require some type of
decoding, which necessarily requires some amount of copying from the
proto out to the caller.

Note: the unit tests verifying this behavior rely on unspecified behavior of std::string. Specifically that a moved-from large string ends up empty. This is not guaranteed, but it *appears* to the true in all cases that I've seen. So we rely on this behavior in the unit test. If this turns out to not be true on some platform, we'll have to adjust the unit tests. I've commented the tests about this.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/980)

### DIFF
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -287,6 +287,15 @@ StatusOr<std::string> Value::GetValue(std::string const&,
   return pv.string_value();
 }
 
+StatusOr<std::string> Value::GetValue(std::string const&,
+                                      google::protobuf::Value&& pv,
+                                      google::spanner::v1::Type const&) {
+  if (pv.kind_case() != google::protobuf::Value::kStringValue) {
+    return Status(StatusCode::kUnknown, "missing STRING");
+  }
+  return std::move(*pv.mutable_string_value());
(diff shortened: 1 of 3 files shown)


devjgm:
Yep, changed to `feat:`, thanks for pointing that out. I also spotted a typo in a comment that I changed: frong -> from.

PR #999, opened 2019-10-29
chore: add feature template; update bug template

mr-salty:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/999)


mr-salty:
FYI, most of the text for the FR one came from the default thing github supplies, although I edited it to make it more similar to the bug report template.

PR #1004, opened 2019-10-30
ci: upgrade builds to use Bazel 1.0

coryan:
Now that Bazel reached 1.0 and it is expected to follow semver, it seems
better to test with that version. The downside is that we may not notice
breakage for users in the 0.28.0-0.29.0 range. But it seems that Bazel
users are rapidly moving towards using 1.0 anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1004)


coryan:
PTAL. We finally have all the CI builds using Bazel-1.0 for google-cloud-cpp-spanner.

PR #1009, opened 2019-10-30
style: use the re-running Commit() in the samples

devbww:
We recommend that users call the Commit() function that reruns the
mutating functor after a read-write transaction aborts.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1009)


devbww:
I'm not sure a new `Commit()` overload would help as the type of the lambda has to be unambiguously determined first.

I any case, we only need the trailing return type when the lamba is sometimes returning a `Status` and sometimes a `Mutations`, which means we can simplify four of the samples.

devbww:
Thanks.

PR #1029, opened 2019-11-07
feat: implement spanner_create_storing_index example

### LINKED ISSUE #480: Implement spanner_create_storing_index example.
Use the admin API to update the DDL with a new index with the `STORING` directive:

https://github.com/GoogleCloudPlatform/golang-samples/blob/40a33f7ff5de11da00a6715eaa29ce45fdf92387/spanner/spanner_snippets/snippet.go#L696-L715

tmatsuo:
fixes #480

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1029)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -529,6 +529,36 @@ void AddColumn(std::vector<std::string> const& argv) {
   (argv[0], argv[1], argv[2]);
 }
 
+void AddStoringIndex(std::vector<std::string> const& argv) {
+  if (argv.size() != 3) {
+    throw std::runtime_error(
+        "add-storing-index <project-id> <instance-id> <database-id>");
+  }
+  // [START spanner_create_storing_index]
+  using google::cloud::future;
+  using google::cloud::StatusOr;
+  [](std::string const& project_id, std::string const& instance_id,
+     std::string const& database_id) {
+    google::cloud::spanner::DatabaseAdminClient client;
+    google::cloud::spanner::Database database(project_id, instance_id,
+                                              database_id);
+    future<StatusOr<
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -529,6 +529,36 @@ void AddColumn(std::vector<std::string> const& argv) {
...
+    google::cloud::spanner::DatabaseAdminClient client;
+    google::cloud::spanner::Database database(project_id, instance_id,
+                                              database_id);
+    future<StatusOr<
+        google::spanner::admin::database::v1::UpdateDatabaseDdlMetadata>>
+        future = client.UpdateDatabase(database, {R"""(
+            CREATE INDEX AlbumsByAlbumTitle2 ON Albums(AlbumTitle)
+                STORING (MarketingBudget))"""});
+    StatusOr<google::spanner::admin::database::v1::UpdateDatabaseDdlMetadata>
+        result = future.get();
```
devjgm: Nit: I think it's nice to name these StatusOr variables after the type/object that they (might) contain. Like in this case, maybe name it `metadata`. That way, the rest of the code reads nicely as if the caller has a (smart)pointer to a metadata object, which they kinda do.

tmatsuo: Thanks, renamed the variable. I'll merge when green

tmatsuo:
Windows CMake build failed 3 times in a row with segfault.

tmatsuo:
I'll re-run the test again and try to monitor the VM which runs the test.

tmatsuo:
This time samples passes, but we got:
51 - instance_admin_integration_test (Timeout)

PR #1030, opened 2019-11-08
feat!: replaces PartitionOptions proto with struct

devjgm:
Fixes #452 

Now that PartitionOptions is a struct, it should likely be passed by const-ref and not moved. Well, clang-tidy yells at me otherwise.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1030)

PR #1035, opened 2019-11-11
feat!: replaces PartitionOptions proto with struct

### LINKED ISSUE #452: Consider: writing a wrapper for google::spanner::v1::PartitionOptions.
This is (AFAICT) the only naked proto we expose in the non-admin APIs for the client library.  Currently here:

https://github.com/googleapis/google-cloud-cpp-spanner/blob/87c2d3509872e539ffa8b25db6944779e27f944d/google/cloud/spanner/client_options.h#L111

(but will be moving soon).  I wonder if we should write a wrapper, as we wrap all the other proto types.

devjgm:
Fixes #452

Now that PartitionOptions is a struct, it should likely be passed by const-ref and not moved. Well, clang-tidy yells at me otherwise.

This is #1030 take two, because I accidentally closed that PR instead of merging it. Why didn't I just reopen that PR? Because I pushed the branch again after and now it won't let me reopen that PR.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1035)


mr-salty:
taking it mostly on faith this is the same as the other PR and not a trojan :)

PR #1041, opened 2019-11-12
feat: associate `SpannerStub` with a `Session`

mr-salty:
`SessionPool` manages assigning a `SpannerStub` to each `Session`
`ConnectionImpl` obtains the stub to use for a call from the `Session`
and no longer retains any knowledge of the stub.

This will enable the use of multiple stubs per `Connection`; it's
important for a `Session` to remain associated with the stub/channel
that created it, otherwise there is a performance penalty.

Part of #307

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -106,11 +106,10 @@ ConnectionImpl::ConnectionImpl(Database db, std::shared_ptr<SpannerStub> stub,
                                std::unique_ptr<RetryPolicy> retry_policy,
                                std::unique_ptr<BackoffPolicy> backoff_policy)
     : db_(std::move(db)),
-      stub_(std::move(stub)),
       retry_policy_prototype_(std::move(retry_policy)),
       backoff_policy_prototype_(std::move(backoff_policy)),
       session_pool_(std::make_shared<SessionPool>(
-          db_, stub_, retry_policy_prototype_->clone(),
+          db_, std::move(stub), retry_policy_prototype_->clone(),
           backoff_policy_prototype_->clone())) {}
 
 RowStream ConnectionImpl::Read(ReadParams params) {
@@ -302,6 +301,8 @@ Status ConnectionImpl::PrepareSession(SessionHolder& session,
       return std::move(session_or).status();
     }
     session = std::move(*session_or);
+  } else {
+    session_pool_->AssignStubIfNeeded(session);
   }
   return Status();
 }
@@ -328,10 +329,10 @@ RowStream ConnectionImpl::ReadImpl(SessionHolder& session,
     request.set_partition_token(*std::move(params.partition_token));
   }
 
-  auto const& stub = stub_;
   // Capture a copy of `stub` to ensure the `shared_ptr<>` remains valid through
   // the lifetime of the lambda. Note that the local variable `stub` is a
   // reference to avoid increasing refcounts twice, but the capture is by value.
+  auto const& stub = session->stub();
   auto factory = [stub, request](std::string const& resume_token) mutable {
     request.set_resume_token(resume_token);
     auto context = google::cloud::internal::make_unique<grpc::ClientContext>();
@@ -382,9 +383,9 @@ StatusOr<std::vector<ReadPartition>> ConnectionImpl::PartitionReadImpl(
   auto response = internal::RetryLoop(
       retry_policy_prototype_->clone(), backoff_policy_prototype_->clone(),
       true,
-      [this](grpc::ClientContext& context,
-             spanner_proto::PartitionReadRequest const& request) {
-        return stub_->PartitionRead(context, request);
+      [&session](grpc::ClientContext& context,
+                 spanner_proto::PartitionReadRequest const& request) {
+        return session->stub()->PartitionRead(context, request);
       },
       request, __func__);
   if (!response.ok()) {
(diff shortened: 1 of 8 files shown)


mr-salty:
___
*[google/cloud/spanner/internal/connection_impl.cc, line 518 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1041#-LtSZUpO-suhvkzjHLHY:-LtSZUpO-suhvkzjHLHZ:bcd3est) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/cda80125c58c58c87a847d510ad14b08c15b9343/google/cloud/spanner/internal/connection_impl.cc#L518)):*
> ```
>   auto const& stub = session->stub();
> ```

FYI, I suspect the `const&` here does nothing now; I was thinking something like the compiler might see that it's doing lifetime extension and can effectively std::move it into the lambda capture, but I guess if it's smart enough to figure that out it could also figure it out if it was just a value.  So, should I delete it?

Is there any way to accomplish the goal of not doing an extra ref/unref other than having `Session` return a `shared_ptr<>&`?  I didn't want to do that because it seems dangerous. 

Does just doing `session->stub()->Call()` now bump the reference count as well or can that be optimized away? I did some googling but couldn't really find anything. I didn't play with godbolt but I guess I could try that.

google/cloud/spanner/internal/session.h
```diff
@@ -42,9 +43,12 @@ class Session {
   Session& operator=(Session&&) = delete;
 
   std::string const& session_name() const { return session_name_; }
+  std::shared_ptr<SpannerStub> stub() const { return stub_; }
+  void set_stub(std::shared_ptr<SpannerStub> stub) { stub_ = stub; }
```
devbww: move(stub)?

google/cloud/spanner/internal/session_pool_test.cc
```diff
@@ -233,6 +233,18 @@ TEST(SessionPool, MaxSessionsBlockUntilRelease) {
...
 
+TEST(SessionPool, AssignStubIfNeeded) {
+  auto mock = std::make_shared<spanner_testing::MockSpannerStub>();
+  auto db = Database("project", "instance", "database");
+  auto pool = MakeSessionPool(db, mock);
+  auto session = MakeDissociatedSessionHolder("session_id", /*stub=*/nullptr);
+  EXPECT_EQ(session->stub(), nullptr);
+  pool->AssignStubIfNeeded(session);
+  EXPECT_EQ(session->stub(), mock);
+  pool->AssignStubIfNeeded(session);
```
devbww: What does this replay purport to test?

google/cloud/spanner/transaction.cc
```diff
@@ -97,7 +97,7 @@ Transaction::Transaction(std::string session_id, std::string transaction_id) {
   google::spanner::v1::TransactionSelector selector;
   selector.set_id(std::move(transaction_id));
   impl_ = std::make_shared<internal::TransactionImpl>(
-      internal::MakeDissociatedSessionHolder(std::move(session_id)),
+      internal::MakeDissociatedSessionHolder(std::move(session_id), nullptr),
```
devbww: At the moment it doesn't look like MakeDissociatedSessionHolder() really needs to be a template.  Might it just take a `std::string session_name` and then be able to provide the `nullptr` to the `Session` ctor itself?

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -512,11 +512,10 @@ StatusOr<ResultType> ConnectionImpl::CommonDmlImpl(
...
   }
-  // Capture a copy of of these member variables to ensure the `shared_ptr<>`
-  // remains valid through the lifetime of the lambda. Note that the local
-  // variables are a reference to avoid increasing refcounts twice, but the
-  // capture is by value.
-  auto const& stub = stub_;
+  // Capture a copy of of these to ensure the `shared_ptr<>` remains valid
+  // through the lifetime of the lambda. Note that the local variables are a
+  // reference to avoid increasing refcounts twice, but the capture is by value.
+  auto const& stub = session->stub();
```
devbww: > FYI, I suspect the const& here does nothing now; I was thinking something like the compiler might see that it's doing lifetime extension and can effectively std::move it into the lambda capture, but I guess if it's smart enough to figure that out it could also figure it out if it was just a value. So, should I delete it?

Yes, I don't think this is doing anything different to `auto const stub = session->stub()`.  You still need to give the stub a name though, so that `retry_resume_fn` can capture it.

> Is there any way to accomplish the goal of not doing an extra ref/unref other than having Session return a shared_ptr<>&? I didn't want to do that because it seems dangerous.

I don't think so, as the refcount has already gone up before `session->stub()` returns by value.  As to whether it is "dangerous" or not (I assume you meant `Session::stub()`), I'll say that depends on how you use it.  It seems OK to me, as long as you're done a `PrepareSession(session)` "_before_" so that `session->stub_` is const.

> Does just doing session->stub()->Call() now bump the reference count as well or can that be optimized away?

I would say it does incr/decr the refcount.  There is nothing, for example, that might stop a concurrent "`session->clear_stub()`" call from happening between the return of `stub()` and the call of `Call()`.

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -141,6 +141,14 @@ StatusOr<SessionHolder> SessionPool::Allocate(bool dissociate_from_pool) {
   }
 }
 
+void SessionPool::AssignStubIfNeeded(SessionHolder& session) {
+  if (session->stub() == nullptr) {
```
devbww: Does it make any additional sense to move this logic into `Session`.  That is, `session->AssignStubIfNeeded(stub_)`?

google/cloud/spanner/internal/session.h
```diff
@@ -44,7 +45,11 @@ class Session {
   std::string const& session_name() const { return session_name_; }
 
  private:
-  std::string session_name_;
+  friend class SessionPool;  // for access to stub()
```
devbww: PS: Is it worth it to try to hide stub()?

PR #1050, opened 2019-11-13
feat: add some plumbing to enable multiple channels

mr-salty:
Note that `Client` might create multiple channels but the `SessionPool`
only uses the first one right now. I'll add proper support for multiple
channels in a subsequent change but wanted to get some of the more
mechanical parts out of the way.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1050)

### DIFF
--- a/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
+++ b/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
@@ -314,7 +314,8 @@ class ReadExperiment : public Experiment {
       auto options = cs::ConnectionOptions().set_channel_pool_domain(
           "task:" + std::to_string(i));
       clients.emplace_back(cs::Client(cs::MakeConnection(database, options)));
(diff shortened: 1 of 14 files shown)


mr-salty:
___
*[google/cloud/spanner/client.cc, line 226 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1050#-LtaieCGDxbJ8BkKhYTu:-LtaieCGDxbJ8BkKhYTv:bb12581) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/83a018e573f722d125087be4800b1e336c144b29/google/cloud/spanner/client.cc#L226)):*
> ```cpp
>   stubs.reserve(num_channels);
>   for (int i = 0; i < num_channels; ++i) {
>     stubs.push_back(internal::CreateDefaultSpannerStub(options));
> ```

@coryan to resolve our discussion from yesterday, should I modify the `channel_pool_domain` for each one of these channels?  append the value of `i` to each of them or similar? kind of a pain that `options` isn't mutable.

PR #1051, opened 2019-11-13
feat: measure CPU for multiple rows per query

coryan:
Change the CPU overhead benchmark to fetch multiple rows per
query, which amortizes the RPC overhead.

I made several "cleanups" because the code had a lot of duplication
between the "using stubs" vs. "using client" cases.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1051)


coryan:
PTAL.

PR #1072, opened 2019-11-17
refactor: reduce code duplication in samples

coryan:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1072)


google/cloud/spanner/samples/samples.cc
```diff
@@ -1667,6 +1577,26 @@ int RunOneCommand(std::vector<std::string> argv) {
                                   make_command(sample_name, sample));
   };
 
+  using DatabaseAdminSampleFunction =
+      void (*)(google::cloud::spanner::DatabaseAdminClient, std::string const&,
+               std::string const&, std::string const&);
+  auto make_database_command_entry = [](char const* sample_name,
```
devjgm: Can `sample_name` be a `std::string const&`, or does it need to be a c-string?

PR #1076, opened 2019-11-18
feat: change CPU benchmark to use multiple columns

coryan:
To be fair, also separate the response via stubs into "rows", and fix some problems
with session setup on the Stub case.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1076)

### DIFF
--- a/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
+++ b/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
@@ -79,7 +79,11 @@ struct RowCpuSample {
   Status status;
 };
 
-using SampleSink = std::function<void(std::vector<RowCpuSample>)>;
+std::ostream& operator<<(std::ostream& os, RowCpuSample const& s) {
+  return os << std::boolalpha << s.client_count << ',' << s.thread_count << ','
+            << s.using_stub << ',' << s.row_count << ',' << s.elapsed.count()
+            << ',' << s.cpu_time.count() << ',' << s.status.code();
+}
 
 class Experiment {
  public:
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
```diff
@@ -139,6 +139,9 @@ int main(int argc, char* argv[]) {
       google::cloud::spanner_testing::RandomDatabaseName(generator));
   config.database_id = database.database_id();
 
+  // Once the configuration is and the database name set, print everything out.
```
devbww: "Once the configuration is" what?

google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
```diff
@@ -402,20 +418,37 @@ class ReadExperiment : public Experiment {
...
-        if (!result.chunked_value()) {
-          row_count += result.values_size() / 2;
+        if (result.chunked_value()) {
+          // We do not handle chunked values in the benchmark.
+          continue;
+        }
+        row.resize(columns.size());
+        std::size_t index = 0;
+        for (auto& value : *result.mutable_values()) {
+          row[index++] = std::move(value);
```
devbww: Opt: move post ++ to pre ++ on next line?

coryan: Good idea, fixed.

google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
coryan: Thanks, fixed.

PR #1079, opened 2019-11-18
fix: execute each rerun of an aborted transaction in the same session

### LINKED ISSUE #472: transaction retry should re-use the same session
From https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.TransactionOptions

> Retrying Aborted Transactions
> When a transaction aborts, the application can choose to retry the whole transaction again. To maximize the chances of successfully committing the retry, the client should execute the retry in the same session as the original attempt. The original session's lock priority increases with each consecutive abort, meaning that each attempt has a slightly better chance of success than the previous.

devbww:
Fixes #472.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1079)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -142,9 +142,8 @@ StatusOr<CommitResult> Client::Commit(
   // The status-code discriminator of TransactionRerunPolicy.
   using RerunnablePolicy = internal::SafeTransactionRerun;
 
+  Transaction txn = MakeReadWriteTransaction();
   for (int rerun = 0;; ++rerun) {
-    // TODO(#472): Make this transaction use the same session each time.
-    Transaction txn = MakeReadWriteTransaction();
     StatusOr<Mutations> mutations;
 #if GOOGLE_CLOUD_CPP_HAVE_EXCEPTIONS
     try {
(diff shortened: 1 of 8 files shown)


devbww:
> Logically I look at this as re-running txn, and having to go through some slightly odd machinations to do it... but re-specifying the opts seems weird to me.

Our re-running `Commit()` doesn't take any `ReadWriteOptions`, but you could imagine one that does that would look like ...
```
  Transaction::ReadWriteOpts opts = ...;
  Transaction txn = MakeReadWriteTransaction(opts);
  for (...) {
    ...
    txn = MakeReadWriteTransaction(std::move(txn), opts);
  }
```
That is, it is just up to the caller to (re)specify the options, rather than trying to reconstitute them from the argument `Transaction`.

mr-salty:
>     txn = MakeReadWriteTransaction(std::move(txn), opts);
>   
> That is, it is just up to the caller to (re)specify the options, rather than trying to reconstitute them from the argument `Transaction`.

yeah, my thinking is we may devise a way to just do 'txn->PrepareForRerun()` or similar (which IIRC we discussed and pronounced infeasible).  but, like I said, I don't feel strongly about it.

PR #1085, opened 2019-11-19
bug: fix year formatting in spanner::Date

### LINKED ISSUE #1083: Pad years to 4 digits in spanner::Date objects.
When converting a `spanner::Date` to a proto we should pad the year to 4 digits.  Easy enough to fix but we should write a unit test (and probably an integration test too). See [the documentation](https://cloud.google.com/spanner/docs/data-types#date-type) for more details.

coryan:
The year is required to be 0-padded to 4 digits, the month and date
padding seems to be optional.

Introduce a new integration test to verify data types can be written and
read back.

Fixes #1083

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1085)

### DIFF
--- a/google/cloud/spanner/integration_tests/data_types_integration_test.cc
+++ b/google/cloud/spanner/integration_tests/data_types_integration_test.cc
@@ -1,0 +1,124 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "google/cloud/spanner/client.h"
+#include "google/cloud/spanner/database.h"
+#include "google/cloud/spanner/database_admin_client.h"
+#include "google/cloud/spanner/mutations.h"
+#include "google/cloud/spanner/testing/database_environment.h"
+#include "google/cloud/internal/getenv.h"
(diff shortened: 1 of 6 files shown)


google/cloud/spanner/internal/date.cc
```diff
@@ -25,7 +25,7 @@ namespace internal {
 
 std::string DateToString(Date d) {
   std::array<char, sizeof "-9223372036854775808-01-01"> buf;
-  std::snprintf(buf.data(), buf.size(), "%" PRId64 "-%02d-%02d", d.year(),
+  std::snprintf(buf.data(), buf.size(), "%04" PRId64 "-%02d-%02d", d.year(),
```
devbww: Oops.  Thanks.

google/cloud/spanner/integration_tests/data_types_integration_test.cc
```diff
@@ -0,0 +1,101 @@
...
+  RowType read_back;
+  auto commit_result = client_->Commit(
+      [&client, &read_back](Transaction const& txn) -> StatusOr<Mutations> {
+        auto result = client.ExecuteDml(
+            txn,
+            SqlStatement(
+                "INSERT INTO DataTypes (Id, DateValue, StringValue) "
+                "VALUES(@id, @date, @event)",
+                {{"id", Value("ReadWriteDate-1")},
+                 {"date", Value(Date(161, 3, 8))},
```
devbww: Perhaps we should add a "> 9999" case to see how they handle their "YYYY: Four-digit year" edict.

google/cloud/spanner/internal/date.cc
coryan: no problem.

google/cloud/spanner/integration_tests/data_types_integration_test.cc
coryan: Anything over 9999 results in an error.  That is consistent with the [documentation](https://cloud.google.com/spanner/docs/data-types#date-type); which says  the valid range is `0001-01-01` to `9999-12-31`

PR #1092, opened 2019-11-20
feat: change GetCurrentRow to GetSingularRow

devjgm:
After some discussion with the local folks, we decided to change the semantics and name of `GetCurrentRow` that was added in #1074. Previously `GetCurrentRow` would return the first row in a range that might potentially contain multiple rows. The observation is that those may not be the best or safest semantics for callers.

The use case we're trying to address is when a callers does a query/read and they know that only one row should be returned. Maybe they used `LIMIT 1`, or maybe they specified a primary key to guarantee this. In any case, if they know that only one result can be returned, it would be nice to let them conveniently access that one row without having to create a loop, break out, etc.

The problem with `GetCurrentRow` was that _if_ the row range happened to have _more_ than one row, there would be no error and the user may never notice this. So it seems safer to clearly name this function and change its semantics so that it will produce an error if the given range does not contain *exactly* a single row.

We also considered requiring the caller to pass an rvalue (i.e., std::move the range into the argument). The reason I don't enforce this is because that would then prevent callers for subsequently calling the `RowStream::ReadTimestamp` function.

Adding @devbww as the reviewer since he initially raised some of these questions. And cc: @coryan  and @mr-salty as potentially interested observers who are welcome to comment if they want.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1092)

### DIFF
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -432,25 +432,33 @@ class StreamOf {
 };
 
 /**
- * Returns the current row from the given range, if any.
+ * Returns the only row from a range that contains exactly one row.
  *
- * If there are no rows left, an error `Status` is returned. This is a
- * convenience function that can be useful if the caller knows there will be
- * only a single row returned.
+ * An error is returned if the given range does not contain exactly one row.
+ * This is a convenience function that may be useful when the caller knows that
+ * a range should contain exactly one row, such as when `LIMIT 1` is used in an
+ * SQL query, or when a read is performed on a guaranteed unique key such that
+ * only a single row could possibly match. In cases where the caller does not
+ * know how many rows may be returned, they should instead consume the range in
+ * a loop.
+ *
+ * @warning Due to the fact that a `RowStreamIteartor` is an input iterator,
+ *     this function may consume the first element in the range, even in cases
+ *     where an error is returned. But again, this function should not be used
+ *     if @p range might contain multiple rows.
  */
 template <typename RowRange>
-auto GetCurrentRow(RowRange&& range) -> typename std::decay<
+auto GetSingularRow(RowRange&& range) -> typename std::decay<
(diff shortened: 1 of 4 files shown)


google/cloud/spanner/row.h
```diff
@@ -432,18 +432,33 @@ class StreamOf {
...
   auto it = std::forward<RowRange>(range).begin();
-  if (it != std::forward<RowRange>(range).end()) return *it;
-  return Status(StatusCode::kResourceExhausted, "No more rows");
+  if (it != e) {
+    auto row = std::move(*it);
+    if (++it == e) return row;
+  }
+  return Status(StatusCode::kInvalidArgument,
+                std::string(it != e ? "too many" : "no") +
+                    " rows (expected exactly one)");
```
devbww: Opt: I might be tempted by ...
```
if (it == e) return Status(..., "no rows");
auto row = std::move(*it);
if (++it != e) return Status(..., "more than one row");
return row;
```
but your choice.

devjgm: Done.

coryan:
This is always used as `spanner::GetSingularRow(spanner::StreamOf<RowType>(stream))`, WDYT if we change it so the usage is `spanner::GetSingularRow<RowType>(stream)`?

devjgm:
> This is always used as `spanner::GetSingularRow(spanner::StreamOf<RowType>(stream))`, WDYT if we change it so the usage is `spanner::GetSingularRow<RowType>(stream)`?

We could do that. It would make the cases where users want to parse the row to a `RowType` tuple easier. But it would make any case where a user didn't want to use a "StreamOf<T>" impossible. As you pointed out, our 3 uses currently all use `StreamOf<T>`, so this change would work now. But I'm not sure our 3 existing uses suggest that all future uses will be the same.

Another option is that we could add another function (or an overload) so that we support both.

Thoughts?

PR #1095, opened 2019-11-21
feat: benchmark CPU overhead for UPDATE

### LINKED ISSUE #1016: Implement CPU time benchmarks for UPDATE
This is one of the work items coming out of #539, and depends on #1013.  Once the scaffolding is implemented we need to add the following benchmarks to it:

- [ ] The CPU time needed to perform a DML UPDATE with a  column with a single 1KiB STRING value
- [ ] The CPU time needed to perform a DML UPDATE with a  column with a single 1KiB BYTE value
- [ ] The CPU time needed to perform a DML UPDATE with a  column with a single INT64 value
- [ ] The CPU time needed to perform a DML UPDATE with a  column with a single FLOAT value
- [ ] The CPU time needed to perform a DML UPDATE with a  column with a single TIMESTAMP value
- [ ] The CPU time needed to perform a DML UPDATE with a  column with a single DATE value

coryan:
Add new benchmarks to measure the CPU overhead (if any) of making
`UPDATE` queries via the client library vs. the raw gRPC code. The
benchmarks can measure the overhead for different data types.

Fixes #1016

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1095)

### DIFF
--- a/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
+++ b/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
@@ -411,8 +411,9 @@ class ExperimentImpl {
     std::lock_guard<std::mutex> lk(mu_);
     std::copy(samples.begin(), samples.end(),
               std::ostream_iterator<RowCpuSample>(std::cout, "\n"));
-    auto it = std::find_if(samples.begin(), samples.end(),
-        [](RowCpuSample const& x) { return !x.status.ok(); });
+    auto it =
+        std::find_if(samples.begin(), samples.end(),
+                     [](RowCpuSample const& x) { return !x.status.ok(); });
     if (it == samples.end()) {
       return;
     }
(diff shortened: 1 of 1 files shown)


coryan:
I need to hold off on this for a minute, some results make no sense.

coryan:
Good for review now.

google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
```diff
@@ -395,13 +381,147 @@ class ReadExperiment : public Experiment {
...
+    for (std::int64_t key = 0; key != config.table_size; ++key) {
+      // Each thread does a fraction of the key space.
+      if (key % task_count != task_id) continue;
+      // Have one of the threads report progress about 50 times.
+      if (task_id == 0 && key % report_period == 0) {
+        std::cout << '.' << std::flush;
+      }
+      mutation.EmplaceRow(key, value0, value1, value2, value3, value4, value5,
+                          value6, value7, value8, value9);
+      current_mutations++;
```
devjgm: Nit: Using post-increment when you don't need post-increment semantics will cause some human readers to skim back in the code to find the type of the variable being post-incremented to see if the use of post-increment is a pessimization or equivalent to pre-increment. As a general rule of thumb, I'd say use post-increment IFF you need post-increment semantics. Otherwise, use pre-increment.

coryan: Done and rebased.

Issue #512, opened 2019-08-30
[Flake]: CreateDatabase fails with UNAVAILABLE.


coryan:
We just a build failure because the `CreateDatabase` operation failed, and we do not retry this operation.  Maybe we should consider `CreateDatabase` idempotent (it can only succeed the one time, right?  The second time it would fail with `kAlreadyExists` I think), **or** we could configure the tests to retry even non-idempotent operations because the database has randomly generated database_id so we know all failures are "safe".

https://source.cloud.google.com/results/invocations/eee411d8-1d59-46a5-b042-9d6a7965e31d/targets

```
================================================================
Running the integration tests Fri Aug 30 15:14:01 UTC 2019
================================================================
Loading:
Loading: 0 packages loaded
INFO: Build option --test_env has changed, discarding analysis cache.
Analyzing: 41 targets (0 packages loaded, 0 targets configured)
DEBUG: Rule 'boringssl' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "93ddb1d452da9a6faf784d99bf66dd102db0aefa3ee341ff258d285732800c3f"
DEBUG: Call stack for the definition of repository 'boringssl' which is a http_archive (rule definition at /h/.cache/bazel/_bazel_kbuilder/a7ecd8237744645c5d189c197108d6d2/external/bazel_tools/tools/build_defs/repo/http.bzl:237:16):
 - /h/.cache/bazel/_bazel_kbuilder/a7ecd8237744645c5d189c197108d6d2/external/com_github_grpc_grpc/bazel/grpc_deps.bzl:110:9
 - /v/WORKSPACE:40:1
INFO: Analyzed 41 targets (0 packages loaded, 2799 targets configured).
Building: checking cached actions
INFO: Found 36 targets and 5 test targets...
[1 / 84] [Prepa] BazelWorkspaceStatusAction stable-status.txt
[1,472 / 1,476] Testing //google/cloud/spanner/integration_tests:spanner_client_database_admin_integration_test; 0s local ... (4 actions running)
[1,472 / 1,476] Testing //google/cloud/spanner/integration_tests:spanner_client_database_admin_integration_test; 11s local ... (4 actions running)
[1,472 / 1,476] Testing //google/cloud/spanner/integration_tests:spanner_client_database_admin_integration_test; 20s local ... (4 actions running)
[...]

PR #1101, opened 2019-11-22
feat: open the grpc::Channel in background for DatastoreAdminStub

tmatsuo:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1101)

### DIFF
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -125,6 +125,7 @@ add_library(
     instance_admin_connection.h
     internal/api_client_header.cc
     internal/api_client_header.h
+    internal/background_channel_opener.h
     internal/background_threads_impl.cc
     internal/background_threads_impl.h
     internal/build_info.h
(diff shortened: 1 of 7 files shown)


tmatsuo:
@coryan Added a test for the helper function.

It turned out that, `grpc::Channel.GetState(true)` only change the state to `CONNECTING`. To be honest, I'm not 100% sure that it will fix the flake in #512

It's not quite ready to merge, but can you take a quick look and let me know what you think? Is it going the right direction?

coryan:
We should probably close this and take another look later then.  Maybe we can do a `Get*` operation in the background to force a connection, but seems like all of this needs more research.

tmatsuo:
Ok, I'll close this. I'll keep the branch for a while just for reference.

The tests passed except for windows CMake build:
https://source.cloud.google.com/results/invocations/d31cfff6-e61a-4435-97c7-22b0b45f7521/targets

PR #1109, opened 2019-11-22
feat: add SessionPoolOptions with labels to MakeConnection()

### LINKED ISSUE #421: Support "labels" in the client.
The client (and the sessions associated with a client) should have "labels" associated with them:

https://github.com/googleapis/googleapis/blob/9c9f778aedde02f9826d2ae5d0f9c96409ba0f25/google/spanner/v1/spanner.proto#L323-L332

devbww:
Moves `SessionPoolOptions` out of the `internal` namespace,
and adds a `std::map<std::string, std::string> labels` field.
Plumbs those labels through to the `BatchCreateSessions()`
call.

Fixes #421.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1109)

### DIFF
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -193,6 +193,7 @@ add_library(
     retry_policy.h
     row.cc
     row.h
+    session_pool_options.h
     sql_statement.cc
     sql_statement.h
     timestamp.h
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -219,24 +219,29 @@ StatusOr<PartitionedDmlResult> Client::ExecutePartitionedDml(
   return conn_->ExecutePartitionedDml({std::move(statement)});
 }
 
-std::shared_ptr<Connection> MakeConnection(Database const& db,
-                                           ConnectionOptions const& options) {
-  return MakeConnection(db, options, internal::DefaultConnectionRetryPolicy(),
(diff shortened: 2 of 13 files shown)


mr-salty:
___
*[google/cloud/spanner/client.h, line 549 at r2](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1109#-LuZSt0z6vFHLXfY0vWo:-LuZSt0z6vFHLXfY0vWp:bf5p8ts) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/edcedd766d72a57a23e052fb3271d554f9423b0e/google/cloud/spanner/client.h#L549)):*
> ```objc
> 
> /**
>  * @copydoc MakeConnection(Database const&, ConnectionOptions const&, SessionPoolOptions)
> ```

[clang-tidy](https://source.cloud.google.com/results/invocations/ed1e09fe-1ed8-4ba8-a1a3-7e4baedb42b9/targets/cloud-cpp%2Fspanner%2Fdocker%2Fclang-tidy-presubmit/log) thinks this line should wrap

devbww:
> clang-tidy thinks this line should wrap

Fixed in #1122.

devbww:
> this can be deleted

Done.

devbww:
> So, I'm guessing SessionPoolOptions doesn't have that issue due to the map, but it seems really odd that these are different, and that will show up at the call site

Ack, although for now it is what it is.

Issue #145, opened 2019-06-27
Reconcile implementations of RFC3339 formatting and parsing


devbww:
Reconcile the following implementations of date/time formatting and parsing, and find a good home for a common library.

```
googleapis/google-cloud-cpp-common:google/cloud/internal/format_time_point.h
googleapis/google-cloud-cpp-common:google/cloud/internal/parse_rfc3339.h
googleapis/google-cloud-cpp-spanner:google/cloud/spanner/internal/time.h
googleapis/google-cloud-cpp-spanner:google/cloud/spanner/internal/date.h
```

devbww:
For the record, the Spanner implementation can now be found in ...
```
google/cloud/spanner/timestamp.h
google/cloud/spanner/timestamp.cc
google/cloud/spanner/internal/time_format.h
google/cloud/spanner/internal/time_format.cc
```

devbww:
Superseded by [google-cloud-cpp-common \#135](https://github.com/googleapis/google-cloud-cpp-common/issues/135).

PR #1110, opened 2019-11-22
bug: optimize Timestamp conversions to/from string

### LINKED ISSUE #1088: Optimize TimestampFromString()
`TimestampFromString()` appears to be the slowest of our "string -> data-type" conversion functions.  Beneath the hood it uses `std::get_time()`, which is overly general for our purpose, and could be avoided.

Aside: At the same time, we should probably avoid `std::put_time()` in `TimestampToString()`.

devbww:
Introduce versions of `internal::FormatTime()` and `internal::ParseTime()`
that do not take a format-specifier string. Instead they only format and
parse _"%Y-%m-%dT%H:%M:%S"_, allowing them to be very specific and fast.  In
particular this allows us to avoid `std::put_time()` and `std::get_time()`
(and we avoid `std::sscanf()` too). Use these to implemented the Spanner
`TimestampToString()` and `TimestampFromString()` conversion functions.
Fixes #1088.

Note: The new, format-specifier-less `internal::FormatTime()` produces a
4-character year when possible (unlike the _"%Y"_ format specifier).

Also introduce formatting and parsing of _"[-+]HH:MM"_ UTC offset fields.
While we will always produce a _"Z"_ offset in `TimestampToString()` for
Spanner, this will enable some refactoring of the time-formatting code
between all the Cloud C++ projects. See #145.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1110)

### DIFF
--- a/google/cloud/spanner/internal/time.cc
+++ b/google/cloud/spanner/internal/time.cc
@@ -14,7 +14,9 @@
 
 #include "google/cloud/spanner/internal/time.h"
 #include "google/cloud/spanner/internal/time_format.h"
+#include <array>
 #include <chrono>
+#include <cstdlib>
 #include <cstring>
 #include <ctime>
 #include <iomanip>
@@ -52,8 +54,7 @@ std::chrono::nanoseconds FromProto(google::protobuf::Duration const& proto) {
 
 namespace {
 
-// RFC3339 "date-time" prefix (no "time-secfrac" or "time-offset").
-constexpr auto kTimeFormat = "%Y-%m-%dT%H:%M:%S";
+constexpr auto kDigits = "0123456789";
 
 // Split a Timestamp into a seconds-since-epoch and a (>=0) subsecond.
 std::pair<std::chrono::seconds, Timestamp::duration> SplitTime(Timestamp ts) {
@@ -157,10 +158,10 @@ Timestamp FromProto(google::protobuf::Timestamp const& proto) {
 
 // TODO(#145): Reconcile this implementation with FormatRfc3339() in
 // google/cloud/internal/format_time_point.h in google-cloud-cpp.
-std::string TimestampToString(Timestamp ts) {
+std::string TimestampToString(Timestamp ts, std::chrono::minutes utc_offset) {
   std::ostringstream output;
   auto p = SplitTime(ts);
-  output << FormatTime(kTimeFormat, ZTime(p.first));
+  output << FormatTime(ZTime(p.first + utc_offset));
 
   if (auto ss = p.second.count()) {
     int width = 0;
@@ -176,7 +177,13 @@ std::string TimestampToString(Timestamp ts) {
     output << '.' << std::setfill('0') << std::setw(width) << ss;
   }
 
-  output << 'Z';
+  if (auto const minutes = utc_offset.count()) {
+    output << std::setfill('0') << std::internal << std::showpos
+           << std::setw(1 + 2) << (minutes / 60) << ':' << std::noshowpos
+           << std::setw(2) << std::abs(minutes % 60);
+  } else {
+    output << 'Z';
+  }
   return output.str();
 }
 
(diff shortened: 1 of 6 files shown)


devbww:
For the record, here is what I got from my first attempt to run the "read-timestamp" benchmark.

Before:
```
  UsingStub     experiment CpuTimePerRow
1     false read-timestamp           Inf
2      true read-timestamp      4.087792
```
After:
```
  UsingStub     experiment CpuTimePerRow
1     false read-timestamp     75.211959
2      true read-timestamp      4.110392
```
An infinite improvement.  :-)

google/cloud/spanner/internal/time.cc
```diff
@@ -185,43 +192,92 @@ std::string TimestampToString(Timestamp ts) {
...
+      case 'Z':  // Zulu time
+      case 'z': {
+        ++pos;
+        break;
+      }
+      case '+':  // [-+]HH:MM
+      case '-': {
+        // Parse colon-separated hours, minutes, but not (yet) seconds.
+        std::array<std::chrono::minutes::rep, 2> fields = {{0, 0}};
+        auto it = fields.begin();
```
devbww: > Should we refactor ParseInt() from time_format.cc and use it here?

I would say not.  The implementations are sufficiently simple and sufficiently distinct that trying to factor out commonality would, I think, be more trouble than it is worth.

google/cloud/spanner/internal/time_format.cc
```diff
@@ -34,6 +37,62 @@ namespace spanner {
...
+char const* ParseInt(char const* dp, int min, int max, int* vp) {
+  int value = 0;
+  bool neg = false;
+  if (*dp == '-') {
+    neg = true;
+    ++dp;
+  }
+  char const* const bp = dp;
+  constexpr int kMin = std::numeric_limits<int>::min();
+  while (char const* cp = std::strchr(kDigits, *dp)) {
```
devbww: > I am guessing this explicit look is here because std::from_chars() is unavailable to us, and std::strtol() is too slow? Consider a comment explaining why.

Added such a comment.

google/cloud/spanner/internal/time_format.cc
```diff
@@ -34,6 +37,62 @@ namespace spanner {
...
+
+// Note: tm.tm_year and tm.tm_mon are unadjusted (i.e., have true values).
+bool ValidDay(std::tm const& tm) {
+  static constexpr std::array<int, 1 + 12> kMonthDays = {
+      {-1, 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31}  // non leap year
+  };
+  if (tm.tm_mon == 2 && LeapYear(tm.tm_year)) {
+    return tm.tm_mday <= 29;
+  }
+  return tm.tm_mday <= kMonthDays[tm.tm_mon];
```
devbww: > Should we worry about out of range values for tm.tm_mon?

No ... we only pass a `tm` whose `tm_mon` and `tm_day` fields have already been range-checked to [1:12] and [1:31] respectively.

google/cloud/spanner/internal/time_format.cc
```diff
@@ -70,6 +169,38 @@ std::size_t ParseTime(char const* fmt, std::string const& s, std::tm* tm) {
 #endif
 }
 
+std::size_t ParseTime(std::string const& s, std::tm* tm) {
+  constexpr auto kYearMin = std::numeric_limits<int>::min() + 1900;
+  constexpr auto kYearMax = std::numeric_limits<int>::max();
+  std::tm tmp{};
+  char const* dp = ParseInt(s.c_str(), kYearMin, kYearMax, &tmp.tm_year);
+  if (dp != nullptr && *dp == '-') {
```
devbww: > nit: I prefer to return early instead of indenting

Yeah ... a possibility.  But in this particular case I believe I find indenting more readable.  Certainly having only one "success" and one "error" `return` helps a little.

google/cloud/spanner/internal/time_format_test.cc
```diff
@@ -27,13 +27,86 @@ constexpr auto kTimeFormat = "%Y-%m-%dT%H:%M:%S";
...
+#endif
+
+  tm.tm_year = 123 - 1900;
+  tm.tm_mon = 6 - 1;
+  tm.tm_mday = 21;
+  tm.tm_hour = 16;
+  tm.tm_min = 52;
+  tm.tm_sec = 22;
+#if defined(__APPLE__) || defined(_WIN32)
+  EXPECT_EQ("0123-06-21T16:52:22", FormatTime(kTimeFormat, tm));
```
devbww: > Do we know if Cloud Spanner accepts both formats?

I'll say it doesn't matter because, in the case of Spanner, we don't use the general form of `FormatTime()` (i.e., the one with a `fmt` string), so we're are not subject to the `"%Y"` vagaries.

devbww:
> When we take a dep on Abseil, which I think will be soon, I think we can delete ~all of this and offload the work to Abseil Time. Does that sound right?

Yes, I believe that would be possible, although there may be some specifics we're overlooking at the moment.  Certainly the parsing code looks much the same.

PR #1115, opened 2019-11-25
feat: parallelize benchmark smoke tests

coryan:
Run the smoke tests in parallel to keep the CI build times under
control.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1115)

### DIFF
--- a/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
+++ b/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
@@ -414,13 +414,13 @@ class ExperimentImpl {
   }
 
   bool UseStub(Config const& config) {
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
```diff
@@ -401,6 +413,36 @@ class ExperimentImpl {
                                  cs::MakeKeyBoundClosed(cs::Value(end)));
   }
 
+  bool UseStub(Config const& config) {
+    std::lock_guard<std::mutex> lk(mu_);
```
devbww: Move down to immediately before use of `generator_`?

Issue #914, opened 2019-10-01
Properly handle Session NOT_FOUND


mr-salty:
@coryan reported that one of his tests failed with the following error:

`Exception caught in HTTP handler: Permanent failure in RunTransactionWithPolicies: Session not found: <XXX> [NOT FOUND]`

We should handle this error internally and avoid exposing it to the user.  From https://cloud.google.com/spanner/docs/sessions

```
Attempts to use a deleted session result in NOT_FOUND. If you encounter this error, create and use a new session, add the new session to the cache, and remove the deleted session from the cache.
``` 

However, I'm not entirely sure how to accomplish this given that a `Transaction` is bound to a `Session`.  For RW transactions using `RunTransaction` we should be able to treat it similarly to an `ABORTED` and retry the whole transaction with a new session.

For RO transactions or non-idempotent mutations, it's less clear what we should do. The page linked-to above gives some general advice, but nothing specifically about this case.

mr-salty:
Carlos also noted that NOT_FOUND could apply to other parts of the request (database, table, etc.), so we may need to perform a `GetSession` to determine the actual issue. If it's a problem with the request itself, that should be returned to the user; `Session` related errors should be handled internally.

coryan:
Also, when we get this error we should remove the session from the pool.  Otherwise the next request to use that session will get the same error.

devbww:
Note that the message above ...
```
Permanent failure in RunTransactionWithPolicies: Session not found: <XXX> [NOT FOUND]
```
says `Session not found`, so you may be able to distinguish failures without resorting to `GetSession()`.

devjgm:
Note: the java library inspects the error message string to determine of the NOT_FOUND means "session not found". See here: https://github.com/googleapis/google-cloud-java/blob/2c109fc8a6462e6e0dc61b5503c2b64902f55383/google-cloud-clients/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L1192

PR #1138, opened 2019-12-12
feat: mark not-found sessions as bad

devjgm:
This fixes part of #914 

A followup PR will need to have the SessionPool look at this Session::bad() bit and remove it from the pool (or something).

Note: much of this was pair-programmed with @devbww

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1138)

### DIFF
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -183,12 +183,12 @@ StatusOr<CommitResult> Client::Commit(
     }
     if (internal::IsSessionNotFound(status)) {
       // Marks the session bad and creates a new Transaction for the next loop.
-      internal::Visit(
-          txn, [](internal::SessionHolder& s,
-                  google::spanner::v1::TransactionSelector&, std::int64_t) {
-            s->set_bad();
-            return true;
-          });
+      internal::Visit(txn, [](internal::SessionHolder& s,
+                              google::spanner::v1::TransactionSelector const&,
+                              std::int64_t) {
+        s->set_bad();
+        return true;
+      });
       txn = MakeReadWriteTransaction();
     } else {
       // Create a new transaction for the next loop, but share lock priority
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -19,6 +19,7 @@
 #include "google/cloud/spanner/internal/spanner_stub.h"
 #include "google/cloud/spanner/internal/status_utils.h"
 #include "google/cloud/spanner/retry_policy.h"
+#include "google/cloud/spanner/transaction.h"
 #include "google/cloud/log.h"
 #include <grpcpp/grpcpp.h>
 #include <thread>
@@ -191,7 +192,7 @@ StatusOr<CommitResult> Client::Commit(
       });
       txn = MakeReadWriteTransaction();
     } else {
-      // Create a new transaction for the next loop, but share lock priority
+      // Create a new transaction for the next loop, but reuse the session
       // so that we have a slightly better chance of avoiding another abort.
       txn = MakeReadWriteTransaction(std::move(txn));
     }
--- a/google/cloud/spanner/client.cc
+++ b/google/cloud/spanner/client.cc
@@ -17,6 +17,7 @@
 #include "google/cloud/spanner/internal/connection_impl.h"
 #include "google/cloud/spanner/internal/retry_loop.h"
 #include "google/cloud/spanner/internal/spanner_stub.h"
+#include "google/cloud/spanner/internal/status_utils.h"
 #include "google/cloud/spanner/retry_policy.h"
 #include "google/cloud/log.h"
 #include <grpcpp/grpcpp.h>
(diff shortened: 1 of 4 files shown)


google/cloud/spanner/client.cc
```diff
@@ -176,14 +177,25 @@ StatusOr<CommitResult> Client::Commit(
...
+      internal::Visit(txn, [](internal::SessionHolder& s,
+                              google::spanner::v1::TransactionSelector const&,
+                              std::int64_t) {
+        s->set_bad();
+        return true;
+      });
+      txn = MakeReadWriteTransaction();
+    } else {
+      // Create a new transaction for the next loop, but share lock priority
+      // so that we have a slightly better chance of avoiding another abort.
```
devbww: Now that we've mentioned "session" and used a `SessionHolder` above, we can probably amend this comment to something like:
```
// Create a new transaction for the next loop, but reuse the session
// so that we have a slightly better chance of avoiding another abort.
```
That makes things a little clearer.

google/cloud/spanner/internal/session.h
```diff
@@ -43,12 +46,16 @@ class Session {
 
   std::string const& session_name() const { return session_name_; }
 
+  void set_bad() { is_bad_.store(true, std::memory_order_release); }
```
devbww: I believe we can relax these ordering constraints to `memory_order_relaxed`.  There is no state (other than `is_bad_` itself) that the `set_bad()` caller communicates to the `is_bad()` caller, so memory ordering is irrelevant.

google/cloud/spanner/client_test.cc
```diff
@@ -770,6 +770,57 @@ TEST(ClientTest, CommitMutatorSessionAffinity) {
...
+      .WillOnce([&timestamp](Connection::CommitParams const& cp) {
+        EXPECT_THAT(cp.transaction, HasSession("session-3"));
+        return CommitResult{*timestamp};
+      });
+
+  int n = 0;
+  auto mutator = [&n](Transaction const& txn) -> StatusOr<Mutations> {
+    SetSessionName(txn, "session-" + std::to_string(++n));
+    if (n < 3) return Status(StatusCode::kNotFound, "Session not found");
+    return Mutations{MakeDeleteMutation("table", KeySet::All())};
```
devbww: `return Mutations{}` should be sufficient (and perhaps even a little more common for a mutator that binds a session name).

google/cloud/spanner/client_test.cc
```diff
@@ -770,6 +770,57 @@ TEST(ClientTest, CommitMutatorSessionAffinity) {
...
+  auto conn = std::make_shared<MockConnection>();
+  EXPECT_CALL(*conn, Commit(_))
+      .WillOnce([&timestamp](Connection::CommitParams const& cp) {
+        EXPECT_THAT(cp.transaction, HasSession("session-3"));
+        return CommitResult{*timestamp};
+      });
+
+  int n = 0;
+  auto mutator = [&n](Transaction const& txn) -> StatusOr<Mutations> {
+    SetSessionName(txn, "session-" + std::to_string(++n));
```
devbww: Should `EXPECT_THAT(txn, DoesNotHaveSession())` to check that we got an unbound transaction before binding the new session name.

google/cloud/spanner/client_test.cc
```diff
@@ -770,6 +770,57 @@ TEST(ClientTest, CommitMutatorSessionAffinity) {
...
+      })
+      .WillOnce([&timestamp](Connection::CommitParams const& cp) {
+        EXPECT_THAT(cp.transaction, HasSession("session-2"));
+        return CommitResult{*timestamp};
+      });
+
+  int n = 0;
+  auto mutator = [&n](Transaction const& txn) -> StatusOr<Mutations> {
+    SetSessionName(txn, "session-" + std::to_string(++n));
+    return Mutations{MakeDeleteMutation("table", KeySet::All())};
```
devbww: Same comments as previous mutator.

google/cloud/spanner/client.cc
```diff
@@ -176,14 +177,25 @@ StatusOr<CommitResult> Client::Commit(
...
+    // A transient failure (e.g., kAborted), so consider rerunning.
     if (!rerun_policy->OnFailure(status)) {
       return status;  // reruns exhausted
     }
+    if (internal::IsSessionNotFound(status)) {
+      // Marks the session bad and creates a new Transaction for the next loop.
+      internal::Visit(txn, [](internal::SessionHolder& s,
+                              google::spanner::v1::TransactionSelector const&,
+                              std::int64_t) {
+        s->set_bad();
```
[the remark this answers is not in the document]
devjgm: Actually, it was nullptr in the first draft of my test, which is what caused the segv I mentioned yesterday. However, I don't *think* that it can be nullptr in real code.

That said, I was tempted to -- and would still be happy to -- change this to `if (s) s->set_bad()` to guard against this case. Preference?

google/cloud/spanner/client.cc
```diff
@@ -176,14 +177,25 @@ StatusOr<CommitResult> Client::Commit(
...
       }
     }
-    // A transient failure (i.e., kAborted), so consider rerunning.
+    // A transient failure (e.g., kAborted), so consider rerunning.
     if (!rerun_policy->OnFailure(status)) {
       return status;  // reruns exhausted
     }
+    if (internal::IsSessionNotFound(status)) {
+      // Marks the session bad and creates a new Transaction for the next loop.
+      internal::Visit(txn, [](internal::SessionHolder& s,
```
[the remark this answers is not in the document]
devjgm: Done

google/cloud/spanner/client.cc
devjgm: Done.

google/cloud/spanner/internal/session.h
devjgm: Done.

google/cloud/spanner/client_test.cc
devjgm: Done

google/cloud/spanner/client_test.cc
devjgm: Done

google/cloud/spanner/client_test.cc
devjgm: Done

PR #1139, opened 2019-12-12
docs: use cmake-out directory and add how to obtain spanner

### LINKED ISSUE #1131: Spanner install instructions use "$HOME/project", which I think does not exist
Looking at the debian buster install instructions [here](https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/INSTALL.md#debian-10---buster), I see the use of strings like`cd $HOME/project`, but I see nowhere where that "project" directory gets created. From context I think it's supposed to contain the spanner source code, but I'm not certain and it's not clear from the doc.

We should clarify this.

FWIW, I found this while walking through the install instructions in a clean docker image and pasting in each command (as root). The following command didn't work and I think the lack of a "project" dir is why:

```
cd $HOME/project
cmake -H. -B/o
cmake --build /o -- -j "${NCPU:-4}"
cd /o
ctest -LE integration-tests --output-on-failure
cd $HOME/project
sudo cmake --build /o --target install
```

remyabel:
Fixes #1131

This regenerates the `INSTALL.md` and Dockerfiles to use the `cmake-out`
directory instead of `/o` like in common. This also adds a blurb at the
beginning on how to obtain spanner so that `$HOME/project` will exist.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1139)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -11,7 +11,7 @@ Or obtain the tarball release:
 ```bash
 mkdir -p $HOME/project
 wget -q https://github.com/googleapis/google-cloud-cpp-spanner/archive/v0.4.0.tar.gz
-tar -xf v0.4.0.tar.gz -C $HOME/project
+tar -xf v0.4.0.tar.gz -C $HOME/project --strip=1
 ```
 
 # Installing google-cloud-cpp-spanner
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,3 +1,19 @@
+# Obtaining google-cloud-cpp-spanner
+
+There are two primary ways of obtaining `google-cloud-cpp-spanner`. You can use git:
+
+```bash
+git clone git@github.com:googleapis/google-cloud-cpp-spanner.git $HOME/project
+```
+
+Or obtain the tarball release:
+
+```bash
+mkdir -p $HOME/project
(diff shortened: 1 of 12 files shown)


INSTALL.md
```diff
@@ -1,3 +1,19 @@
...
+```bash
+git clone git@github.com:googleapis/google-cloud-cpp-spanner.git $HOME/project
+```
+
+Or obtain the tarball release:
+
+```bash
+mkdir -p $HOME/project
+wget -q https://github.com/googleapis/google-cloud-cpp-spanner/archive/v0.4.0.tar.gz
+tar -xf v0.4.0.tar.gz -C $HOME/project
```
devjgm: I think this will result in a new subdirectory under `$HOME/project` named like `google-cloud-cpp-spanner-v0.4.0/` that then contains the files.

We could fix this using `tar -xf v0.4.0.tar.gz -C $HOME/project --strip 1`, which *I think* works on all the Linux flavors?

Or we could just extract into $HOME, then move the named dir to "project".

remyabel: Thanks. FYI, that makes you the [second Google employee](https://meta.stackexchange.com/questions/214729/hello-from-google-spam-or-no) to call me Remy. Feel free to call me Tom.

devjgm: Oops, my apologies. Not sure why I thought that, but thanks for correcting me, Tom.

PR #1143, opened 2019-12-13
feat: commit begins transaction if not already begun

### LINKED ISSUE #689: Resolve use of `single_use_transaction` in `Commit`
Currently, our `Commit` implementation (in `internal/connection_impl.cc`) handles a transaction that has not been started (i.e. `Commit()` was the first method called using that `Transaction`) by using the `single_use_transaction` field in `CommitRequest`. This makes the operation non-idempotent (see the comment at the end, from the proto definition).

That behavior seems questionable and surprising to me, and I think we should re-evaluate it. I think it's surprising enough on its own, if `Commit` is the only call you make inside of a `RunTransaction` functor.  it also means if you were calling `ExecuteSql` followed by `Commit` (again inside a `RunTransaction` functor) and remove the former call (or add one if it didn't already exist), you've changed the semantics of the `Commit` call substantially.

I propose our `CommitImpl` should explicitly call `BeginTransaction`, and then `Commit`, potentially retrying either of them, or re-running the entire transaction (meaning, those two operations). Also note that Session Pooling may eliminate this situation to begin with (the recommendation is to "prepare" read-write sessions by preemptively calling `BeginTransaction` on them).

If the user wants the non-idempotent `single_use_transaction` behavior, we should add another `Commit` overload that takes `ReadWriteOptions`. I assume this exists because there is some performance benefit to it, but it does not seem like the right default behavior.

See also #627 

---
comment from `CommitRequest::single_use_transaction`:

    // Execute mutations in a temporary transaction. Note that unlike
    // commit of a previously-started transaction, commit with a
    // temporary transaction is non-idempotent. That is, if the
    // `CommitRequest` is sent to Cloud Spanner more than once (for
    // instance, due to retries in the application, or in the
    // transport library), it is possible that the mutations are
    // executed more than once. If this is undesirable, use
    // [BeginTransaction][google.spanner.v1.Spanner.BeginTransaction] and
    // [Commit][google.spanner.v1.Spanner.Commit] instead.

devjgm:
Fixes #689

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1143)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -728,7 +728,7 @@ StatusOr<CommitResult> ConnectionImpl::CommitImpl(
   if (s.selector_case() != spanner_proto::TransactionSelector::kId) {
     spanner_proto::BeginTransactionRequest begin;
     begin.set_session(session->session_name());
-    *begin.mutable_options() = s.begin();
+    *begin.mutable_options() = s.has_begin() ? s.begin() : s.single_use();
     auto stub = session_pool_->GetStub(*session);
     auto response = internal::RetryLoop(
         retry_policy_prototype_->clone(), backoff_policy_prototype_->clone(),
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -16,6 +16,7 @@
 #include "google/cloud/spanner/internal/partial_result_set_resume.h"
 #include "google/cloud/spanner/internal/partial_result_set_source.h"
 #include "google/cloud/spanner/internal/retry_loop.h"
+#include "google/cloud/spanner/internal/status_utils.h"
 #include "google/cloud/spanner/internal/time.h"
 #include "google/cloud/spanner/query_partition.h"
 #include "google/cloud/spanner/read_partition.h"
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -1290,21 +1292,78 @@ TEST(ConnectionImplTest, CommitGetSession_Retry) {
...
-        EXPECT_TRUE(request.has_single_use_transaction());
+        EXPECT_EQ(txn.id(), request.transaction_id());
         return Status(StatusCode::kPermissionDenied, "uh-oh in Commit");
       });
   auto commit = conn->Commit({MakeReadWriteTransaction(), {}});
   EXPECT_EQ(StatusCode::kPermissionDenied, commit.status().code());
   EXPECT_THAT(commit.status().message(), HasSubstr("uh-oh in Commit"));
 }
 
+TEST(ConnectionImplTest, CommitBeginTransaction_Retry) {
```
devbww: Nit: Similar comment to before about underscores in test names.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -706,20 +707,33 @@ StatusOr<CommitResult> ConnectionImpl::CommitImpl(
...
-  } else if (s.has_begin()) {
-    *request.mutable_single_use_transaction() = s.begin();
-  } else {
-    request.set_transaction_id(s.id());
-    is_idempotent = true;
+
+  if (s.selector_case() != spanner_proto::TransactionSelector::kId) {
+    spanner_proto::BeginTransactionRequest begin;
+    begin.set_session(session->session_name());
+    *begin.mutable_options() = s.begin();
```
devbww: Even though it doesn't make sense to pass a single-use/read-only transaction to `Commit()`, it also doesn't make sense to pass any read-only transaction, yet we support the latter and not the former.  So, we might as well look at `s.has_single_use()` and use its `TransactionOptions`, rather than using the default `TransactionOptions`, and let the chips fall where they may.

google/cloud/spanner/internal/connection_impl_test.cc
```diff
@@ -1290,21 +1292,78 @@ TEST(ConnectionImplTest, CommitGetSession_Retry) {
...
+            return MakeSessionsResponse({"test-session-name"});
+          });
+  EXPECT_CALL(*mock, BeginTransaction(_, _))
+      .WillOnce(Return(Status(StatusCode::kNotFound, "Session not found")));
+  auto txn = MakeReadWriteTransaction();
+  auto commit = conn->Commit({txn, {}});
+  EXPECT_FALSE(commit.ok());
+  auto status = commit.status();
+  EXPECT_TRUE(IsSessionNotFound(status)) << status;
+  /* EXPECT_THAT(txn, HasBadSession()); */  // Uncomment after #1141
```
devbww: Reminder.

google/cloud/spanner/internal/connection_impl_test.cc
devjgm: Filed https://github.com/googleapis/google-cloud-cpp-spanner/issues/1144

google/cloud/spanner/internal/connection_impl.cc
devjgm: I wouldn't say that we "support" read-only transactions. Any caller who passes in a read-only transaction to `Commit()` will get an error. I believe that's true before and after this PR.

To call `Commit()` correctly, callers must pass in a read-write transaction. RW transactions have all their options in the "begin" field of the selector, so that's why I think we only need to copy that field from the selector into the `mutable_options` in the request.

I don't think we should write extra code to deal with things like single-use RO transactions when callers are not allowed to pass them, and if they did, it's guaranteed to fail.

devjgm: Discussed this offline. I changed the code to consider the case where the input selector was a single_use transaction as @devbww was suggesting. We know this will give an error, but it's what the user asked for.

google/cloud/spanner/internal/connection_impl_test.cc
devjgm: Done

PR #1148, opened 2019-12-13
refactor: add a factory method for `SessionPool`

mr-salty:
I've been bitten a few times now by needing to call `shared_from_this()`
in the constructor (which doesn't work because the `shared_ptr` that's
going to hold `this` is not yet constructed), so this splits out code that
might need to do that into `Initialize()` which is called by the factory.

Session creation doesn't have to do it right now but it will, and we'll
also start the background thread processing from there... just wanted to
split this out in the interest of smaller CLs.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1148)

### DIFF
--- a/google/cloud/spanner/internal/session_pool.cc
+++ b/google/cloud/spanner/internal/session_pool.cc
@@ -68,6 +68,9 @@ SessionPool::SessionPool(Database db,
 
 void SessionPool::Initialize() {
   // Eagerly initialize the pool with `min_sessions` sessions.
+  // TODO(#307) this was moved to `Initialize` in preparation of using
+  // `shared_from_this()` in the process of creating sessions, which cannot
+  // be done in the constructor.
   if (options_.min_sessions() > 0) {
     std::unique_lock<std::mutex> lk(mu_);
     int num_channels = static_cast<int>(channels_.size());
--- a/google/cloud/spanner/internal/session_pool.cc
+++ b/google/cloud/spanner/internal/session_pool.cc
@@ -29,6 +29,17 @@ namespace internal {
 
 namespace spanner_proto = ::google::spanner::v1;
 
+std::shared_ptr<SessionPool> MakeSessionPool(
+    Database db, std::vector<std::shared_ptr<SpannerStub>> stubs,
+    SessionPoolOptions options, std::unique_ptr<RetryPolicy> retry_policy,
+    std::unique_ptr<BackoffPolicy> backoff_policy) {
+  auto pool = std::make_shared<SessionPool>(
+      std::move(db), std::move(stubs), std::move(options),
+      std::move(retry_policy), std::move(backoff_policy));
+  pool->Initialize();
+  return pool;
+}
+
 SessionPool::SessionPool(Database db,
                          std::vector<std::shared_ptr<SpannerStub>> stubs,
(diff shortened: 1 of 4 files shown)


google/cloud/spanner/internal/session_pool.h
```diff
@@ -55,16 +55,21 @@ namespace internal {
 class SessionPool : public std::enable_shared_from_this<SessionPool> {
  public:
   /**
-   * Create a `SessionPool`.
+   * Construct a `SessionPool`.
    *
-   * The parameters allow the `SessionPool` to make remote calls needed to
-   * manage the pool, and to associate `Session`s with the stubs used to
-   * create them. `stubs` must not be empty.
+   * @warning callers should use the `MakeSessionPool()` factory method rather
```
devbww: Do you want to make the ctor private and befriend `MakeSessionPool()` so that this is required?

google/cloud/spanner/internal/session_pool.h
```diff
@@ -55,16 +55,21 @@ namespace internal {
...
+   * than calling the constructor and `Initialize()` directly.
    */
   SessionPool(Database db, std::vector<std::shared_ptr<SpannerStub>> stubs,
               SessionPoolOptions options,
               std::unique_ptr<RetryPolicy> retry_policy,
               std::unique_ptr<BackoffPolicy> backoff_policy);
+  /**
+   * If using the factory method is not possible, call `Initialize()` exactly
```
devbww: I guess the question behind my previous comment is, when might it not be possible to use the factory?

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -29,6 +29,17 @@ namespace internal {
...
 namespace spanner_proto = ::google::spanner::v1;
 
+std::shared_ptr<SessionPool> MakeSessionPool(
+    Database db, std::vector<std::shared_ptr<SpannerStub>> stubs,
+    SessionPoolOptions options, std::unique_ptr<RetryPolicy> retry_policy,
+    std::unique_ptr<BackoffPolicy> backoff_policy) {
+  auto pool = std::make_shared<SessionPool>(
+      std::move(db), std::move(stubs), std::move(options),
+      std::move(retry_policy), std::move(backoff_policy));
+  pool->Initialize();
```
devbww: Do we need a breadcrumb to remind us why this can't be in the ctor, given that the usual reason (possibility of failure) don't apply.

google/cloud/spanner/internal/session_pool.h
mr-salty: I did do that initially but it means I can't use `make_shared`, which I guess isn't a big deal, but I also recall a long discussion with Greg about this in a similar context, where he advocated keeping everything public, so that was enough to make me succumb.

google/cloud/spanner/internal/session_pool.h
mr-salty: It should always be possible to use the factory.

google/cloud/spanner/internal/session_pool.cc
mr-salty: Done.

PR #1155, opened 2019-12-16
feat: add build to check API breakage

### LINKED ISSUE #651: Setup builds to detect API breakage
Just like we do for the bigtable and storage libraries, we should have scripts to detect any API breakage.  Even if we do decide to break the API, we should know when we do so.

coryan:
Set API baseline, create the builds and scripts to check it.

Fixes #651 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1155)


ci/kokoro/docker/check-api.sh
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
+#
+# Copyright 2018 Google LLC
```
devbww: nit: s/2018/2019/

ci/kokoro/docker/dump-reports.sh
```diff
@@ -0,0 +1,48 @@
...
+# limitations under the License.
+
+set -eu
+
+if [[ -z "${PROJECT_ROOT+x}" ]]; then
+  readonly PROJECT_ROOT="$(cd "$(dirname "$0")/../../.."; pwd)"
+fi
+source "${PROJECT_ROOT}/ci/kokoro/define-docker-variables.sh"
+
+# If w3m is installed there is nothing to do.
```
devbww: s/is/is not/

ci/kokoro/docker/check-api.sh
coryan: Done.

ci/kokoro/docker/dump-reports.sh
coryan: Done.

devbww:
For what it's worth, I saw these messages flash by.
```
Detect public symbols
ERROR: requires Universal Ctags to work properly
Reading debug-info
WARNING: the object should be compiled with -Og option for better analysis
```

PR #1161, opened 2019-12-19
feat: log responses to streaming APIs

### LINKED ISSUE #1160: Log the responses for streaming RPCs
We should log RPC responses for `SpannerStub::ExecuteStreamingSql()` and for `SpannerStub::StreamingSql()`.

coryan:
If the GOOGLE_CLOUD_CPP_ENABLE_TRACING environment variable contains
`rpc-streams` we enable logging for the streaming APIs.

Refactored the mocks for internal::PartialResultSetReader to a single
file.

Update the API baseline, all changes are in the `internal` namespace,
therefore this is not a breaking change.

Fixes #1160

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1161)


google/cloud/spanner/internal/logging_result_set_reader_test.cc
```diff
@@ -0,0 +1,114 @@
...
+  auto result = reader.Read();
+  ASSERT_TRUE(result.has_value());
+  EXPECT_EQ("test-token", result->resume_token());
+
+  HasLogLineWith("Read");
+  HasLogLineWith("test-token");
+
+  result = reader.Read();
+  ASSERT_FALSE(result.has_value());
+  HasLogLineWith("(optional-with-no-value)");
```
devbww: Something to consider for the future: Should we have some way to make the expectation only apply to log lines accumulated since the last expectation?

coryan: Done.

PR #1162, opened 2019-12-19
feat: optimistic concurrency control loop for IAM

### LINKED ISSUE #613: Consider a wrapper for optimistic concurrency control.
The `SetIamPolicy()` functions use OCC ([Optimistic Concurrency Control](https://en.wikipedia.org/wiki/Optimistic_concurrency_control)) to change the IAM policies.

Typically the application is expected to write a loop like this:

```
while (!given_up()) {
  auto current = client.GetIamPolicy(...);
  if (!current) continue; // try again, but we can discuss that...
  // ... make some changes to "current" ...
  auto updated = client.SetIamPolicy(..., *std::move(current));
  if (updated) return updated; // yay!
  if (!updated for OCC failure) continue; // try again
  if (!updated for some permanent error) return updated.status();
}
```

Feels like we could wrap that in a function, something like:

```
// Run the OCC loop calling mutator each time, returning an unset optional indicates that
// no changes are needed.
StatusOr<Policy> ChangeIamPolicy(std::function<optional<Policy>(Policy)> mutator);
```

coryan:
Adds functions to DatabaseAdminClient and InstanceAdminClient that run a
optimistic concurrency control (OCC) loop for IAM updates. That makes
the customer code a bit easier to write and less error prone. This is
analogous to the Commit() loops for the spanner::Client.

This fixes #613 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1162)

### DIFF
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -206,7 +206,8 @@ add_library(
     value.h
     version.cc
     version.h
-    version_info.h)
+    version_info.h
+    iam_updater.h)
 target_include_directories(
     spanner_client
     PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -117,6 +117,7 @@ add_library(
     database_admin_connection.h
     date.cc
     date.h
+    iam_updater.h
     instance.cc
     instance.h
     instance_admin_client.cc
@@ -206,8 +207,7 @@ add_library(
     value.h
     version.cc
     version.h
-    version_info.h
-    iam_updater.h)
+    version_info.h)
 target_include_directories(
     spanner_client
     PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
--- a/google/cloud/spanner/database_admin_client.cc
+++ b/google/cloud/spanner/database_admin_client.cc
@@ -90,8 +90,7 @@ StatusOr<google::iam::v1::Policy> DatabaseAdminClient::SetIamPolicy(
 
   Status last_status =
       Status(StatusCode::kUnknown, "Exhausted before first call");
-  for (; !rerun_policy->IsExhausted();
-       std::this_thread::sleep_for(backoff_policy->OnCompletion())) {
+  do {
     auto current_policy = GetIamPolicy(db);
     if (!current_policy) {
       last_status = std::move(current_policy).status();
@@ -108,10 +107,9 @@ StatusOr<google::iam::v1::Policy> DatabaseAdminClient::SetIamPolicy(
       }
       last_status = std::move(result).status();
     }
-    if (!rerun_policy->OnFailure(last_status)) {
-      return last_status;
-    }
-  }
+    if (!rerun_policy->OnFailure(last_status)) break;
+    std::this_thread::sleep_for(backoff_policy->OnCompletion());
+  } while (!rerun_policy->IsExhausted());
   return last_status;
 }
 
--- a/google/cloud/spanner/database_admin_client.cc
+++ b/google/cloud/spanner/database_admin_client.cc
@@ -64,6 +64,57 @@ StatusOr<google::iam::v1::Policy> DatabaseAdminClient::SetIamPolicy(
   return conn_->SetIamPolicy({std::move(db), std::move(policy)});
 }
 
+StatusOr<google::iam::v1::Policy> DatabaseAdminClient::SetIamPolicy(
+    Database const& db, IamUpdater const& updater) {
+  auto const rerun_maximum_duration = std::chrono::minutes(15);
+  auto default_rerun_policy =
+      LimitedTimeTransactionRerunPolicy(rerun_maximum_duration).clone();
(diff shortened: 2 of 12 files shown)


google/cloud/spanner/database_admin_client.cc
```diff
@@ -64,6 +64,57 @@ StatusOr<google::iam::v1::Policy> DatabaseAdminClient::SetIamPolicy(
...
+        return current_policy;
+      }
+      desired->set_etag(std::move(etag));
+      auto result = SetIamPolicy(db, *std::move(desired));
+      if (RerunnablePolicy::IsOk(result.status())) {
+        return result;
+      }
+      last_status = std::move(result).status();
+    }
+    if (!rerun_policy->OnFailure(last_status)) {
```
devbww: Nit: It is a little funny that `!OnFailure(s)` means `IsPermanentFailure(s) || IsExhausted()`, so, if we continue, we know we saw `!IsExhausted()`.  Yet, the very next thing we test is `!IsExhausted()`.  For a "limited count" policy this is redundant.  For a "limited time" policy it means we're immediately sampling the clock again, which feels redundant/racy.

We could address that, but the bigger question might be whether any rerun policy should be able to trigger an "Exhausted before first call".  After all, it isn't really a "rerun" if it was never a "run".

google/cloud/spanner/instance_admin_client.cc
```diff
@@ -67,6 +67,57 @@ StatusOr<google::iam::v1::Policy> InstanceAdminClient::SetIamPolicy(
...
+        return current_policy;
+      }
+      desired->set_etag(std::move(etag));
+      auto result = SetIamPolicy(in, *std::move(desired));
+      if (RerunnablePolicy::IsOk(result.status())) {
+        return result;
+      }
+      last_status = std::move(result).status();
+    }
+    if (!rerun_policy->OnFailure(last_status)) {
```
devbww: Ditto.

google/cloud/spanner/samples/samples.cc
```diff
@@ -266,42 +266,42 @@ void AddDatabaseReader(google::cloud::spanner::InstanceAdminClient client,
...
+
+        auto member_pos = std::find(binding.members().begin(),
+                                    binding.members().end(), new_reader);
+        if (member_pos != binding.members().end()) {
+          std::cout << "The entity " << new_reader
+                    << " is already a database reader:\n"
+                    << current.DebugString();
+          return {};
+        }
+        if (member_pos == binding.members().end()) {
```
devbww: Isn't this `if (true)`?

coryan: Fixed.

google/cloud/spanner/database_admin_client.cc
coryan: I think it is funny too.  WDTY of the do/while loop instead?

google/cloud/spanner/instance_admin_client.cc
coryan: Changed this too.

google/cloud/spanner/database_admin_client.h
```diff
@@ -244,6 +242,50 @@ class DatabaseAdminClient {
   StatusOr<google::iam::v1::Policy> SetIamPolicy(
       Database db, google::iam::v1::Policy policy);
 
+  /**
+   * Updates the IAM policy for an instance using an optimistic concurrency
+   * control loop.
+   *
+   * This function repeatedly reads the current IAM policy in @p in, and then
```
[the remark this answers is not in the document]
coryan: Fixed.

google/cloud/spanner/CMakeLists.txt
```diff
@@ -206,7 +206,8 @@ add_library(
     value.h
     version.cc
     version.h
-    version_info.h)
+    version_info.h
+    iam_updater.h)
```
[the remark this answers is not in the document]
coryan: Fixed.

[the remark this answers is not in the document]
coryan: And we should have cmake-format do this for us, I will check on that later.

google/cloud/spanner/database_admin_client.cc
devbww: I like it better.  But things like ...
```
  do {
   ...
    if (!rerun_policy->OnFailure(last_status)) break;
    std::this_thread::sleep_for(backoff_policy->OnCompletion());
  } while (!rerun_policy->IsExhausted());
```
reinforce the idea that the rerun and backoff policies should be unified.  It is no use sleeping beyond the deadline, for example.

The main feature would be that `rerun_policy->OnFailure(status)` would either return `false` _immediately_ on a permanent error, or `false` _when_ the count/deadline has been exceeded, or `true` _when_ the next attempt should be made.

But that is something for another day.

google/cloud/spanner/instance_admin_client.cc
```diff
@@ -93,8 +93,7 @@ StatusOr<google::iam::v1::Policy> InstanceAdminClient::SetIamPolicy(
 
   Status last_status =
       Status(StatusCode::kUnknown, "Exhausted before first call");
```
devbww: It looks like this doesn't need such specific initialization now.

PR #1166, opened 2020-01-06
fix!: support full TIMESTAMP range in g::c::spanner::Timestamp

### LINKED ISSUE #1098: Timestamp type does not support full range documented by Spanner
Spanner [documents](https://cloud.google.com/spanner/docs/data-types#timestamp-type) the range for the `TIMESTAMP` type as

> 0001-01-01 00:00:00 to 9999-12-31 23:59:59.999999999 UTC.

The C++ type we use for `Timestamp` is 
https://github.com/googleapis/google-cloud-cpp-spanner/blob/0945decf6fe09daad3b7a1eac3e11b82fbab806c/google/cloud/spanner/timestamp.h#L37-L38

Which is 64-bits of nanoseconds, which will give us a range of +/- ~292 years.

We should figure out how we want to handle this. We probably want to support the full range of Spanner timestamps.

devbww:
Move `google::cloud::spanner::Timestamp` from an alias for
```
  std::chrono::time_point<std::chrono::system_clock,
                          std::chrono::nanoseconds>
```
to a custom class that supports the full range of Spanner TIMESTAMP
values: 0001-01-01T00:00:00Z to 9999-12-31T23:59:59.999999999Z.

Conversions are provided from/to strings, `google::protobuf::Timestamp`,
and `std::chrono::time_point<std::chrono::system_clock, D>` types
using standard `std::chrono::duration` types.

Conversion from raw "seconds/nanoseconds since Unix epoch" counts is
also provided.

Fixes #1098.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1166)


google/cloud/spanner/timestamp.h
```diff
@@ -16,26 +16,137 @@
...
+  ///@{
+  static StatusOr<Timestamp> FromString(std::string const&);
+  std::string ToString() const;
+  ///@}
+
+  /// Conversion from/to `google::protobuf::Timestamp`. These conversions never
+  /// fail, but may accept/produce protobufs outside their documented range.
+  ///@{
+  static Timestamp FromProto(google::protobuf::Timestamp const&);
+  google::protobuf::Timestamp ToProto() const;
```
devjgm: In other files, our `FromProto` and `ToProto` functions are not part of any public API. Instead, the're defined as non-members in an internal namespace. Any reason not to do the same in this file?

google/cloud/spanner/timestamp.h
```diff
@@ -16,26 +16,137 @@
...
+    if (!count) return count.status();
+    return UnixEpoch<Duration>() +
+           Duration(static_cast<typename Duration::rep>(*count));
+  }
+  ///@}
+
+  /// Build a `Timestamp` from raw counts of seconds/nanoseconds since the Unix
+  /// epoch (minus leap seconds), like a C++17 `std::timespec` but out-of-range
+  /// `nsec` values are normalized. Fails if the time cannot be represented.
+  static StatusOr<Timestamp> FromCounts(std::intmax_t sec, std::intmax_t nsec);
```
devjgm: I'd imagine users should never need this function. Could we move this to a non-member in an internal namespace?

google/cloud/spanner/timestamp.h
```diff
@@ -16,26 +16,137 @@
...
+  ///@{
+  static Timestamp FromProto(google::protobuf::Timestamp const&);
+  google::protobuf::Timestamp ToProto() const;
+  ///@}
+
+  /// Conversion from/to `std::chrono::time_point` on the system clock.
+  /// May produce out-of-range errors in either direction, depending on the
+  /// properties of `Duration` and the `std::chrono::system_clock` epoch.
+  ///@{
+  template <class Duration>
```
devjgm: Will this allow any `Duration`? If not, if an invalid `Duration` is specified, will an error `Status` be returned? 

If not, we could write a small type trait to verify that the given Durations are valid, then we could static_assert() in the function body.

google/cloud/spanner/timestamp.h
```diff
@@ -16,26 +16,137 @@
...
+  static StatusOr<Timestamp> FromChrono(sys_time<Duration> const& tp) {
+    return FromRatio((tp - UnixEpoch<Duration>()).count(),
+                     Duration::period::num, Duration::period::den);
+  }
+  template <class Duration>
+  StatusOr<sys_time<Duration>> ToChrono() const {
```
devjgm: Is this function tested?

google/cloud/spanner/integration_tests/data_types_integration_test.cc
```diff
@@ -173,22 +173,22 @@ TEST_F(DataTypeIntegrationTest, WriteReadBytes) {
...
-  // auto min = internal::TimestampFromString("0001-01-01T00:00:00Z");
-  // auto max = internal::TimestampFromString("9999-12-31T23:59:59.999999999Z");
-
-  // ASSERT_STATUS_OK(min);
-  // ASSERT_STATUS_OK(max);
+  auto min = Timestamp::FromString("0001-01-01T00:00:00Z");
+  ASSERT_STATUS_OK(min);
+  auto max = Timestamp::FromString("9999-12-31T23:59:59.999999999Z");
+  ASSERT_STATUS_OK(max);
+  auto now = Timestamp::FromChrono(std::chrono::system_clock::now());
```
devjgm: I think this would be the first type in this Spanner API that has a factory function like `Foo::FromBlah()`. In other cases we would make this a non-member factory named `MakeTimestamp()`. What do you think about changing this API to match?

You could also change `Timestamp::ToChrono<T>()` to be `T Timestamp::get<T>()`, which would match what we do in other places, like `Value`, and `Bytes`.

Not that one way is strictly better/worse than the other; my question is simply about the consistency between the various classes in this API.

Thoughts?

google/cloud/spanner/timestamp.h
```diff
@@ -16,26 +16,137 @@
...
+  template <class Duration>
+  using sys_time = std::chrono::time_point<std::chrono::system_clock, Duration>;
+
+ public:
+  /// Default construction yields the Unix epoch.
+  Timestamp() : Timestamp(0, 0) {}
+
+  /// Conversion from/to RFC3339 string.
+  ///@{
+  static StatusOr<Timestamp> FromString(std::string const&);
```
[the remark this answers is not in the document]
devbww: Done.

google/cloud/spanner/timestamp.h
devbww: I think the main reason is that `google::protobuf::Timestamp` is not a spanner-internal type.  That is, I can imagine that the user might be dealing with that type as much as they might use `std::chrono`.

Secondarily, if `ToProto()` is a non-member then there must exist an accessor for the sec/nsec fields.  My thought was that `ToProto()` is that accessor.

google/cloud/spanner/timestamp.h
devbww: I'm not so sure that users will never need it.  It is hard to predict how they might originate their time values.  We certainly make good use of it in our tests.

That said, it is a little strange that it doesn't have a "To" partner.  If I follow through on the "`ToProto()` is the general accessor" idea, then perhaps `FromProto()` is the general factory instead.  But my problem with that is that I wanted a normalizing sec/nsec factory, and flagrantly violating the range documentation of `google::protobuf::Timestamp::nanos` didn't feel right.

google/cloud/spanner/timestamp.h
devbww: Yes, it does allow any `Duration`.

That said, it is difficult to convert between sec/nsec and a `time_point` that counts in, say, 987654321/1234567890 of a second without risking intermediate overflow or precision loss.  The current implementation only claims to avoid (most of) those problems for the standard durations (see the comments on `FromRatio()` and `ToRatio()`).

google/cloud/spanner/timestamp.h
devbww: I say yes.  See the `Timestamp.ToChrono` test.

google/cloud/spanner/integration_tests/data_types_integration_test.cc
devbww: I was mostly swayed by the internal consistency of closely-declared "static 'From' factory" and "'To'  converter" pairs for each of string/proto/chrono than by any consistency with other classes.

That said, I'm happy to discuss.

PR #1167, opened 2020-01-06
chore: upgrade grpc to v1.26.0

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1167)

### DIFF
--- a/bazel/google_cloud_cpp_spanner_deps.bzl
+++ b/bazel/google_cloud_cpp_spanner_deps.bzl
@@ -77,12 +77,12 @@ def google_cloud_cpp_spanner_deps():
     if "com_github_grpc_grpc" not in native.existing_rules():
         http_archive(
             name = "com_github_grpc_grpc",
-            strip_prefix = "grpc-1.24.2",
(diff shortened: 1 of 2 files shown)


scotthart:
> Do we understand why the `vptr` change is not needed here? Maybe because we do not define any gRPC servers?
> 
In the bigtable library we use/pass arguments of type `grpc::ServerContext*` in multiple places. This class directly inherits from `grpc_impl::ServerContextBase`, which had one method become virtual in the new version, v1.26.0. The ubsan errors observed from the bigtable repo were:
```
undefined reference to 'typeinfo for grpc_impl::ServerContextBase'
```
In spanner, we only use `grpc::GenericServerContext`, once, as part of our testing. `grpc::GenericServerContext` inherits from `grpc::ServerContext`

coryan:
FWIW, the 1.26.x branch allegedly has fixes for the problems you are seeing.

devbww:
This looks like it is probably obsolete after the recent merge of #1232.

PR #1170, opened 2020-01-08
fix: avoid deadlock waiting for Session allocation

### LINKED ISSUE #1000: spanner::Client deadlocks with about 100 threads
Running this function, with the same `spanner::Client` on about 100 threads deadlocks:

```C++
int RunTask(Config const& config, cloud_spanner::Client client,
            RandomKeyGenerator const& key_generator) {
  int count = 0;
  std::string value(1024, 'A');
  for (auto start = std::chrono::steady_clock::now(),
            deadline = start + config.iteration_duration;
       start < deadline; start = std::chrono::steady_clock::now()) {
    auto key = key_generator();
    auto m = cloud_spanner::MakeInsertOrUpdateMutation(
        "KeyValue", {"Key", "Value"}, key, value);
    auto result = client.Commit([&m](cloud_spanner::Transaction const&) {
      return cloud_spanner::Mutations{m};
    });
    ++count;
  }
  return count;
}
```

This may be a problem with the client library code, with gRPC, with the service, or maybe it is not a problem and my code is just broken. In any case, we should investigate before cutting Beta or GA.

mr-salty:
Track the number of threads waiting for a `Session` to ensure we always
notify properly; the prior implementation had race conditions that
could lead to a thread waiting forever even when a `Session` was
available (see #1000 for details).

I also removed the workaround for this bug in the benchmark. With
the workaround removed, I ran the benchmark using the flags below,
both with and without the `SessionPool` fix. Without the fix it
hangs reliably, with the fix it never hangs.

```
.build/google/cloud/spanner/benchmarks/single_row_throughput_benchmark \
  --project=${GOOGLE_CLOUD_PROJECT} \
  --instance=${GOOGLE_CLOUD_CPP_SPANNER_INSTANCE} \
  --table-size=1000 \
  --minimum-clients=1 --maximum-clients=1 \
  --minimum-threads=1000 --maximum-threads=1000 \
  --iteration-duration=2 --samples=100 \
  --experiment=read
```

fixes #1000

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1170)

### DIFF
--- a/google/cloud/spanner/internal/session_pool.cc
+++ b/google/cloud/spanner/internal/session_pool.cc
@@ -110,9 +110,11 @@ StatusOr<SessionHolder> SessionPool::Allocate(bool dissociate_from_pool) {
       if (options_.action_on_exhaustion() == ActionOnExhaustion::FAIL) {
         return Status(StatusCode::kResourceExhausted, "session pool exhausted");
       }
+      ++num_waiting_for_session_;
       cond_.wait(lk, [this] {
         return !sessions_.empty() || total_sessions_ < max_pool_size_;
       });
+      --num_waiting_for_session_;
       continue;
     }
 
@@ -123,8 +125,10 @@ StatusOr<SessionHolder> SessionPool::Allocate(bool dissociate_from_pool) {
     // simulaneous calls if additional sessions are needed. We can also use the
     // number of waiters in the `sessions_to_create` calculation below.
     if (create_in_progress_) {
+      ++num_waiting_for_session_;
       cond_.wait(lk,
                  [this] { return !sessions_.empty() || !create_in_progress_; });
+      --num_waiting_for_session_;
       continue;
     }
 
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/internal/session_pool.cc
```diff
@@ -166,10 +165,8 @@ void SessionPool::Release(std::unique_ptr<Session> session) {
     --total_sessions_;
     return;
   }
-  bool notify = sessions_.empty();
   sessions_.push_back(std::move(session));
-  // If sessions_ was empty, wake up someone who was waiting for a session.
-  if (notify) {
+  if (num_waiting_for_session_ > 0) {
```
devbww: I'm fine with this ... while it does not eliminate the possibility of useless (and purportedly expensive) notifications, it does reduce them.

And all the better if you can use `num_waiting_for_session_` to make `CreateSessions()` a little smarter.

google/cloud/spanner/internal/session_pool.h
```diff
@@ -100,6 +100,15 @@ class SessionPool : public std::enable_shared_from_this<SessionPool> {
   // Release session back to the pool.
   void Release(std::unique_ptr<Session> session);
 
+  // Called when a thread needs to wait for a `Session` to become available.
+  // @p specifies the condition to wait for.
+  template <typename Predicate>
+  void Wait(std::unique_lock<std::mutex>& lk, Predicate&& p) {
```
devbww: This could be a non-member, implementation-file helper if you also passed it a reference to the count.

PR #1174, opened 2020-01-08
feat: allow re-using a database across benchmark runs

mr-salty:
If the user specifies a non-empty `--database <database_name>` on the command
line, it will be re-used across runs. The database will be created and
populated if it doesn't already exist (i.e. on the first run) but it
will not be dropped at the end of the run as it normally would be.  This
saves a lot of setup time if you're running the benchmark repeatedly.

Otherwise, the behavior is unchanged.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1174)


mr-salty:
@coryan I debated introducing a`--reuse_database` flag and having the program use a non-random name, I don't have too much of a preference either way, LMK if you feel differently.

mr-salty:
yeah, it really helped me immensely when debugging that deadlock issue the other night... well, luckily (for my motivation to do this) it wasn't until later that I realized I could have used a much smaller `--table_size` to reduce the overhead ;)  but this is still better regardless.

PR #1189, opened 2020-01-10
feat: a small script to plot benchmark results

coryan:
Generate pretty (maybe) plots from the benchmark data, for example:

foo

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1189)


google/cloud/spanner/benchmarks/single_row_throughput_plots.py
```diff
@@ -0,0 +1,54 @@
...
+# %%
+import argparse
+import numpy as np
+import pandas as pd
+import plotnine as p9
+
+
+# %%
+def load_benchmark_output(file):
+    """Loads the output from the storage_throughput_vs_cpu_benchmark."""
```
devbww: Nit: "storage_throughput_vs_cpu_benchmark" looks like a google-cloud-cpp remnant.  Similarly in the file comment above.

coryan: Fixed in both places.

coryan:
PTAL

PR #1224, opened 2020-01-30
chore: remove underscore suffix from include guards

devjgm:
Part of #355 This is needed so that we can enable clang-tidy on our headers. 

After this PR, I'll update the headers in our -common and -cpp repos, and I'll update our [style guide](https://github.com/googleapis/google-cloud-cpp/blob/master/doc/cpp-style-guide.md).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1224)


coryan:
The build breakage is a known flake as I recall, you probably want to rebuild it just in case.

PR #1228, opened 2020-01-30
doc: document how to run samples using Bazel

### LINKED ISSUE #1133: Need easy instructions for running "samples" commands
In order to have a nice "getting started" page for C++ similar to [this one for Java](https://cloud.google.com/spanner/docs/getting-started/java/), we need some reasonably simple instructions for users can run the individual sample commands.

Some possible options:

1. Direct users to the INSTALL.md file and have them install the deps and compile all the spanner code. This is likely a lot of work for a "getting started" doc.

2. Use bazel. Bazel is pretty simple since it builds all the deps automatically. The downside is that most users do not use bazel, so they'd first need to go install bazel.

3. Use a Dockerfile where we can automate this. The downside is that users need to install docker if they don't already have it.

4. Use vcpkg. Users would need to install vcpkg if they don't already have it.

5. We host/provide a few *statically linked, precompiled* "samples" binaries. We'd need one for Windows, Mac, and Linux (at least). Users could then directly run these pre-compiled binaries when going through the "samples" exercises and they'd have nothing to compile themselves.

6. Some other script of some form??

7. .... ???

coryan:
Update the README.md generate to include instructions to build the
and run the examples using Bazel.

This PR changes the target name from

`//google/cloud/spanner/samples:spanner_client_samples`

to 

`//google/cloud/spanner/samples:samples`

The latter seems easier to explain.

This fixes #1133 

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1228)

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ bazel --output_user_root=C:\b build //google/cloud/spanner/samples:samples
 ```
 
 On Windows and macOS gRPC [requires][grpc-roots-pem-bug] an environment variable
-to find the to manually set the root of trust for SSL. On macOS use:
+to find the root of trust for SSL. On macOS use:
 
 ```console
 wget -q https://raw.githubusercontent.com/grpc/grpc/master/etc/roots.pem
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -131,13 +131,77 @@ in the [release notes](README.md#release-notes) below.
...
+
+> You must provide this option in **all** Bazel invocations shown below.
+
+```console
+mkdir C:\b
+bazel --output_user_root=C:\b build //google/cloud/spanner/samples:samples
+```
+
+On Windows and macOS gRPC [requires][grpc-roots-pem-bug] an environment variable
+to find the to manually set the root of trust for SSL. On macOS use:
```
devjgm: phrasing

coryan: Argh, that was ugly.  Thanks.

PR #1232, opened 2020-02-03
ci: use gRPC-1.26.x to workaround a known bug

devbww:
Workaround grpc/grpc#21280 by compiling against a specific commit
(without a tag) of the v1.26.x gRPC branch.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1232)


devbww:
It looks like you may have approved the google_cloud_cpp version twice, instead of hitting this one?

PR #1233, opened 2020-02-03
fix!: avoid single-arg forwarding c'tor

devjgm:
Single-argument forwarding constructors can hide implicitly generated
copy/move constructors. Even if this issue is avoided with the proper
use of SFINAE, clang-tidy may still complain (incorrectly, I think).
http://clang.llvm.org/extra/clang-tidy/checks/bugprone-forwarding-reference-overload.html

To fix this I renamed the class `StreamOf` -> `TupleStream`. I then
changed its constructor to accept two iterators instead of a single
range. Then I added a `StreamOf<T>()` non-member factory function. This
should make it so our existing uses all continue to work.

So, this really shouldn't be much of an API break.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1233)

### DIFF
--- a/google/cloud/spanner/row.h
+++ b/google/cloud/spanner/row.h
@@ -372,10 +372,12 @@ class TupleStreamIterator {
 };
 
 /**
- * A `StreamOf<Tuple>` defines a range that parses `Tuple` objects from the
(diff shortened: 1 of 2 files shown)


devjgm:
Related to #355

PR #1237, opened 2020-02-04
chore: change Transaction's Visit to accept const-ref

devjgm:
Without this change, clang-tidy's performance-unnecessary-value-param
checker warns that we're unnecessarily passing values around. This
change quiets the warning, and I think makes the API slighly simpler to
use since it doesn't require the caller to std::move in order to avoid
the unnecessary increment/decrement in Transaction::impl_.

Related to #355

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1237)

### DIFF
--- a/google/cloud/spanner/internal/connection_impl.cc
+++ b/google/cloud/spanner/internal/connection_impl.cc
@@ -104,8 +104,9 @@ ConnectionImpl::ConnectionImpl(Database db,
       tracing_options_(options.tracing_options()) {}
 
 RowStream ConnectionImpl::Read(ReadParams params) {
+  auto txn = params.transaction;
   return internal::Visit(
-      params.transaction,
+      txn,
       [this, &params](SessionHolder& session,
                       spanner_proto::TransactionSelector& s, std::int64_t) {
         return ReadImpl(session, s, std::move(params));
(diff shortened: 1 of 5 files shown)


devbww:
Appears to have introduced a use-after-free error.

google/cloud/spanner/internal/connection_impl.cc
```diff
@@ -197,18 +192,17 @@ StatusOr<std::vector<QueryPartition>> ConnectionImpl::PartitionQuery(
...
+      });
 }
 
 StatusOr<CommitResult> ConnectionImpl::Commit(CommitParams params) {
   return internal::Visit(
-      std::move(params.transaction),
+      params.transaction,
       [this, &params](SessionHolder& session,
                       spanner_proto::TransactionSelector& s, std::int64_t) {
         return this->CommitImpl(session, s, std::move(params));
```
devbww: If we're not going to move the transaction out of `params`, then we can't allow `params` to die when `CommitImpl()` completes.  And the same is true for the other similar patterns.

Indeed, moving `params` into `CommitImpl()` seems wrong given that the functor may be called multiple times.

devjgm: Hmm, I don't think I understand the Transaction implementation enough to see the problem right now. It seems like this change from taking the Transaction by value to const-ref should have been safe, but the asan build disagrees with me. Maybe tomorrow or something we can talk about this at a whiteboard. Thanks.

devjgm: OK, I may understand, and I just pushed a commit that (I think) fixes this. In cases where Visit's functor "moves" from the transaction (usually via moving the params), the code now first makes a copy of the transaction to pass as the first argument to Visit.

WDYT about this fix? Is there some other fix that you'd prefer?

devbww: I would still claim that any `mutator` that consumes a value during a call (be it captured by value or reference) is wrong.

> the code now first makes a copy of the transaction

Which kind of defeats the "don't pass by value because it copies" warning.

devjgm: > I would still claim that any mutator that consumes a value during a call (be it captured by value or reference) is wrong.

To clarify, you're saying that a Visit functor that consumes any value is wrong? If so, is that unrelated to the change in this PR to pass Transaction by const-ref?

> Which kind of defeats the "don't pass by value because it copies" warning.

It's debatable, but I'd say no. A copy is only needed if the mutator consumes the Transaction. Those are both things that the caller controls. If the caller is doing something weird, then they may need to make a copy. 

---

How would you like to proceed? The options I see are:

1. Drop this PR and tell clang-tidy to ignore the `-performance-unnecessary-value-param` warnings in our `.clang-tidy` config.

2. Merge this PR.

3. Make some other change to address this warning.

I'm fine with any of those options.

devjgm: Note: another option is to silence this clang-tidy warning only for the one `internal::Visit` function, as in this PR: https://github.com/googleapis/google-cloud-cpp-spanner/pull/1239 I'm fine with this option too. If we approve that PR, then we can drop this one.

devbww: I would say it is a "good thing" that `Visit()` takes a new reference to the `Transaction` it is about to call a member function on when that member function can invoke user code that might drop the original reference to the `Transaction`.  `clang-tidy` simply doesn't understand such a situation.  So, perhaps a version of (1) that targets that particular usage.  [I think that is what you have in the new PR.]

Even then, it seems we still have problems with mutators that destroy values they didn't create.  A mutator should be a mostly(*) pure function, otherwise re-invoking it is problematic.  So, I think we need to audit all our mutators.

(*) We do have some mutators that have benign side effects like counting.

devjgm:
Dropping this PR in favor of https://github.com/googleapis/google-cloud-cpp-spanner/pull/1239

google/cloud/spanner/internal/connection_impl.cc
devbww: I was a little over-reaching in my jump to `Commit()` mutators.  They surely should behave how I described, but there is no evidence that they currently misbehave.

This PR was all about `Visit()` and how the `Transaction` must outlive the functor.  Hence the need for taking the additional copy (reference).

PR #1239, opened 2020-02-04
chore: silence clang-tidy warning on internal::Visit

devjgm:
This fix will silence clang-tidy's `performance-unnecessary-value-param` warning when run on our headers.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1239)

### DIFF
--- a/google/cloud/spanner/transaction.h
+++ b/google/cloud/spanner/transaction.h
@@ -205,7 +205,7 @@ inline Transaction MakeReadWriteTransaction(
  * success.
  */
 inline Transaction MakeReadWriteTransaction(
-    Transaction txn, Transaction::ReadWriteOptions opts = {}) {
+    Transaction const& txn, Transaction::ReadWriteOptions opts = {}) {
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/transaction.h
```diff
@@ -205,7 +205,7 @@ inline Transaction MakeReadWriteTransaction(
  * success.
  */
 inline Transaction MakeReadWriteTransaction(
-    Transaction txn, Transaction::ReadWriteOptions opts = {}) {
+    Transaction const& txn, Transaction::ReadWriteOptions opts = {}) {
```
devbww:
```
google/cloud/spanner/client.cc:197:38: error: passing result of std::move() as a const reference argument; no move will actually happen [performance-move-const-arg,-warnings-as-errors]
      txn = MakeReadWriteTransaction(std::move(txn));
                                     ^~~~~~~~~~   ~
```

devjgm:
Related to #355

PR #1241, opened 2020-02-04
fix: make c'tor explicit

devjgm:
This prevents a upcoming warning when we enable clang-tidy on headers.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1241)


devjgm:
We decided offline to disable [this clang-tidy warning](https://clang.llvm.org/extra/clang-tidy/checks/modernize-return-braced-init-list.html), so this PR is no longer needed.

PR #1238, opened 2020-02-04
chore: upgrade to google-cloud-cpp-common v0.18.0

devbww:
Also picks up the pending upgrade to protobuf v3.11.3.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1238)

PR #1242, opened 2020-02-04
ci: upgrade protobuf to v3.11.3

coryan:
Upgrade protobuf to v3.11.3, this is needed to build with the latest and
greatest googleapis on Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1242)


coryan:
Actually, I now realize that #1238 has the same changes, let's just do it there.

PR #1259, opened 2020-02-12
feat: use `BackgroundThreads` in `SessionPool`

mr-salty:
Currently this just schedules a recurring 5s timer and cancels it in
the destructor.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1259)


google/cloud/spanner/internal/session_pool.cc
```diff
@@ -89,6 +99,45 @@ void SessionPool::Initialize() {
...
+}
+
+SessionPool::~SessionPool() {
+  // Taking `mu_` here is not strictly required because all references to this
+  // object are via `shared_ptr` and the fact we're in the destructor implies
+  // there are no more references.
+  //
+  // Note that it *is* possible we're racing against the timer lambda in
+  // `ScheduleBackgroundWork`. That does have a `weak_ptr` to `this`, but
+  // the `lock()` call (which must not have yet executed, else we wouldn't be
```
devbww: "not have yet executed" needs to be more precise methinks.  We could, for example, be executing `lock()` right now.  Or it may have just completed execution and returned null.

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -89,6 +99,45 @@ void SessionPool::Initialize() {
...
+  // in the destructor) will return `nullptr` in that case, so it will not
+  // do any work nor reschedule the timer.
+  std::unique_lock<std::mutex> lk(mu_);
+  timer_future_.cancel();
+}
+
+void SessionPool::ScheduleBackgroundWork(std::chrono::seconds relative_time) {
+  // `mu_` must be held by the caller.
```
devbww: Why is that?  It seems to me that these calls had better be serialized already.

PR #1266, opened 2020-02-14
refactor: Migrate away from `grpc_utils` namespace

### LINKED ISSUE #1261: Migrate away from `grpc_utils` namespace
`g-c-cpp-common` has deprecated the `grpc_utils` namespace (and it's use in `#include` paths and build rules); backwards compatibility has been maintained via forwarding but we should move to the non-`grpc_utils` versions.

https://github.com/googleapis/google-cloud-cpp-common/pull/158

mr-salty:
Fixes #1261

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1266)


google/cloud/spanner/internal/database_admin_stub.cc
```diff
@@ -46,7 +46,7 @@ class DefaultDatabaseAdminStub : public DatabaseAdminStub {
     grpc::Status status =
         database_admin_->CreateDatabase(&client_context, request, &response);
     if (!status.ok()) {
-      return google::cloud::grpc_utils::MakeStatusFromRpcError(status);
+      return MakeStatusFromRpcError(status);
```
devjgm: FYI, these unqualified calls to `MakeStatusFromRpcError()` will use ADL. That may be OK. but qualifying them with `google::cloud::` would avoid the ADL and make their location clear. It was previously not using ADL. Not a huge deal either way.

I'll let you decide if you'd like to change anything.

mr-salty:
___
*[google/cloud/spanner/internal/database_admin_stub.cc, line 49 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1266#-M00NGGZjgA51guEG_Pm-r1-49:-M00QOMs-nlLmkVVu1G1:b-2lw0j1) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/b2ebb8637247680f3748c219f5997101ee2cc8fd/google/cloud/spanner/internal/database_admin_stub.cc#L49)):*
<details><summary><i>Previously, devjgm (Greg Miller) wrote…</i></summary><blockquote>

FYI, these unqualified calls to `MakeStatusFromRpcError()` will use ADL. That may be OK. but qualifying them with `google::cloud::` would avoid the ADL and make their location clear. It was previously not using ADL. Not a huge deal either way.

I'll let you decide if you'd like to change anything.
</blockquote></details>

Blah. I'm happy to go either way as long as we're consistent - LMK what you think.

Some calls (in other files) were already using `grpc_utils::MakeStatusFromRpcError()` which turned into an unqualified call in this CL, so I decided to make them all look the same - but I didn't consider ADL, I was only thinking about the enclosing namespace `::google::cloud`.

mr-salty:
___
*[google/cloud/spanner/internal/database_admin_stub.cc, line 49 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1266#-M00NGGZjgA51guEG_Pm-r1-49:-M03OHfZDttN1-10oI8Q:b5nhq1q) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/b2ebb8637247680f3748c219f5997101ee2cc8fd/google/cloud/spanner/internal/database_admin_stub.cc#L49)):*
<details><summary><i>Previously, devjgm (Greg Miller) wrote…</i></summary><blockquote>

I'm happy with whatever you decide. I don't know if we have or need a hard rule for this. In general, I think the heuristic that I tend to use is that an unqualified function call is likely defined in the same file or at least the same namespace as what I'm calling. But if the function is in some other namespace, I'd expect some namespace qualification. I would probably `perl -p -i -e 's,MakeStatusFromRpcError,google::cloud::MakeStatusFromRpcError,g' **.{h,cc}` in this case. But if that makes things too inconsistent with other things we're doing, then I'm happy to go w/ consistency until we decide otherwise.

Up to you. Whatever LGTM.
</blockquote></details>

Having been bitten before by ADL with make_unique<>, I'll change it.

I think we should have a guideline, just for the sake of consisistency - in the old code we were mixing the two, even within the same file. Of course without a clang-tidy it's going to be easy to do that again. I don't see a clang-tidy about ADL, I wonder if such a thing would be feasible to do and useful, or if it would have too many false positives - it would probably have to be disabled for operators at least.

PR #1269, opened 2020-02-18
feat: environment-variable override of the spanner endpoint

devbww:
When the `SPANNER_EMULATOR_HOST` environment variable is set, the
endpoint and credentials for data and admin connections are overridden
by `${SPANNER_EMULATOR_HOST}` and `grpc::InsecureChannelCredentials()`
respectively.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1269)

### DIFF
--- a/google/cloud/spanner/connection_options.h
+++ b/google/cloud/spanner/connection_options.h
@@ -52,7 +52,7 @@ class ConnectionOptions {
   /// `grpc::GoogleDefaultCredentials()`.
   ConnectionOptions& set_credentials(
       std::shared_ptr<grpc::ChannelCredentials> v) {
-    credentials_ = std::move(v);
+    if (!emulator_override_) credentials_ = std::move(v);
     return *this;
   }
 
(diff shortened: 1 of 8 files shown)


google/cloud/spanner/connection_options.h
```diff
@@ -52,7 +52,7 @@ class ConnectionOptions {
   /// `grpc::GoogleDefaultCredentials()`.
   ConnectionOptions& set_credentials(
       std::shared_ptr<grpc::ChannelCredentials> v) {
-    credentials_ = std::move(v);
+    if (!emulator_override_) credentials_ = std::move(v);
```
devjgm: I forgot to mention that we should probably add some documentation somewhere about this environment variable. I'm thinking that connection_options.h is probably the right spot.

devbww: For the record, we not longer need this documentation after #1277.

PR #1277, opened 2020-02-21
feat: make emulator endpoint/credentials changes as late as possible

devbww:
Rather than making the `${SPANNER_EMULATOR_HOST}` endpoint/credentials
changes during `ConnectionOptions` construction (which then requires
that we ignore further attempts to set those fields), do them as late
as possible (i.e., just before creating each variety of channel).

This then facilitates the use of a common `ConnectionOptions` class
between Cloud C++ projects by not requiring special behavior in the
`set` functions.

While this modifies the signatures of `CreateDefault*Stub()`, it is
only a change from pass-by-const-ref to pass-by-value, which will not
require any change to user code.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1277)


google/cloud/spanner/connection_options_test.cc
```diff
@@ -55,22 +55,14 @@ TEST(ConnectionOptionsTest, AdminEndpoint) {
...
+  // grpc::InsecureChannelCredentials().
   EnvironmentVariableRestore restore("SPANNER_EMULATOR_HOST");
   google::cloud::internal::SetEnv("SPANNER_EMULATOR_HOST", "localhost:9010");
-
   ConnectionOptions options(std::shared_ptr<grpc::ChannelCredentials>{});
-
-  // When SPANNER_EMULATOR_HOST is set, the passed credentials are overridden
-  // (by grpc::InsecureChannelCredentials()) and further updates are ignored.
+  options = internal::EmulatorOverrides(options);
   EXPECT_NE(std::shared_ptr<grpc::ChannelCredentials>{}, options.credentials());
```
devjgm: Are we able to verify that the credentials are indeed the InsecureChannelCredentials rather than just not ChannelCredentials?

coryan: Two shinny pennies say it is `InsecureChannelCredentials`: that is what we use for Cloud Bigtable, and there is no way we are asking folks to setup trusted SSL certificates to run the emulator.

devbww: > Are we able to verify that the credentials are indeed the InsecureChannelCredentials rather than just not ChannelCredentials?

Not that I'm aware of.  Note that we are checking that the credentials went from null to non-null, which means `EmulatorOverrides()` did something active.

coryan: Doh! I misread the question.  I tried checking the type of credentials, could not find a way. They are hidden inside the gRPC "C++ wrappers for C wrappers around C++".

Issue #1260, opened 2020-02-12
Use ConnectionOptions from `g-c-cpp-common`


mr-salty:
We now have a version of `ConnectionOptions` in `common` (see googleapis/google-cloud-cpp-common#165) - we should remove the code from this repository, leave the header and a type alias for backwards compatibility.

From the above-referenced issue, note:

> This is not strictly a copy, the defaults are different for each service, for example: the endpoint. The class is a template, parametized by traits to set the defaults. Each client library will instantiate this class with the right set of defaults.

PR #1279, opened 2020-02-21
cleanup: change to common's {Connection,Tracing}Options

devjgm:
Fixes: #1260
Fixes: #1251

This PR removes most tests that are now unnecessary. I kept some of the tests in `connection_options_test.cc`, because these are the tests that verify properties that are configured by Spanner's custom Traits class.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1279)


google/cloud/spanner/tracing_options.h
```diff
@@ -16,6 +16,7 @@
 #define GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_TRACING_OPTIONS_H
 
 #include "google/cloud/spanner/version.h"
+#include "google/cloud/tracing_options.h"
 #include <cstdint>
```
devbww: Unused.

google/cloud/spanner/tracing_options.h
```diff
@@ -16,6 +16,7 @@
...
 #include <string>
```
devbww: Unused.

google/cloud/spanner/tracing_options.h
devjgm: Removed

google/cloud/spanner/tracing_options.h
devjgm: removed

google/cloud/spanner/CMakeLists.txt
```diff
@@ -300,7 +299,6 @@ function (spanner_client_define_tests)
 
     set(spanner_client_unit_tests
         # cmake-format: sortable
-        connection_options_test.cc
```
tmatsuo: Shouldn't we keep this test?

devjgm: Yes, you're right. Good catch. I accidentally deleted that one. PR coming up.

PR #1305, opened 2020-02-27
fix: create "init_client" snippet for "Getting started" doc

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1305)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -697,8 +697,11 @@ void Quickstart(std::string const& project_id, std::string const& instance_id,
                 std::string const& database_id) {
   namespace spanner = ::google::cloud::spanner;
 
-  spanner::Client client(spanner::MakeConnection(
-      spanner::Database(project_id, instance_id, database_id)));
+  //! [START init_client]
+  auto database = spanner::Database(project_id, instance_id, database_id);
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/samples/samples.cc
```diff
@@ -697,8 +697,11 @@ void Quickstart(std::string const& project_id, std::string const& instance_id,
                 std::string const& database_id) {
   namespace spanner = ::google::cloud::spanner;
 
-  spanner::Client client(spanner::MakeConnection(
-      spanner::Database(project_id, instance_id, database_id)));
+  //! [START init_client]
```
coryan: Is that one of the canonical region tags?  Shouldn't this be something like `spanner_init_client` or something?

devjgm: AFAICT, there is no canonical tag for this particular case. Most of the libraries use the "spanner_quickstart" tag, which shows the wrong code snippet. Go and Java show the correct snippet, but they use a different region tag. Go uses a function named `createClients`, and Java uses a region tag named `init_client`. We decided to copy Java's region tag since we're showing the equivalent thing to what Java's showing.

Separately, we'll file a bug that all the other languages are showing the wrong snippet in this case. I imagine they'll all standardize on this tag.

PR #1317, opened 2020-02-27
ci: cache docker images

devjgm:
Enable use of GCR to cache docker images to speed up builds.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1317)


coryan:
All the builds pass and the code is good, I am merging on your behalf so everybody starts getting faster builds tomorrow morning.

PR #1319, opened 2020-02-28
feat: add a `Commit()` helper that takes the `Mutations` directly

### LINKED ISSUE #1315: Consider: Client::Commit(Mutations)
Seems like it may be a convenient function, to save users from having to create an unnecessary lambda.

devbww:
This convenience function saves the user from having to create a
small, but non-trivial lambda.

Fixes #1315.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1319)

### DIFF
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -481,13 +481,10 @@ class Client {
       std::function<StatusOr<Mutations>(Transaction)> const& mutator);
 
   /**
-   * Commits a read-write transaction.
+   * Commits the given @p mutations atomically in order.
    *
-   * Same as above, with a mutator that returns @p mutations.
-   *
-   * @param mutations The mutations to be executed when this transaction
-   *     commits. All mutations are applied atomically, in the order they
-   *     appear in this list.
+   * This function uses the re-run loop described above with the default
+   * policies.
    */
   StatusOr<CommitResult> Commit(Mutations mutations);
 
--- a/google/cloud/spanner/client.h
+++ b/google/cloud/spanner/client.h
@@ -483,6 +483,17 @@ class Client {
   /**
    * Commits a read-write transaction.
    *
+   * Same as above, with a mutator that returns @p mutations.
+   *
+   * @param mutations The mutations to be executed when this transaction
+   *     commits. All mutations are applied atomically, in the order they
+   *     appear in this list.
+   */
+  StatusOr<CommitResult> Commit(Mutations mutations);
+
+  /**
+   * Commits a read-write transaction.
+   *
    * The commit might return a `kAborted` error. This can occur at any time.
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/client.h
```diff
@@ -478,6 +478,17 @@ class Client {
   StatusOr<CommitResult> Commit(
       std::function<StatusOr<Mutations>(Transaction)> const& mutator);
 
+  /**
+   * Commits a read-write transaction.
```
devjgm: This comment doesn't seem quite right since it mentions a `transaction` and a `mutator`, which this function doesn't accept. What do you think about something like: 

`Commits the given @p mutations.`

devbww: None of the "same as above" functions accept a `Transaction`, yet they all really "commit a read-write transaction."  So, I'm inclined to leave that for now, but am certainly open to revamping them all.  (Although, read on.)

As for `mutator`, it is meant to refer to the argument to the "same as above" function.  That is, that we generate "a mutator that returns @p mutations" and call the previous function (which is, of course, exactly what we do).  The "mutator" mention does not include @p.

That said, what I really wanted to get across is that this function includes the re-run loop described above (unlike `Commit(Transaction, Mutations)` described below).  If you believe `Commits the given @p mutations` is sufficient to convey that, then I'm happy to go there.  LMK.

devjgm: > None of the "same as above" functions accept a `Transaction`,

All those functions accept a caller-provided `mutator`, so mentioning a `Transaction` makes sense. I.e., the caller will understand what transaction the documentation is referring to.

> yet they all really "commit a read-write transaction." So, I'm inclined to leave that for now, but am certainly open to revamping them all. (Although, read on.)

The difference with this function is that there is no `Transaction` anywhere for the caller to see/use/think about. `Transaction` is just not part of this function's interface.

> As for `mutator`, it is meant to refer to the argument to the "same as above" function. That is, that we generate "a mutator that returns @p mutations" and call the previous function (which is, of course, exactly what we do). The "mutator" mention does not include @p.

Those are implementation details. If we think we need to document the implementation of this function, then maybe we should call that out as the fact that we're documenting the implementation (or inline the function body). I, personally, don't think it's necessary to document the implementation in this case.

> 
> That said, what I really wanted to get across is that this function includes the re-run loop described above (unlike `Commit(Transaction, Mutations)` described below). If you believe `Commits the given @p mutations` is sufficient to convey that, then I'm happy to go there. LMK.

My vote:

```cc
/**
 * Commits the given @p mutations atomically in order.
 *
 * This function uses the re-run loop described above with the default policies.
 */
```

devbww: Done.

PR #1322, opened 2020-02-28
cleanup: simplify error handling in examples

coryan:
Now that the `Commit()` loop reruns when `StatusOr<T>::value()` throws
the "right" errors we can remove a bit of code from the samples.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1322)

### DIFF
--- a/google/cloud/spanner/samples/samples.cc
+++ b/google/cloud/spanner/samples/samples.cc
@@ -1075,30 +1075,23 @@ void ReadWriteTransaction(google::cloud::spanner::Client client) {
     auto rows = client.Read(std::move(txn), "Albums", std::move(key),
                             {"MarketingBudget"});
     using RowType = std::tuple<std::int64_t>;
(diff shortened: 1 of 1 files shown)


devbww:
It looks like these would no behave differently in a "noex" build.  Feature or bug?

coryan:
> It looks like these would no behave differently in a "noex" build. Feature or bug?

In general, this program (samples.cc) does not compile with `noex`. It already throws in multiple places, so I think the change is fine.  Is that what you are asking about?

devbww:
> this program (samples.cc) does not compile with noex. It already throws in multiple places

Acknowledged.  I'm still not sure that isn't a problem, but ...

> Is that what you are asking about?

what I'm more concerned about is whether individual snippets should behave differently in exception and no-exception builds.  If a cut-and-pasted snippet won't work without exceptions, we should try to make that clear somehow.

All that said, I'm happy to be overruled by some other approver.

PR #1326, opened 2020-03-03
chore: update dependency versions

coryan:
This updates the version used for the -common and googleapis
dependencies. It also removes (instead of updating) the crc32c
installations, as the library is not used.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1326)

### DIFF
--- a/bazel/google_cloud_cpp_spanner_deps.bzl
+++ b/bazel/google_cloud_cpp_spanner_deps.bzl
@@ -42,11 +42,11 @@ def google_cloud_cpp_spanner_deps():
     if "com_github_googleapis_google_cloud_cpp_common" not in native.existing_rules():
         http_archive(
             name = "com_github_googleapis_google_cloud_cpp_common",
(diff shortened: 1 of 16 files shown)


bazel/google_cloud_cpp_spanner_deps.bzl
```diff
@@ -42,11 +42,11 @@ def google_cloud_cpp_spanner_deps():
     if "com_github_googleapis_google_cloud_cpp_common" not in native.existing_rules():
         http_archive(
             name = "com_github_googleapis_google_cloud_cpp_common",
-            strip_prefix = "google-cloud-cpp-common-0.20.0",
+            strip_prefix = "google-cloud-cpp-common-0.21.0",
```
devbww: `ci/etc/kokoro/install/version-config.sh` too?

coryan: Indeed. Thanks!

PR #1328, opened 2020-03-03
ci: reduce false negatives in code coverage

coryan:
For some reason the use of `& Function` was confusing the code coverage
tool and marking lines as "not covered" when in fact they are.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1328)


devbww:
Do #1328 and #1329 overlap?  I approved this one.

PR #1329, opened 2020-03-03
ci: fix code coverage results in pull requests

coryan:
Inline the upload-coverage.sh script as it is used in only one place and
it had overly complex parameter passing: some via environment variables,
e.g. expected PROJECT_ID, and some via command-line arguments, and some
variables were recomputed, e.g., IMAGE.

This is why the code coverage results were broken in PRs, PROJECT_ID was
not set as an environment variable, which make the script assume the
Docker image used in the build had a different name than what we
actually used.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1329)


ci/kokoro/docker/build.sh
```diff
@@ -464,7 +464,64 @@ else
...
+      "--env" "CI_BUILD_ID=${KOKORO_BUILD_NUMBER:-}"
+      "--env" "CI_JOB_ID=${KOKORO_BUILD_NUMBER:-}"
+  )
+
+  echo -n "Uploading code coverage to codecov.io..."
+  # Run the upload script from codecov.io within a Docker container. Save the log
+  # to a file because it can be very large (multiple MiB in size).
+  sudo docker run "${docker_flags[@]}" "${IMAGE}:latest" /bin/bash -c \
+      "/bin/bash <(curl -s https://codecov.io/bash) >/v/${BUILD_OUTPUT}/codecov.log 2>&1"
+  exit_status=$?
```
devbww: It looks like this will lose any indication that `publish-refdocs.sh` failed.

Perhaps we should never overwrite a non-zero `exit_status` with a zero?

coryan: Done.

coryan:
PTAL.

Issue #1303, opened 2020-02-26
Add output streaming support to spanner::Value


devjgm:
Currently [spanner::Value](https://github.com/googleapis/google-cloud-cpp-spanner/blob/master/google/cloud/spanner/value.h) cannot be stringified or streamed. However, it seems useful (e.g., when printing output from a `select *`) to be able to stringify and print `spanner::Value` objects, even if the caller doesn't know their C++ type. Therefore, we probably want to add:

```cc
std::ostream& operator<<(std::ostream& os, Value const& v);
```

The stringified format for most types is obvious. For the less obvious types, like ARRAY and STRUCT, we should use the string format used in Spanner's SQL syntax.

See [array docs](https://cloud.google.com/spanner/docs/arrays).
See [struct docs](https://cloud.google.com/spanner/docs/query-syntax#using-structs-with-select)

devjgm:
Done. We're still tweaking bits of the formatting of certain parts of a streamed Value, but those are being tracked in separate issues.

PR #1336, opened 2020-03-05
feat: add ostream operator for Value

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1336)

### DIFF
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -116,7 +116,6 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
       return os
              << "B\""
              << internal::BytesFromBase64(v.string_value())->get<std::string>()
-
              << "\"";
 
     case google::spanner::v1::ARRAY: {
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -88,6 +88,71 @@ bool Equal(google::spanner::v1::Type const& pt1,
   }
 }
 
+std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
+                           google::spanner::v1::Type const& t) {
+  if (v.kind_case() == google::protobuf::Value::kNullValue) {
+    return os << "NULL";
+  }
+
+  switch (t.code()) {
+    case google::spanner::v1::BOOL:
+      return os << (v.bool_value() ? "TRUE" : "FALSE");
+
+    case google::spanner::v1::TIMESTAMP:
+    case google::spanner::v1::DATE:
+    case google::spanner::v1::INT64:
+      return os << v.string_value();
+
+    case google::spanner::v1::FLOAT64:
+      if (v.kind_case() == google::protobuf::Value::kStringValue) {
+        return os << v.string_value();
+      }
+      return os << v.number_value();
+
+    case google::spanner::v1::STRING:
+      return os << "\"" << v.string_value() << "\"";
+
+    case google::spanner::v1::BYTES:
+      return os
+             << "B\""
+             << internal::BytesFromBase64(v.string_value())->get<std::string>()
+
+             << "\"";
+
+    case google::spanner::v1::ARRAY: {
+      const char* delimiter = "";
+      os << '[';
+      for (auto const& e : v.list_value().values()) {
+        os << delimiter;
+        StreamHelper(os, e, t.array_element_type());
+        delimiter = ", ";
+      }
+      return os << ']';
+    }
+
+    case google::spanner::v1::STRUCT: {
+      const char* delimiter = "";
+      os << '(';
+      for (int i = 0; i < v.list_value().values_size(); ++i) {
+        os << delimiter;
+        if (!t.struct_type().fields(i).name().empty()) {
+          os << t.struct_type().fields(i).name() << ": ";
+        }
+        StreamHelper(os, v.list_value().values(i),
+                     t.struct_type().fields(i).type());
+        delimiter = ", ";
+      }
+      return os << ')';
+    }
+
+    case google::spanner::v1::TYPE_CODE_UNSPECIFIED:
+    case google::spanner::v1::TypeCode_INT_MIN_SENTINEL_DO_NOT_USE_:
+    case google::spanner::v1::TypeCode_INT_MAX_SENTINEL_DO_NOT_USE_:
+      break;
+  }
+  return os;
+}
+
 }  // namespace
 
 namespace internal {
@@ -108,6 +173,10 @@ bool operator==(Value const& a, Value const& b) {
 
 void PrintTo(Value const& v, std::ostream* os) {
   *os << v.type_.ShortDebugString() << "; " << v.value_.ShortDebugString();
+}
+
+std::ostream& operator<<(std::ostream& os, Value const& v) {
+  return StreamHelper(os, v.value_, v.type_);
 }
 
 //
--- a/google/cloud/spanner/value.h
+++ b/google/cloud/spanner/value.h
@@ -314,6 +314,20 @@ class Value {
    */
   friend void PrintTo(Value const& v, std::ostream* os);
 
+  /**
+   * Outputs string representation of a given Value to the provided stream.
+   *
+   * @warning This is intended for debugging and human consumption only, not
+   *   machine consumption as the output format may change without notice.
+   *
+   * @par Example:
+   * @code
+   *   spanner::Value v{42};
+   *   std::cout << v << "\n";
+   * @endcode
+   */
+  friend std::ostream& operator<<(std::ostream& os, Value const& v);
+
  private:
   // Metafunction that returns true if `T` is an optional<U>
   template <typename T>
--- a/google/cloud/spanner/value_test.cc
+++ b/google/cloud/spanner/value_test.cc
@@ -973,7 +973,6 @@ INSTANTIATE_TEST_SUITE_P(
         std::make_tuple("StringNullValue", Value(optional<std::string>()),
                         "NULL"),
         std::make_tuple("StringLiteralNullValue", Value("NULL"), "\"NULL\""),
-
         std::make_tuple("Bytes", Value(Bytes(std::string("DEADBEEF"))),
                         "B\"DEADBEEF\""),
         std::make_tuple(
--- a/google/cloud/spanner/value_test.cc
+++ b/google/cloud/spanner/value_test.cc
@@ -35,6 +35,8 @@ namespace {
 
 using ::google::cloud::spanner_testing::IsProtoEqual;
 using ::testing::Not;
+using ::testing::TestParamInfo;
+using ::testing::TestWithParam;
 
 std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds>
 MakeTimePoint(std::time_t sec, std::chrono::nanoseconds::rep nanos) {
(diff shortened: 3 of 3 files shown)


scotthart:
fixes #1303

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,79 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+  }
+  return os << std::get<2>(delimiters_type_function(i));
+}
+
+std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
+                           google::spanner::v1::Type const& t) {
+  if (v.kind_case() == google::protobuf::Value::kNullValue &&
+      t.code() != google::spanner::v1::ARRAY &&
+      t.code() != google::spanner::v1::STRUCT) {
+    return os;
```
devbww: Q. Should this render "NULL"?

google/cloud/spanner/value_test.cc
```diff
@@ -904,6 +912,75 @@ TEST(Value, CommitTimestamp) {
...
+        std::make_tuple("NaNFloat64", Value(std::nan("NaN")), "NaN"),
+        std::make_tuple("String", Value("Seatac Astronomy"),
+                        "Seatac Astronomy"),
+        std::make_tuple("Bytes", Value("DEADBEEF"), "DEADBEEF"),
+        std::make_tuple(
+            "Timestamp",
+            Value(MakeTimestamp(MakeTimePoint(1561147549LL, 0)).value()),
+            "2019-06-21T20:05:49Z"),
+        std::make_tuple("Date", Value(Date(1970, 1, 1)), "1970-01-01")),
+
```
devbww: Nit: Delete blank line to match other instantiations (or vice versa)?

google/cloud/spanner/value.cc
scotthart: Perhaps, but for a string field how would I differentiate between a null value and a literal of "NULL"?

google/cloud/spanner/value_test.cc
scotthart: Done.

google/cloud/spanner/value.cc
devbww: Perhaps a string field renders with '"' delimiters (just like structs and arrays have delimiters)?

google/cloud/spanner/value_test.cc
```diff
@@ -945,13 +945,12 @@ INSTANTIATE_TEST_SUITE_P(
         std::make_tuple("NaNFloat64", Value(std::nan("NaN")), "NaN"),
         std::make_tuple("String", Value("Seatac Astronomy"),
                         "Seatac Astronomy"),
-        std::make_tuple("Bytes", Value("DEADBEEF"), "DEADBEEF"),
+        std::make_tuple("Bytes", Value(Bytes("DEADBEEF")), "DEADBEEF"),
```
devbww: Perhaps BYTES should render with the `b` prefix (and the quotes) ... `b"DEADBEEF"`.

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,79 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+    case google::spanner::v1::FLOAT64:
+      if (v.kind_case() == google::protobuf::Value::kStringValue) {
+        return os << v.string_value();
+      }
+      return os << std::to_string(v.number_value());
+
+    case google::spanner::v1::TIMESTAMP:
+    case google::spanner::v1::DATE:
+    case google::spanner::v1::STRING:
+    case google::spanner::v1::BYTES:
```
devbww: We might want to consider escape sequence for non-printable chars.

google/cloud/spanner/value_test.cc
```diff
@@ -944,8 +944,8 @@ INSTANTIATE_TEST_SUITE_P(
                         "-Infinity"),
         std::make_tuple("NaNFloat64", Value(std::nan("NaN")), "NaN"),
         std::make_tuple("String", Value("Seatac Astronomy"),
-                        "Seatac Astronomy"),
-        std::make_tuple("Bytes", Value(Bytes("DEADBEEF")), "DEADBEEF"),
+                        "\"Seatac Astronomy\""),
+        std::make_tuple("Bytes", Value(Bytes("DEADBEEF")), "B\"DEADBEEF"),
```
devbww: It looks like there should be a trailing `\"` there.

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,83 @@ bool Equal(google::spanner::v1::Type const& pt1,
   }
 }
 
+std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
+                           google::spanner::v1::Type const& t) {
+  if (v.kind_case() == google::protobuf::Value::kNullValue &&
+      t.code() != google::spanner::v1::ARRAY &&
+      t.code() != google::spanner::v1::STRUCT) {
+    return os << "";
```
devbww: Why stream the `""`?  Did you mean `"NULL"`?

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,83 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+  }
+
+  int i = 0;
```
devbww: I think the observation was that the `i` could be local to the `for` loops.  Its use after the loop could be replaced with another expression given that the loops don't `break`.  But it also looks the the "size - 1" loops could be replaced with full loops by using the `std::string sep` idiom.

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,83 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+    case google::spanner::v1::INT64:
+      return os << v.string_value();
+
+    case google::spanner::v1::FLOAT64:
+      if (v.kind_case() == google::protobuf::Value::kStringValue) {
+        return os << v.string_value();
+      }
+      return os << std::to_string(v.number_value());
+
+    case google::spanner::v1::TIMESTAMP:
```
devbww: Might as well group these with `INT64`.

google/cloud/spanner/value.h
```diff
@@ -314,6 +314,18 @@ class Value {
    */
   friend void PrintTo(Value const& v, std::ostream* os);
 
+  /**
+   * Outputs string representation of a given Value to the provided stream.
```
devjgm: I think this should copy the warning from the `PrintTo` function:

```
@warning This is intended for debugging and human consumption only, not
    machine consumption as the output format may change without notice.
```

scotthart:
created #1345 to track opportunity to improve bytes formatting.

google/cloud/spanner/value.h
scotthart: Done.

google/cloud/spanner/value.cc
devjgm: Right. For example, I think the array case might look like this:

```cc
    case google::spanner::v1::ARRAY: {
      os << '[';
      char const* sep = nullptr;
      for (auto const& val : v.list_value().values()) {
        if (sep != nullptr) os << sep;
        StreamHelper(os, val, t.array_element_type());
        sep = ", ";
      }
      return os << ']';
    }
```

Struct could simplify to something like this as well.

google/cloud/spanner/value.cc
scotthart: Done.

google/cloud/spanner/value.cc
scotthart: Done.

google/cloud/spanner/value.cc
scotthart: Done.

google/cloud/spanner/value_test.cc
scotthart: Done.

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,75 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+        StreamHelper(os, e, t.array_element_type());
+        delimiter = ", ";
+      }
+      return os << ']';
+    }
+    case google::spanner::v1::STRUCT: {
+      const char* delimiter = "";
+      os << '(';
+      for (int i = 0;
+           v.list_value().values_size() > 0 &&
```
devbww: This looks redundant.

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,77 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+    case google::spanner::v1::TIMESTAMP:
+    case google::spanner::v1::DATE:
+    case google::spanner::v1::INT64:
+      return os << v.string_value();
+
+    case google::spanner::v1::FLOAT64:
+      if (v.kind_case() == google::protobuf::Value::kStringValue) {
+        return os << v.string_value();
+      }
+      return os << std::to_string(v.number_value());
```
devjgm: Is this `std::to_string` needed?

devjgm:
ftr, I didn't mean `Thanks, Scott?`, I meant `Thanks, Scott!`

google/cloud/spanner/value.cc
scotthart: Depends. 
Without the to_string 42.0 is printed as 42 and 5.2 as 5.2.
With to_string you get 42.000000 and 5.200000

devjgm: I'd go w/ the first output, w/o the std::to_string, but I'll let you decide.

google/cloud/spanner/value.cc
scotthart: Done.

google/cloud/spanner/value.cc
scotthart: Done.

google/cloud/spanner/value_test.cc
```diff
@@ -904,6 +912,81 @@ TEST(Value, CommitTimestamp) {
...
+        std::make_tuple("InfinityFloat64",
+                        Value(std::numeric_limits<double>::infinity()),
+                        "Infinity"),
+        std::make_tuple("NegativeInfinityFloat64",
+                        Value(-std::numeric_limits<double>::infinity()),
+                        "-Infinity"),
+        std::make_tuple("NaNFloat64", Value(std::nan("NaN")), "NaN"),
+        std::make_tuple("String", Value("Seatac Astronomy"),
+                        "\"Seatac Astronomy\""),
+        std::make_tuple("Bytes", Value(Bytes("DEADBEEF")), "B\"DEADBEEF\""),
```
devbww: Creating a `Bytes` value from an array `a[N]` is equivalent to `Bytes(&a[0], &a[N])`.  So, `Bytes("DEADBEEF")` would mean something like `Bytes({'D','E','A','D','B','E','E','F','\0'})`.  That is, there is a NUL in there, so the expected value should not match.

Use `Bytes(std::string("DEADBEEF"))` instead, at least until we can determine whether this is a bug or a feature.

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,73 @@ bool Equal(google::spanner::v1::Type const& pt1,
   }
 }
 
+std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
+                           google::spanner::v1::Type const& t) {
+  if (v.kind_case() == google::protobuf::Value::kNullValue &&
+      t.code() != google::spanner::v1::ARRAY &&
```
devjgm: I believe there's a difference between an empty array and a null array, and we probably want to print them differently too. I think all NULL values should print the string "NULL", including arrays and structs. And empty values should print as "[]" or "()". To make this work, I think you may just need to remove the two ARRAY and STRUCT conditions here.

WDYT?

google/cloud/spanner/value.cc
```diff
@@ -88,6 +88,71 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+      return os << v.number_value();
+
+    case google::spanner::v1::STRING:
+      return os << "\"" << v.string_value() << "\"";
+
+    case google::spanner::v1::BYTES:
+      return os
+             << "B\""
+             << internal::BytesFromBase64(v.string_value())->get<std::string>()
+
```
devjgm: Remove blank line.

PR #1340, opened 2020-03-06
cleanup: use same clang-tidy in all repositories

coryan:
This repository was already using the clang-tidy distributed with
Fedora:30, just changing to Fedora:31 brings it in line with other
repositories (`-common`, `-pubsub`, and `-bigquery`).

Cleanup the ubuntu Dockerfile, removing the formatting tools that we do
not use in any of the builds based on that image.

Part of the work for googleapis/google-cloud-cpp-common#204

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1340)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,6 +6,10 @@
 #  -google-readability-namespace-comments: the *_CLIENT_NS is a macro, and
 #      clang-tidy fails to match it against the initial value.
 #
+#  -modernize-use-trailing-return-type: clang-tidy recommends using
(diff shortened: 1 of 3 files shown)


.clang-tidy
```diff
@@ -36,6 +40,7 @@ Checks: >
   -modernize-return-braced-init-list,
   -performance-move-const-arg,
   -readability-named-parameter,
+  -modernize-use-trailing-return-type,
```
devjgm: I think this list is (mostly?) sorted, so maybe move this up 2 lines.

Issue #1172, opened 2020-01-08
maintain desired SessionPool size


mr-salty:
There are [2 parameters](https://github.com/googleapis/google-cloud-cpp-spanner/blob/e33b4f814e7c30f5c055431d07ea4ae99ac619f0/google/cloud/spanner/session_pool_options.h#L134-L136) that specify the desired `SessionPool` size (plus `max_sessions_per_channel` which establishes an upper bound). The `SessionPool` should create or delete `Sessions` in the background to maintain the desired size.

* `min_sessions` specifies the minimum number of `Sessions` associated with the pool, whether idle or in-use. Currently we create `min_sessions` `Sessions` at pool creation time but do not replenish `Sessions` if the number drops below that
* `max_idle_sessions` specified the maximum number of idle `Sessions` to keep in the pool; any `Sessions` in excess of this number should be periodically deleted.

mr-salty:
Still TODO here:

* Use the async `BatchCreateSessions` call once #1365 is done.
* Determine when sessions need to be deleted (by comparing the pool size vs. `options_`, fairly trivial) and use async `DeleteSession` to delete them.
* Get rid of uses of non-async `BatchCreateSessions` (this could probably be done after GA)

coryan:
@mr-salty I think we should close this now?

mr-salty:
I still have the "clean up old sessions" PR pending, I think it's done but I really need to do the fake clock before I can write tests for it. I can file another issue for the "delete" part though.

coryan:
>  I can file another issue for the "delete" part though.

That sounds like the way to go.

mr-salty:
done (#1435)

PR #1343, opened 2020-03-09
feat: refactor and enhance session creation logic

mr-salty:
Centralize all Session creation in the `Grow()` method, which computes
the number of Sessions to create on each channel to balance the load
across channels. Also introduces an async mode (not implemented here)
which will eventually become the only mode.

Also fixes a bug where the per-channel session counts were not being
updated properly when `Sessions` were marked bad or disassociated.
Maybe this part should be its own PR, and I don't really like how it
obfuscates the locking requirements so I'm open to suggestions on how
to do this in a cleaner way.

Part of #1172

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1343)

### DIFF
--- a/google/cloud/spanner/internal/session_pool.cc
+++ b/google/cloud/spanner/internal/session_pool.cc
@@ -25,6 +25,7 @@
 #include <chrono>
 #include <random>
 #include <thread>
+#include <vector>
 
 namespace google {
 namespace cloud {
@@ -73,31 +74,12 @@ SessionPool::SessionPool(Database db,
   }
   // `channels_` is never resized after this point.
   next_dissociated_stub_channel_ = channels_.begin();
-  next_channel_for_create_sessions_ = channels_.begin();
 }
 
 void SessionPool::Initialize() {
-  // Eagerly initialize the pool with `min_sessions` sessions.
-  // TODO(#307) this was moved to `Initialize` in preparation of using
-  // `shared_from_this()` in the process of creating sessions, which cannot
-  // be done in the constructor.
   if (options_.min_sessions() > 0) {
     std::unique_lock<std::mutex> lk(mu_);
-    int num_channels = static_cast<int>(channels_.size());
-    int sessions_per_channel = options_.min_sessions() / num_channels;
-    // If the number of sessions doesn't divide evenly by the number of
-    // channels, add one extra session to the first `extra_sessions` channels.
-    int extra_sessions = options_.min_sessions() % num_channels;
-    for (auto& channel : channels_) {
-      int num_sessions = sessions_per_channel;
-      if (extra_sessions > 0) {
-        ++num_sessions;
-        --extra_sessions;
-      }
-      // Just ignore failures; we'll try again when the caller requests a
-      // session, and we'll be in a position to return an error at that time.
-      (void)CreateSessions(lk, channel, options_.labels(), num_sessions);
-    }
+    (void)Grow(lk, options_.min_sessions(), /*async=*/false);
   }
   ScheduleBackgroundWork(std::chrono::seconds(5));
 }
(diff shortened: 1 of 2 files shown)


mr-salty:
___
*[google/cloud/spanner/internal/session.h, line 39 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1343#-M1zFpVp-6uDJT9FWD-e:-M2-PDoy9lnLeYsjUOUq:b-n5wo5r) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/237875ec6fe352f12a01bb8f82ba29f76a3c41ca/google/cloud/spanner/internal/session.h#L39)):*

FYI: passing pointers like this feels icky, and I consider it a code smell.  Is there an abstraction we are missing? Maybe an API to say "call here to decrease the session counter"?   Or maybe the "dissociate_from_pool" should be a `std::function` or abstract class that gets called when the session is removed, and sometimes we initialize it with a noop, and sometimes with a full object that decrements counters and so on?
</blockquote></details>

yeah, I mentioned this in the description, it's not ideal but I also don't think adding another layer of abstraction makes it any better.  Let me split this out to another PR like I should have done in the first place.

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -133,11 +115,99 @@ void SessionPool::ScheduleBackgroundWork(std::chrono::seconds relative_time) {
...
+/**
+ * Grow the session pool by creating up to `sessions_to_create` sessions and
+ * adding them to the pool.  Note that `lk` may be dropped in the process.
+ *
+ * TODO(#1271) eliminate the `async` parameter and do all creation
+ * asynchronously. The main obstacle is making existing tests pass.
+ */
+Status SessionPool::Grow(std::unique_lock<std::mutex>& lk,
+                         int sessions_to_create, bool async) {
+  create_in_progress_ = true;
```
devbww: Nit: Could this simply be moved down to avoid the `= false` in the "already at max" case?  It seems to be something that should be closely bound to `lk`.

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -133,11 +115,103 @@ void SessionPool::ScheduleBackgroundWork(std::chrono::seconds relative_time) {
...
+  // number of Sessions on each channel equal.
+  //
+  // However, the counts may become unequal over time, and we do not want
+  // to delete sessions just to make the counts equal, so do the best we
+  // can within those constraints.
+  int target_total_sessions =
+      (std::min)(total_sessions_ + sessions_to_create, session_limit);
+
+  // Sort the channels in *descending* order of session count.
+  std::vector<std::shared_ptr<Channel>> channels_by_count;
```
devjgm: Is this just equivalent to

```cc
auto chanenls_by_count = channels_;
```

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -133,11 +115,103 @@ void SessionPool::ScheduleBackgroundWork(std::chrono::seconds relative_time) {
...
+  create_in_progress_ = true;
+
+  // Compute how many Sessions to create on each Channel, trying to keep the
+  // number of Sessions on each channel equal.
+  //
+  // However, the counts may become unequal over time, and we do not want
+  // to delete sessions just to make the counts equal, so do the best we
+  // can within those constraints.
+  int target_total_sessions =
+      (std::min)(total_sessions_ + sessions_to_create, session_limit);
```
devjgm: fwiw, I don't think the `(std::min)` trick is needed in .cc files, because we know what our includes are and if our build works, then it works. This differes from `.h` files where callers may include us after including the offending Windows include that defines the macro for `min`.

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -133,11 +115,103 @@ void SessionPool::ScheduleBackgroundWork(std::chrono::seconds relative_time) {
...
+/**
+ * Grow the session pool by creating up to `sessions_to_create` sessions and
+ * adding them to the pool.  Note that `lk` may be released and reacquired in
+ * this method.
+ *
+ * TODO(#1271) eliminate the `async` parameter and do all creation
+ * asynchronously. The main obstacle is making existing tests pass.
+ */
+Status SessionPool::Grow(std::unique_lock<std::mutex>& lk,
+                         int sessions_to_create, bool async) {
```
devjgm: bool params like this lead to unclear call sites, hence the need to comment `/*async=*/` at call sites. Consider using an enum to make the code itself perfectly clear even without the comment.

PR #1353, opened 2020-03-10
feat: Keep track of the last time a `Session` was used

mr-salty:
Assume that the session was used (i.e. a RPC was made) just prior to it
being returned to the pool.

I debated doing this in `Stub()` which would be marginally more accurate,
but would require taking a lock we don't otherwise have to take. Since
our notion of "last use time" is already just an approximation of what
the backend considers "last use time" and the timeframes we care about
are measured in minutes, it didn't seem to be worthwhile.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1353)

### DIFF
--- a/google/cloud/spanner/internal/session.h
+++ b/google/cloud/spanner/internal/session.h
@@ -18,6 +18,7 @@
 #include "google/cloud/spanner/internal/channel.h"
 #include "google/cloud/spanner/version.h"
 #include <atomic>
+#include <chrono>
 #include <memory>
 #include <string>
 #include <utility>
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/internal/session.h
```diff
@@ -53,12 +55,23 @@ class Session {
...
+  // Give `SessionPool` access to the private methods below.
+  friend class SessionPool;
   std::shared_ptr<Channel> const& channel() const { return channel_; }
 
+  // The caller is responsible for ensuring these methods are used in a
+  // thread-safe manner (i.e. using external locking).
+  std::chrono::steady_clock::time_point last_use_time() const {
+    return last_use_time_;
+  }
+  void reset_last_use_time() {
```
devbww: Nit: "reset" smells like zero/clear.  Perhaps something like `update_last_use_time()`?

devjgm:
> You can probably get away with a different source for `now`, but because the change is so small this is a wild guess on my part. Let's see what happens in the next PR.

+1. I think creating a Clock class/interface is going to be overkill. I think we only need a something that can return a "now", like `std::function<std::chrono::steady_clock::time_point()> now` that can be "injected".

PR #1358, opened 2020-03-11
feat: add Managed Backup feature

tmatsuo:
This is a squashed commit for Spanner Managed backup feature.  The code is once reviewed internally.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1358)


tmatsuo:
The code coverage is low probably because I didn't add unit tests for database_admin_client. Adding them now.

google/cloud/spanner/integration_tests/database_admin_integration_test.cc
```diff
@@ -165,6 +169,103 @@ TEST(DatabaseAdminClient, DatabaseBasicCRUD) {
   db_list = get_current_databases();
   ASSERT_EQ(1, std::count(db_list.begin(), db_list.end(), db.FullName()));
 
+  // Tests for Backup are taking long time. To run them, set
+  // RUN_SLOW_INTEGRATION_TESTS environment variable to yes.
+  if (run_slow_integration_tests == "yes") {
+    auto backup_future = client.CreateBackup(
```
devbww: You probably want to check if these need a `!emulator` guard.

google/cloud/spanner/samples/samples.cc
```diff
@@ -2301,6 +2573,49 @@ void RunAll(bool emulator) {
   std::cout << "\nRunning spanner_delete_data sample\n";
   DeleteData(client);
 
+  if (run_slow_integration_tests == "yes") {
+    std::string backup_id =
+        google::cloud::spanner_testing::RandomBackupName(generator);
+
+    std::cout << "\nRunning spanner_create_backup sample\n";
+    CreateBackup(database_admin_client, project_id, instance_id, database_id,
```
devbww: You probably want to check if these need a `!emulator` guard.

google/cloud/spanner/integration_tests/database_admin_integration_test.cc
tmatsuo: Thanks Added the guard.

google/cloud/spanner/samples/samples.cc
tmatsuo: Thanks, added too

tmatsuo:
Oops, `DatabaseAdminClientTest.CreateBackupCancel` is not fixed yet. I may have misunderstood why it's failing.

tmatsuo:
It's weird, but with a longer maximum delay, the test became stable.

tmatsuo:
Hmm I can not reproduce the failure with `RUNS_PER_TEST=100`, but apparently this doesn't solve the issue on kokoro.
Now I'm trying to explicitly verify the expectations on the mock. This version runs green 100 times locally.

tmatsuo:
I ended up trying to suppress the leak.

devjgm:
> The API checks have a false positive, but also a real API breakage. We should consider using non-pure virtuals for the new functions in `DatabaseAdminConnection` /cc: @devjgm FYI

I'm fine w/ the use of non-pure virtual functions as you have them now. 

That said, I'll offer an opinion that I think it's a shame that all the other methods are pure virtual and these new ones are not. That leaves a seemingly arbitrary difference in the code that may be confusing to readers. Why are some methods pure and others not? Yes, the reason is to avoid breaking folks who subclass the connection, but really, I think that's a usecase that we'd rather not even support.

I think the only reason we have virtual methods at all is to enable gmock. And in that case, we provide our own mock subclass that users should use. And if users are using our mock connection, then the addition of a new pure-virtual method in the connection base class wouldn't break anyone.

So, one possible alternative is that we could document our "connection" interfaces saying that they are *NOT* intended for users to create their own arbitrary subclasses. They exist for the purpose of using gmock, and callers should use our provided gmock subclasses in their own tests. This will avoid them breaking as we add new methods.

coryan:
> > The API checks have a false positive, but also a real API breakage. We should consider using non-pure virtuals for the new functions in `DatabaseAdminConnection` /cc: @devjgm FYI
> 
> I'm fine w/ the use of non-pure virtual functions as you have them now.

I think we should move this discussion to a bug or feature request, as we both seem to be Okay with the code as-is.

> 
> That said, I'll offer an opinion that I think it's a shame that all the other methods are pure virtual and these new ones are not.

Agree.

> That leaves a seemingly arbitrary difference in the code that may be confusing to readers. Why are some methods pure and others not? Yes, the reason is to avoid breaking folks who subclass the connection, but really, I think that's a usecase that we'd rather not even support.
> 
> I think the only reason we have virtual methods at all is to enable gmock.

Or any other mocking tool presumably?  Applications may use FakeIt:

https://github.com/eranpeer/FakeIt

or FSeam:

http://freeyoursoul.online/fseam-a-mocking-framework-that-requires-no-change-in-code-part-2/

or something else we do not know about.  They may have also decorated a `Connection` to integrate with OpenCensus, or OpenTelemetry (I have a hunch this is the perfect place to inject spans).

All of which is to say: we do not know how things are used, and I think there are legitimate uses for inheriting from this classes, a blanket prohibition seems less than ideal.

PR #1361, opened 2020-03-12
feat: SqlStatement uses `op<<(Value)`

devjgm:
Nobody should use `PrintTo(Value, os)`. This PR documents that users
should not call `PrintTo` and that it will be removed in an upcoming
release. All streaming of `Value` objects should be done using
`operator<<`.

Note this PR changes the output format of `SqlStatement`, and it
documents that the output format is for human consumption only and may
change without notice in the future (same as is documented for Value).

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1361)


google/cloud/spanner/value.h
```diff
@@ -305,20 +305,11 @@ class Value {
...
-   *   machine consumption as the output format may change without notice.
-   */
-  friend void PrintTo(Value const& v, std::ostream* os);
-
   /**
    * Outputs string representation of a given Value to the provided stream.
    *
    * @warning This is intended for debugging and human consumption only, not
-   *   machine consumption as the output format may change without notice.
+   *     machine consumption as the output format may change without notice.
```
devbww: Nit: Now that you're editing this, you might as well put a comma after "consumption".

google/cloud/spanner/sql_statement.h
```diff
@@ -103,8 +103,15 @@ class SqlStatement {
     return !(a == b);
   }
 
- private:
+  /**
+   * Outputs a string representation of the given @p stmt to the given @p os.
+   *
+   * @warning This is intended for debugging and human consumption only, not
+   *     machine consumption as the output format may change without notice.
```
devbww: See the other comment about a comma after "consumption".

google/cloud/spanner/value.h
devjgm: Done.

google/cloud/spanner/sql_statement.h
devjgm: Done

PR #1362, opened 2020-03-12
feat: Value streaming respects io manipulators

devjgm:
This PR changes the `Value` `op<<` operator for the types `bool`,
`int64_t`, and `double` so that they print by default as they would as
normal C++ types. This also enables them to work with C++ IO stream
manipulators. For example, this change makes the following work:

```cc
std::cout << std::hex << Value(42);  // Output: 2a
```

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1362)

### DIFF
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -96,21 +96,20 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
 
   switch (t.code()) {
     case google::spanner::v1::BOOL:
-      return os << (v.bool_value() ? "TRUE" : "FALSE");
+      return os << v.bool_value();
 
-    case google::spanner::v1::TIMESTAMP:
-    case google::spanner::v1::DATE:
     case google::spanner::v1::INT64:
-      return os << v.string_value();
+      return os << internal::FromProto(t, v).get<std::int64_t>().value();
 
     case google::spanner::v1::FLOAT64:
-      if (v.kind_case() == google::protobuf::Value::kStringValue) {
-        return os << v.string_value();
-      }
-      return os << v.number_value();
+      return os << internal::FromProto(t, v).get<double>().value();
 
     case google::spanner::v1::STRING:
       return os << "\"" << v.string_value() << "\"";
+
+    case google::spanner::v1::TIMESTAMP:
+    case google::spanner::v1::DATE:
+      return os << v.string_value();
 
     case google::spanner::v1::BYTES:
       return os
@@ -144,10 +143,8 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
       return os << ')';
     }
 
-    case google::spanner::v1::TYPE_CODE_UNSPECIFIED:
-    case google::spanner::v1::TypeCode_INT_MIN_SENTINEL_DO_NOT_USE_:
-    case google::spanner::v1::TypeCode_INT_MAX_SENTINEL_DO_NOT_USE_:
-      break;
+    default:
+      return os << "Error: unknown value type code " << t.code();
   }
   return os;
 }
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/value_test.cc
```diff
@@ -937,8 +938,17 @@ TEST(Value, CommitTimestamp) {
...
+//
+// 1. The test case's name. This is for gtest and needs to be unique.
+// 2. The `Value` instance to stringify.
+// 3. The expect stringified string value
+// 4. An optional lambda (nullptr is allowed) that may apply IO manipulators to
+//    the ostream that may affect the rendered output of the stringified `Value.
 class ValueStreamOperator
-    : public TestWithParam<std::tuple<std::string, Value, std::string>> {
+    : public TestWithParam<std::tuple<std::string, Value, std::string,
+                                      std::function<void(std::ostream&)>>> {
```
devbww: Might it be nicer to make this non-nullable and change to `std::function<std::ostream(std::ostream&)>`?

google/cloud/spanner/value.cc
```diff
@@ -144,10 +143,8 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
       return os << ')';
     }
 
-    case google::spanner::v1::TYPE_CODE_UNSPECIFIED:
-    case google::spanner::v1::TypeCode_INT_MIN_SENTINEL_DO_NOT_USE_:
-    case google::spanner::v1::TypeCode_INT_MAX_SENTINEL_DO_NOT_USE_:
-      break;
+    default:
```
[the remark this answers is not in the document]
devjgm: Possibly. But I don't think we should be mentioning proto fields w/ "DO_NOT_USE" in the name.

google/cloud/spanner/value_test.cc
devjgm: Done.

google/cloud/spanner/value_test.cc
```diff
@@ -937,101 +936,124 @@ TEST(Value, CommitTimestamp) {
...
+  };
+  auto const boolalpha = [](std::ostream& os) -> std::ostream& {
+    return os << std::boolalpha;
+  };
+  auto const float4 = [](std::ostream& os) -> std::ostream& {
+    return os << std::showpoint << std::setprecision(4);
+  };
+  auto const alphahex = [](std::ostream& os) -> std::ostream& {
+    return os << std::boolalpha << std::hex;
+  };
```
devbww: Possibly
```
  auto const alphahex = [&hex, &boolalpha](std::ostream& os) -> std::ostream& {
    return hex(boolalpha(os));
  };
```
to localize the particular variety of bool and hex formatting to one place?

google/cloud/spanner/value_test.cc
```diff
@@ -937,101 +936,124 @@ TEST(Value, CommitTimestamp) {
...
+      {MakeNullValue<Bytes>(), "NULL", normal},
+      {MakeNullValue<Date>(), "NULL", normal},
+      {MakeNullValue<Timestamp>(), "NULL", normal},
+
+      // Tests arrays
+      {Value(std::vector<bool>{false, true}), "[0, 1]", normal},
+      {Value(std::vector<bool>{false, true}), "[false, true]", boolalpha},
+      {Value(std::vector<std::int64_t>{10, 11}), "[10, 11]", normal},
+      {Value(std::vector<std::int64_t>{10, 11}), "[a, b]", hex},
+      {Value(std::vector<double>{1.0, 2.0}), "[1, 2]", normal},
```
devbww: Add a `float4` case?

devjgm: Done.

google/cloud/spanner/value_test.cc
devjgm: Unless you feel strongly, I think the current code is simpler and clearer and would prefer to keep it.

PR #1371, opened 2020-03-13
feat: escape non-printable bytes in Value::op<<

devjgm:
Non-printable `Byte` values are now printed using their octal escape sequence. We don't use hex escape sequences because in C++ `"\x41B"` is a 3-digit hex number, not 2-digits.

We might consider escaping non-printable string characters too. I didn't do that here because it's not clear that we want that feature since normal `std::string` objects don't do that.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1371)

### DIFF
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -15,14 +15,14 @@
 #include "google/cloud/spanner/value.h"
 #include "google/cloud/spanner/internal/date.h"
 #include "google/cloud/log.h"
+#include <array>
 #include <cctype>
 #include <cerrno>
 #include <cmath>
+#include <cstdio>
 #include <cstdlib>
 #include <cstring>
-#include <iomanip>
 #include <ios>
-#include <sstream>
 #include <string>
 
 namespace google {
@@ -117,20 +117,19 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
     case google::spanner::v1::BYTES: {
       auto const bytes = internal::BytesFromBase64(v.string_value())
                              ->get<std::vector<unsigned char>>();
-      // Uses a separate stream so we don't mess up the formatting configured
-      // on the caller-provided `os` stream. The format used is B"...", where
-      // printable characters are included normally, and unprintable characters
-      // are printed in octal like "\011".
-      std::ostringstream ss;
-      ss << std::setfill('0') << std::oct;
-      for (char const byte : bytes) {
-        if (!std::isprint(byte)) {
-          ss << R"(\)" << std::setw(3) << static_cast<int>(byte);
+      os << R"(B")";
+      for (auto const byte : bytes) {
+        if (std::isprint(byte) != 0) {
+          os << byte;
         } else {
-          ss << byte;
+          // This uses snprintf rather than iomanip so we don't mess up the
+          // formatting on `os` for other streaming operations.
+          std::array<char, sizeof(R"(\000)")> buf;
+          auto n = std::snprintf(&buf[0], buf.size(), R"(\%03o)", byte);
+          if (n == buf.size() - 1) os << buf.data();
         }
       }
-      return os << R"(B")" << ss.str() << R"(")";
+      return os << R"(")";
     }
 
     case google::spanner::v1::ARRAY: {
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -126,7 +126,7 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
           // formatting on `os` for other streaming operations.
           std::array<char, sizeof(R"(\000)")> buf;
           auto n = std::snprintf(&buf[0], buf.size(), R"(\%03o)", byte);
-          if (n == buf.size() - 1) os << buf.data();
+          if (n == static_cast<int>(buf.size() - 1)) os << buf.data();
         }
       }
       return os << R"(")";
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -125,7 +125,7 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
           // This uses snprintf rather than iomanip so we don't mess up the
           // formatting on `os` for other streaming operations.
           std::array<char, sizeof(R"(\000)")> buf;
-          auto n = std::snprintf(&buf[0], buf.size(), R"(\%03o)", byte);
+          auto n = std::snprintf(buf.data(), buf.size(), R"(\%03o)", byte);
           if (n == static_cast<int>(buf.size() - 1)) os << buf.data();
         }
       }
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -119,7 +119,7 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
                              ->get<std::vector<unsigned char>>();
       os << R"(B")";
       for (auto const byte : bytes) {
-        if (std::isprint(byte) != 0) {
+        if (std::isprint(byte)) {
           os << byte;
         } else {
           // This uses snprintf rather than iomanip so we don't mess up the
--- a/google/cloud/spanner/value.cc
+++ b/google/cloud/spanner/value.cc
@@ -15,11 +15,14 @@
 #include "google/cloud/spanner/value.h"
 #include "google/cloud/spanner/internal/date.h"
 #include "google/cloud/log.h"
+#include <cctype>
 #include <cerrno>
 #include <cmath>
 #include <cstdlib>
 #include <cstring>
+#include <iomanip>
 #include <ios>
+#include <sstream>
 #include <string>
 
 namespace google {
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/value.cc
```diff
@@ -111,11 +159,15 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
     case google::spanner::v1::DATE:
       return os << v.string_value();
 
-    case google::spanner::v1::BYTES:
-      return os
-             << "B\""
-             << internal::BytesFromBase64(v.string_value())->get<std::string>()
-             << "\"";
+    case google::spanner::v1::BYTES: {
```
devbww: Should we do the same escaping for `google::spanner::v1::STRING`?

google/cloud/spanner/value_test.cc
```diff
@@ -1056,6 +1068,55 @@ TEST(Value, OutputStream) {
...
+        expected << "\\v";
+        break;
+      case '\f':
+        expected << "\\f";
+        break;
+      case '\r':
+        expected << "\\r";
+        break;
+      default:
+        if (std::isprint(b)) {
```
devbww: `#include <cctype>`

google/cloud/spanner/value_test.cc
```diff
@@ -1056,6 +1068,55 @@ TEST(Value, OutputStream) {
...
+          expected << b;
+        } else {
+          expected << std::setfill('0') << std::oct;
```
devbww: `#include <iomanip>`

google/cloud/spanner/value.cc
```diff
@@ -88,6 +90,52 @@ bool Equal(google::spanner::v1::Type const& pt1,
   }
 }
 
+// A 256-element table mapping all `unsigned char` byte values to a printable
+// string. All naturally printable values are used directly. All non-printable
+// bytes use their octal escape sequence. Whitespace characters use their short
+// form like '\t' and '\n'.
```
devbww: I don't think "whitespace characters" is an accurate characterization here.  It doesn't include SPACE for example, and I wouldn't say that `\a` counts as whitespace.

google/cloud/spanner/value.cc
```diff
@@ -88,6 +90,52 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+using ByteType = unsigned char;
+constexpr auto kByteTableSize = std::numeric_limits<ByteType>::max() + 1;
+constexpr std::array<char const*, kByteTableSize> kByteTable = {
+    "\\000", "\\001", "\\002", "\\003", "\\004", "\\005", "\\006", "\\a",
+    "\\b",   "\\t",   "\\n",   "\\v",   "\\f",   "\\r",   "\\016", "\\017",
+    "\\020", "\\021", "\\022", "\\023", "\\024", "\\025", "\\026", "\\027",
+    "\\030", "\\031", "\\032", "\\033", "\\034", "\\035", "\\036", "\\037",
+    " ",     "!",     "\"",    "#",     "$",     "%",     "&",     "'",
```
devbww: Should not `"\""` be `"\\\""`?  (Another reason why the short-form escapes aren't just whitespace.)

google/cloud/spanner/value.cc
```diff
@@ -88,6 +90,52 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
 }
 
+// A 256-element table mapping all `unsigned char` byte values to a printable
+// string. All naturally printable values are used directly. All non-printable
+// bytes use their octal escape sequence. Whitespace characters use their short
+// form like '\t' and '\n'.
+using ByteType = unsigned char;
+constexpr auto kByteTableSize = std::numeric_limits<ByteType>::max() + 1;
+constexpr std::array<char const*, kByteTableSize> kByteTable = {
+    "\\000", "\\001", "\\002", "\\003", "\\004", "\\005", "\\006", "\\a",
```
coryan: Do we care if the application is using a non-ASCII locale?

.clang-tidy
```diff
@@ -73,3 +73,5 @@ CheckOptions:
   - { key: readability-identifier-naming.MemberConstantPrefix,     value: k         }
   - { key: readability-identifier-naming.StaticConstantCase,       value: CamelCase }
   - { key: readability-identifier-naming.StaticConstantPrefix,     value: k         }
+  - { key: readability-implicit-bool-conversion.AllowIntegerConditions,  value: 1   }
+  - { key: readability-implicit-bool-conversion.AllowPointerConditions,  value: 1   }
```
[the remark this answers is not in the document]
devjgm: I'll make that change if you really think `if (std::isprint(c) != 0)` is clearer than `if (std::isprint(c))`. I think the latter is clearer. Even the https://en.cppreference.com/w/cpp/string/byte/isprint documentation for isprint shows two examples where the return from `isprint` is implicitly converted to bool.

google/cloud/spanner/value.cc
devjgm: Fixed.

google/cloud/spanner/value.cc
devjgm: Maybe. Can you say more about what that'd affect here? Is there something different you'd recommend?

google/cloud/spanner/value.cc
```diff
@@ -88,6 +90,52 @@ bool Equal(google::spanner::v1::Type const& pt1,
...
+    "\\340", "\\341", "\\342", "\\343", "\\344", "\\345", "\\346", "\\347",
+    "\\350", "\\351", "\\352", "\\353", "\\354", "\\355", "\\356", "\\357",
+    "\\360", "\\361", "\\362", "\\363", "\\364", "\\365", "\\366", "\\367",
+    "\\370", "\\371", "\\372", "\\373", "\\374", "\\375", "\\376", "\\377",
+};
+
+// We'll index into kByteTable using values of type ByteType, so assert that
+// this will never fail.
+static_assert(std::numeric_limits<ByteType>::min() == 0, "");
+static_assert(std::numeric_limits<ByteType>::max() < kByteTable.size(), "");
```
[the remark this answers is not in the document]
devjgm: Removed

google/cloud/spanner/value.cc
devjgm: I mentioned this in the PR description. No, I do not think so. I think we should treat STRING more like std::string, which doesn't escape. Bytes is special because it (a) doesn't have an example C++ counterpart, and (b) is more likely to contain non-printable characters.

google/cloud/spanner/value.cc
devjgm: Done

google/cloud/spanner/value.cc
devjgm: So, it looks like we could handle other locales if we compute the "isprint" at runtime. That means, we'd run a little code instead of using a lookup table. That'd be fine with me. Do you have a preference?

.clang-tidy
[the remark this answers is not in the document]
coryan: Is not that I like `if (std::isprint(c) != 0)` is that I dislike `if (something_that_was_not_intended_as_as_bool())`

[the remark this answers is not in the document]
devjgm: I'll move this clang-tidy change to another PR so it doesn't slow down this one.

google/cloud/spanner/value.cc
devjgm: I rolled back to an earlier version of this PR where I use `std::isprint` at runtime, which is what I believe you'res suggesting. I think this should now work with non-ASCII locales.

devjgm:
PTAL

devjgm:
All comments addressed. PTAL.

google/cloud/spanner/value.cc
devbww: I'm sorry I missed the bit in the PR description.  I still dissent though.  `std::string` also doesn't wrap its output with quotes.  Rendering a `STRING` value is fundamentally different to printing a string of characters.  But, so be it.

devjgm:
> I would still contend that just using if (std::isprint(c) != 0) here is the lesser compromise over permitting implicit integer/pointer-to-bool conversions everywhere, but I see it is too late now.

Note that the clang-tidy change does not allow *implicit* interger/pointer-to-bool conversions *anywhere*. It only allows *explicit* integer/poiner conversions within the *explicit bool* context of the condition in an `if` of loop.

google/cloud/spanner/value.cc
```diff
@@ -111,11 +114,23 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
...
-             << "B\""
-             << internal::BytesFromBase64(v.string_value())->get<std::string>()
-             << "\"";
+    case google::spanner::v1::BYTES: {
+      auto const bytes = internal::BytesFromBase64(v.string_value())
+                             ->get<std::vector<unsigned char>>();
+      os << R"(B")";
+      for (auto const byte : bytes) {
+        if (std::isprint(byte)) {
+          os << byte;
```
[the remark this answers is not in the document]
devjgm: Good point. Yes, I think you're right. Done.

google/cloud/spanner/value_test.cc
```diff
@@ -988,6 +988,18 @@ TEST(Value, OutputStream) {
...
+      {Value(Bytes(std::string("ab\011Z"))), R"(B"ab\011Z")", normal},
+      {Value(Bytes(std::string("ab\011Z"))), R"(B"ab\011Z")", hex},
+      {Value(Bytes(std::string("ab\001Z\0211"))), R"(B"ab\001Z\0211")", normal},
+      {Value(Bytes(std::string("ab\001Z\0211"))), R"(B"ab\001Z\0211")", hex},
+      {Value(Bytes(std::string(3, '\0'))), R"(B"\000\000\000")", normal},
+      {Value(Bytes(std::string(3, '\0'))), R"(B"\000\000\000")", hex},
+      {Value(Bytes(std::string("!@#$%^&*()-."))), R"(B"!@#$%^&*()-.")", normal},
+      {Value(Bytes(std::string("!@#$%^&*()-."))), R"(B"!@#$%^&*()-.")", hex},
+      {Value(Bytes("foo")), R"(B"foo\000")", normal},
+      {Value(Bytes("foo")), R"(B"foo\000")", hex},
```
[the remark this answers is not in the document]
devjgm: Unless someone can suggest a fully locale-safe way to test this, I think we should not care if the *unit test* relies on some particular locale. The non-test code seems to be locale agnostic, so the actual code should work, even if someone runs it on a non-ascii machine. If *our* the *unit tests* fail to pass on that non-ascii machine, I'm not sure I care. Or maybe I'll care if that ever happens, which I think is unlikely.

google/cloud/spanner/value.cc
```diff
@@ -111,11 +114,23 @@ std::ostream& StreamHelper(std::ostream& os, google::protobuf::Value const& v,
...
+        } else {
+          // This uses snprintf rather than iomanip so we don't mess up the
+          // formatting on `os` for other streaming operations.
+          std::array<char, sizeof(R"(\000)")> buf;
+          auto n = std::snprintf(buf.data(), buf.size(), R"(\%03o)", byte);
+          if (n == static_cast<int>(buf.size() - 1)) os << buf.data();
```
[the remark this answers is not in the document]
devjgm: I don't think it can fail. But I added the else and a unit test for it.

google/cloud/spanner/value.cc
devjgm: This PR is about escaping bytes. Let's move the STRING discussion elsewhere. I filed an issue to track it: https://mail.google.com/mail/u/0/#inbox/FMfcgxwHMPfbQplmGvnhKcQcpzvvsnWM

PR #1383, opened 2020-03-15
fix: the Bytes base64 decoder assumes ASCII execution character set

devbww:
Assert that the execution character set looks like ASCII, as required
by the `kCharToIndexExcessOne[]` lookup table in the base64 decoder.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1383)

### DIFF
--- a/google/cloud/spanner/bytes.cc
+++ b/google/cloud/spanner/bytes.cc
@@ -48,10 +48,13 @@ constexpr std::array<unsigned char, UCHAR_MAX + 1> kCharToIndexExcessOne = {{
     38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52,
 }};
 
+// kCharToIndexExcessOne[] assumes an ASCII execution character set.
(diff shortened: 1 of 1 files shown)


google/cloud/spanner/bytes.cc
```diff
@@ -48,10 +48,13 @@ constexpr std::array<unsigned char, UCHAR_MAX + 1> kCharToIndexExcessOne = {{
     38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52,
 }};
 
+// kCharToIndexExcessOne[] assumes an ASCII execution character set.
+static_assert('A' == 65, "required by base64 decoder");
```
devbww: Note: As far as I can determine, a `static_assert()` must be evaluated in the "execution environment".  For example, if you asserted on the size of a type, it is the execution type you're talking about.  So, the result of this assertion should be affected by something like `-fexec-charset` ... the character set we're compiling for.

PR #1384, opened 2020-03-15
refactor: move Bytes formatting to bytes.{h,cc}

devjgm:
This makes it so that `Value::op<<` can just stream the `Bytes`
directly, and `Value` is no longer responsible for the
formatting/escaping of bytes.

This makes it so that if a user has a `Bytes` object (not a
`Value`), they can still stream it. It's also more efficient this way because we can
directly iterate using the `Bytes::Decoder` rather than having to first copy the
bytes to an unnecessary `vector<unsigned char>`.

There's no change in formatting logic or behavior in this PR. Just a
refactoring/move of the code from value.cc -> bytes.cc.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1384)

### DIFF
--- a/google/cloud/spanner/bytes.cc
+++ b/google/cloud/spanner/bytes.cc
@@ -15,6 +15,7 @@
 #include "google/cloud/spanner/bytes.h"
 #include "google/cloud/status.h"
 #include <array>
+#include <cctype>
 #include <climits>
 #include <cstdio>
 
@@ -78,7 +79,8 @@ std::ostream& operator<<(std::ostream& os, Bytes const& bytes) {
       }
     }
   }
-  return os << R"(")";
+  // Can't use raw string literal here because of a doxygen bug.
+  return os << "\"";
 }
 
 void Bytes::Encoder::Flush() {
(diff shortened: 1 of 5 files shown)


google/cloud/spanner/bytes.cc
```diff
@@ -55,6 +56,31 @@ static_assert(UCHAR_MAX == 255, "required by base64 encoder");
...
 
+// Prints the bytes in the form B"...", where printable bytes are output
+// normally, double quotes are backslash escaped, and non-printable characters
+// are printed as a 3-digit octal escape sequence.
+std::ostream& operator<<(std::ostream& os, Bytes const& bytes) {
+  os << R"(B")";
+  for (auto const byte : Bytes::Decoder(bytes.base64_rep_)) {
+    if (byte == '"') {
+      os << R"(\")";
+    } else if (std::isprint(byte)) {
```
devbww: `#include <cctype>`

devjgm: Fixed. I don't know how you're so good at spotting missing includes.

devbww: Who knows, but it probably means the one in values.cc is no longer needed.

PR #1385, opened 2020-03-15
feat: taught Date how to stream itself

devjgm:
Date was the only non-aggregate "Value" type that didn't already know
how to print itself. Everything else, (bool, int64_t, double,
std::string, Bytes, and Timestamp) were all streamable directly.
`internal/date.h` already had this support, so it just took a little
plumbing to make the user-facing `Date` class to itself also be
streamable.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1385)

### DIFF
--- a/google/cloud/spanner/date.h
+++ b/google/cloud/spanner/date.h
@@ -40,6 +40,9 @@ class Date {
   int month() const { return month_; }
   int day() const { return day_; }
 
+  /// @name Output streaming
+  friend std::ostream& operator<<(std::ostream& os, Date const& date);
+
  private:
   std::int64_t year_;
   int month_;
--- a/google/cloud/spanner/date.h
+++ b/google/cloud/spanner/date.h
@@ -41,14 +41,13 @@ class Date {
   int month() const { return month_; }
   int day() const { return day_; }
 
-  /// @name Output streaming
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/date.h
```diff
@@ -40,6 +41,9 @@ class Date {
   int month() const { return month_; }
   int day() const { return day_; }
 
+  /// @name Output streaming
+  friend std::ostream& operator<<(std::ostream& os, Date const& date);
```
devbww: Why does it need to be a friend?

devjgm: It doesn't technically need to be a friend. But the reasons why I thought it made the most sense declare it as an inline friend like this are:

1. All other `op<<` declarations in our code are inline friends, so making this one the same seems nice.

2. There are other (minor) benefits to inline friend declarations, such as they can *only* be found via ADL (and therefore work well with operators), they don't add to overloads in the namespace thus making compile-times faster, etc.

But if you'd prefer this be an out-of-line non-friend, I'll change it.

devjgm: Nevermind. I removed the friendship.

devbww: And if you'd prefer to keep it, that's fine too.  But the relational operators should probably match.

Issue #1296, opened 2020-02-26
Consider supporting include with angle brackets in Bazel.


coryan:
Some Bazel users may want to use `#include <google/cloud/spanner/client.h>` (note the angle brackes), which requires adding a `includes=(["."])` directive to our Bazel commands. This should be optional because some users of Bazel do not like angle brackets.

devjgm:
A few comments. Since our BUILD file lives in `google/cloud/spanner/BUILD`, the option we'd need to add to our `cc_library` build rule is `includes = ["../../.."],`. That would result in a `-I <...>` option being added to all dependents of our build rule so they can treat our whole project root as a system include directory.

The problem w/ this is that `../../..` resolves to our project root, which has our `WORKSPACE` file, and bazel doesn't like this, giving the following error:

```
ERROR: /v/google/cloud/spanner/BUILD:37:1: in includes attribute of cc_library rule //google/cloud/spanner:spanner_client: '../../../' resolves to the workspace root, which would allow this rule and all of its transitive dependents to include any file in your workspace. Please include only what you need
```

The reason this works for other projects like googletest, grpc, and protobuf is because their code lives in their repo at a different path than the one they want users to include. Usually, they put their code in a top-level `src/` dir.

I *think* they use the `includes = ...` option in their build rule to make the code compile at all. I doubt their goal is to make angle-bracket includes work (I suspect that's just a side effect). Note: including all the googletest, grpc, and protobuf includes with double quotes works too.

[Bazel's recommendation with include path](https://docs.bazel.build/versions/master/bazel-and-cpp.html#include-paths) is to use double quotes, not angle brackets.

# Options

Here are some options.

1. Not support angle-bracket includes if using bazel. This matche's Bazel's advice, and our code works with this policy today.

2. Create a top-level `src/` directory and `mv google/ src/`. This should allow us to use the `includes = ["../../.."]` trick in our BUILD files. However, it will 

3. Create a top-level BUILD file w/ `cc_library` target that users are supposed to use. This will let us add the `includes = ["."]` option that will let users include using quotes or angle brackets. This also has the nice feature of providing users with a much simpler user-facing BUILD file that will contain all of the build targets that they're allowed to use. A draft/example of this is https://github.com/googleapis/google-cloud-cpp-spanner/pull/1396

devjgm:
Fixed in https://github.com/googleapis/google-cloud-cpp-spanner/pull/1396

PR #1396, opened 2020-03-16
feat: adds BUILD target to top-level BUILD file

devjgm:
Adds user-facing bazel build targets to the project's top-level BUILD
file. This file should list *all* of the BUILD targets that users are
allowed to use. In a future PR, visibility restrictions should prohibit
users from reaching down into `google/cloud/spanner/BUILD`.

This change will make things simpler for users since they'll have a
single, uncomplicated, top-level BUILD file that contains all the build
targets that they're allowed to use. It also gives us a place to put an
`includes = ["."]` directive, which will allow Bazel users to include
our headers using either double quotes or angle brackets.

Related to #1296

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1396)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -19,3 +19,11 @@ licenses(["notice"])  # Apache 2.0
 exports_files([
     "LICENSE",
 ])
+
+cc_library(
+    name = "spanner_client",
+    includes = ["."],
+    deps = [
+        "//google/cloud/spanner:spanner_client",
+    ],
+)
--- a/google/cloud/spanner/BUILD
+++ b/google/cloud/spanner/BUILD
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# TODO(#XXX) Change this visibility to "//:__subpackages__" so that users are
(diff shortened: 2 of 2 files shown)


devjgm:
Note: a followup task for me after this PR is included in the next release is to update quickstart/ to use this new build target and using angle-bracket includes.
https://github.com/googleapis/google-cloud-cpp-spanner/issues/1416

mr-salty:
FYI after this change I now get the following error.  I can work around it by doing `bazel build google/...` but still seems bad.

```
$ bazel build ...
ERROR: /home/salty/git/delete/BUILD:23:1: in includes attribute of cc_library rule //:spanner_client: '.' resolves to the workspace root, which would allow this rule and all of its transitive dependents to include any file in your workspace. Please include only what you need
```

devjgm:
Hmm.. ok, let me think about how best to fix that or whether to roll this back.

One fix would be to add `tags = ["manual"]` to the //:spanner_client build target.

PR #1397, opened 2020-03-17
fix: avoid deadlocks during shutdown

coryan:
Avoid deadlocks caused by the SessionPool both controlling the
lifetime of the background threads in its destructor, and being also
destructed by those threads. The background threads are now owned by
the ConnectionImpl.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1397)


coryan:
I am happy to write the `SesssionPool::Shutdown()` stuff if you want.

PR #1400, opened 2020-03-17
ci: run benchmarks in CMake-based builds

coryan:
The benchmarks were not declared "tests" so they would not run as
part as the regular CMake builds. We may not have strict testing
criteria for them, but at least we should run them to avoid code rot.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1400)

### DIFF
--- a/google/cloud/spanner/CMakeLists.txt
+++ b/google/cloud/spanner/CMakeLists.txt
@@ -424,6 +424,7 @@ function (spanner_client_define_benchmarks)
         string(REPLACE "/" "_" target ${fname})
         string(REPLACE ".cc" "" target ${target})
         add_executable(${target} ${fname})
+        add_test(NAME ${target} COMMAND ${target})
         target_link_libraries(${target} PRIVATE googleapis-c++::spanner_client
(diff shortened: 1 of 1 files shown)


devbww:
> but at least we should run them to avoid code rot.

Why would the code rot?  Were not they being built?

My guess is there will be a future issue about not running benchmarks during testing.  If we don't look at the results, it is just a waste of time.  At least, that's why I chose that path in the first place.

coryan:
> > but at least we should run them to avoid code rot.
> 
> Why would the code rot? Were not they being built?

It could start crashing... I don't know how, but why bother thinking?

> 
> My guess is there will be a future issue about not running benchmarks during testing. If we don't look at the results, it is just a waste of time. At least, that's why I chose that path in the first place.

If we don't look at the results, why have them in the code?

Frankly, I don't like to wade through the code coverage resports and have to remember that "oh, yeah, we intentionally don't run the benchmarks for CMake, but we do for Bazel, and I should not think about this code."

Issue #1306, opened 2020-02-27
Create a page describing the environment variables for the library.


coryan:
We have quite a few of them, for example:

- `GOOGLE_CLOUD_CPP_ENABLE_CLOG=yes` turns on logging in the library, basically the library always "logs" but the logging infrastructure has no backend to actually print anything until the application sets a backend or they set this environment variable.

- `GOOGLE_CLOUD_CPP_ENABLE_TRACING=rpc` turns on tracing for all gRPC calls, the library injects an additional `Stub` decorator that prints each gRPC request and response.

- `GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT` changes the default endpoint (`spanner.googleapis.com`) for the spanner client library, useful for debugging.

- `SPANNER_EMULATOR_HOST` changes the default endpoints and uses insecure credentials.

- `GOOGLE_CLOUD_PROJECT` is used in examples and integration tests to configure the GCP project.

- `GOOGLE_CLOUD_CPP_SPANNER_INSTANCE` is used in examples and integration tests to, well, configure the spanner instance.

There are a few others used to control the integration tests that probably should go undocumented.

devbww:
> `GOOGLE_CLOUD_CPP_ENABLE_TRACING=rpc` turns on tracing for all gRPC calls

We also look at `GOOGLE_CLOUD_CPP_ENABLE_TRACING=rpc-streams` for `Read()`, `ExecuteQuery()`, and `ProfileQuery()`.

devjgm:
Is there a reason that `GOOGLE_CLOUD_CPP_ENABLE_TRACING=rpc-streams` is using "rpc-streams" rather than just "rpc"? Is this difference desirable? I'm wondering if we can just simplify this so that `GOOGLE_CLOUD_CPP_ENABLE_TRACING=rpc` enables tracing of *all* RPCs (as its name may seem to imply).

coryan:
> Is there a reason that `GOOGLE_CLOUD_CPP_ENABLE_TRACING=rpc-streams` is using "rpc-streams" rather than just "rpc"? Is this difference desirable?

I think so, but I can be convinced otherwise: you get *lots* of data when `rpc-streams` is turned on, but a lot less when it is not.  

> I'm wondering if we can just simplify this so that `GOOGLE_CLOUD_CPP_ENABLE_TRACING=rpc` enables tracing of _all_ RPCs (as its name may seem to imply).

[...]

PR #1404, opened 2020-03-17
feat: clean up main doxygen page

devjgm:
Much of the getting-started and compiling stuff I'm not sending the user
to the quickstart/ repo, which has code and instructions that we're
testing, and I think it's nice and simple.

Moved the info about authentication into the quickstart/README.md.

Added a summary of most of the environment variables that can configure
the library.
Fixes: #1306

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1404)

### DIFF
--- a/google/cloud/spanner/doc/spanner-main.dox
+++ b/google/cloud/spanner/doc/spanner-main.dox
@@ -69,6 +69,9 @@ behaviors in the library.
   `Client::ProfileQuery()`. This can produce a lot of output, so use with
   caution!
 
+- `GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT=...` changes the default endpoint
+  (spanner.googleapis.com) for the library.
+
 - `GOOGLE_CLOUD_PROJECT=...` is used in examples and integration tests to
   configure the GCP project.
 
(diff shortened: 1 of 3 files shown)


google/cloud/spanner/doc/spanner-main.dox
```diff
@@ -24,164 +24,65 @@ using the C++ client library.
...
+  tests to, well, configure the spanner instance.
 
-# You have to manually call the corresponding function for google-cloud-cpp and
-# gRPC:  https://github.com/bazelbuild/bazel/issues/1550
-load("@com_github_googleapis_google_cloud_cpp_common//bazel:google_cloud_cpp_deps.bzl", "google_cloud_cpp_deps")
+- `GOOGLE_CLOUD_CPP_SPANNER_IAM_TEST_SA=...` is used in examples and
+  integration tests to set the service account for testing IAM operations.
 
-google_cloud_cpp_deps()
+- `SPANNER_EMULATOR_HOST=host:port` tells the library to connect to the
```
devbww: And GOOGLE_CLOUD_CPP_SPANNER_DEFAULT_ENDPOINT, which is subtly diffferent.

devjgm: Done. Added.

FTR, I originally omitted that one on purpose because it seems like something that is only useful for us when testing w/ internal versions of Spanner. I don't think users would ever have a need to set it. But documenting it here is fine with me.

devjgm:
Here's a full size screenshot to preview the rendered doxygen page
dox-screenshot

PR #1271, opened 2020-02-19
feat!: add support for instance-specific endpoints

devbww:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1271)

PR #1405, opened 2020-03-17
feat: implement async session creation

mr-salty:
Also enable session creation in the background thread.

I'm planning to change the existing uses of synchronous creation to
async but that requires very substantial test changes. However, I did
run a couple experiments - one *only* using `AsyncBatchCreateSessions()`,
and one with that, and *only* calling `Grow()` from the background
thread (that too some hackery) and all of the integration tests,
benchmarks, and samples passed in both cases. So, I wanted to get this
out for review while I work on making the unit tests function with async
ops.

Part of #1271

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1405)

### DIFF
--- a/google/cloud/spanner/internal/session_pool.cc
+++ b/google/cloud/spanner/internal/session_pool.cc
@@ -23,6 +23,7 @@
 #include "google/cloud/status.h"
 #include <algorithm>
 #include <chrono>
+#include <cmath>
 #include <random>
 #include <thread>
 #include <vector>
@@ -121,7 +122,8 @@ void SessionPool::DoBackgroundWork() {
 // creating or deleting sessions as necessary.
 void SessionPool::MaintainPoolSize() {
   std::unique_lock<std::mutex> lk(mu_);
-  if (!create_in_progress_ && total_sessions_ < options_.min_sessions()) {
+  if (create_calls_in_progress_ == 0 &&
+      total_sessions_ < options_.min_sessions()) {
     Grow(lk, total_sessions_ - options_.min_sessions(),
          WaitForSessionAllocation::kNoWait);
   }
@@ -144,8 +146,6 @@ Status SessionPool::Grow(std::unique_lock<std::mutex>& lk,
     // Can't grow the pool since we're already at max size.
     return Status(StatusCode::kResourceExhausted, "session pool exhausted");
   }
-
-  create_in_progress_ = true;
 
   // Compute how many Sessions to create on each Channel, trying to keep the
   // number of Sessions on each channel equal.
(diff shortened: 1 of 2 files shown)


google/cloud/spanner/internal/session_pool.cc
```diff
@@ -23,6 +23,7 @@
 #include "google/cloud/status.h"
 #include <algorithm>
 #include <chrono>
+#include <cmath>
```
devbww: Is this related?

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -187,30 +187,27 @@ Status SessionPool::Grow(std::unique_lock<std::mutex>& lk,
...
 
-  // Create all the sessions (note that `lk` can be released during creation,
-  // which is why we don't do this directly in the loop above).
+  // Create all the sessions without the lock held (the lock will be
+  // reacquired independently when the remote calls complete).
+  create_calls_in_progress_ += create_counts.size();
+  lk.unlock();
+  Status return_status;
   for (auto& op : create_counts) {
+    Status status;
```
devbww: `status` could be localized to the `kWait` case.

mr-salty:
___
*[google/cloud/spanner/internal/session_pool.cc, line 327 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1405#-M2eXuBKCZdvSnkW1O3_:-M2eXuBKCZdvSnkW1O3a:b-ata1r8) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/5ba8b234034d671ffb8df15a696a7d713fdf98c1/google/cloud/spanner/internal/session_pool.cc#L327)):*
> ```
> AsyncBatchCreateSessions
> ```

@coryan should I be saving the return value from this somewhere and call cancel() on it in the destructor?  or does it not matter?

mr-salty:
___
*[google/cloud/spanner/internal/session_pool.cc, line 327 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1405#-M2eXuBKCZdvSnkW1O3_:-M2eYBoB2Z6uCtxP0Cyw:bgfz3u9) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/5ba8b234034d671ffb8df15a696a7d713fdf98c1/google/cloud/spanner/internal/session_pool.cc#L327)):*

> ```
> AsyncBatchCreateSessions
> ```

@coryan should I be saving the return value from this somewhere and call cancel() on it in the destructor?  or does it not matter?
</blockquote></details>

to clarify - "this" meaning the `then` result (which I assume is `future<void>`?

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -323,24 +317,22 @@ Status SessionPool::CreateSessions(
...
+void SessionPool::CreateSessionsAsync(
+    std::shared_ptr<Channel> const& channel,
+    std::map<std::string, std::string> const& labels, int num_sessions) {
+  std::weak_ptr<SessionPool> pool = shared_from_this();
+  AsyncBatchCreateSessions(cq_, channel->stub, labels, num_sessions)
+      .then([pool, channel](
+                future<StatusOr<spanner_proto::BatchCreateSessionsResponse>>
+                    result) {
+        if (auto shared_pool = pool.lock()) {
+          shared_pool->HandleBatchCreateSessionsDone(channel,
```
devbww: Should we log something on failure here?

PR #1417, opened 2020-03-18
doc: fixes for how to cut a release

coryan:
The document describing how to cut a release was still referring to
`INSTALL.md`, then I realized that `ci/test-readme` really is just
a place to store scripts to generate markdown, so I renamed it too.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1417)


doc/cutting-a-release.md
```diff
@@ -31,15 +31,15 @@ git log --no-merges --format="format:* %s" \
 Update the instructions to build the library for development:
 
 ```bash
-./ci/test-readme/generate-readme.sh >README.md
+./ci/generate-markdown/generate-readme.sh >README.md
 ```
 
 ### Update INSTALL.md
```
devjgm: I think we should remove the "Update README" and "Update INSTALL" steps. Those scripts should be run whenever the README and packaging files are updated, not during the release. I think these steps are left over from when README used to contain release notes, but that's since been moved to CHANGELOG.md.

Note, I already removed these steps from the doc in -common: https://github.com/googleapis/google-cloud-cpp-common/pull/173

WDYT?

coryan: I think that sometimes we forget to update those files, and this is the last chance before a release. 
 If that is not convincing, I am willing to experiment and see if we forget "rarely", just say so and I will update this PR.

devjgm: I vote for removing all unnecessary manual steps from the release process. 

IFF we end up forgetting to update these files, then maybe we could better fix that problem with some type of presubmit check that complains if README.md is updated w/o also touching the generate-readme.sh script. And/or vice versa.

coryan: Done, PTAL.

mr-salty: I thought I had filed an issue about checking the contents of auto-generated in the clang-tidy build or similar but I can't find it - anyone else aware of that (or willing to file a bug)?

devjgm: I didn't see an issue but I'll do this. PR coming up.

PR #1421, opened 2020-03-18
feat: adds check-markdown.sh script to verify autogen markdown files

devjgm:
This script ensures that `README.md` and `doc/packaging.md`, both of which are auto-generated markdown files, remain consistent with the scripts that autogenerate them. This check will run as part of the "clang-tidy" build.

Related to https://github.com/googleapis/google-cloud-cpp-spanner/pull/1417

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1421)


coryan:
The coverage build will probably hang and then freeze.

PR #1425, opened 2020-03-19
feat!: implement SessionPool session refresh

### LINKED ISSUE #1171: Implement SessionPool session refresh
The `SessionPool` should periodically refresh idle `Sessions` to prevent them from expiring on the backend.

The backend expires sessions after 1 hour [(see docs)](https://cloud.google.com/spanner/docs/sessions) so we need to refresh at a slightly shorter interval. We already have an option for this in [`SessionPoolOptions`](https://github.com/googleapis/google-cloud-cpp-spanner/blob/e33b4f814e7c30f5c055431d07ea4ae99ac619f0/google/cloud/spanner/session_pool_options.h#L138) which defaults to 55 minutes.

The recommended mechanism for refreshing sessions is calling [`GetSession`](https://github.com/googleapis/googleapis/blob/5ef42bcd363ba0440f0ee65b3c80b499e9067ede/google/spanner/v1/spanner.proto#L102).

devbww:
Call `GetSession()` during `SessionPool::DoBackgroundWork()` for any
session whose last-use time is older than the keep-alive interval.

Also change the type of `SessionPoolOptions::keep_alive_interval_`
from `std::chrono::minutes` to `std::chrono::seconds`.  This allows
the running time of the test to be drastically reduced. It is not
an API breakage because `minutes` convert implicitly to `seconds`.

Fixes #1171.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1425)

### DIFF
--- a/google/cloud/spanner/internal/session.h
+++ b/google/cloud/spanner/internal/session.h
@@ -40,7 +40,7 @@ class Session {
       : session_name_(std::move(session_name)),
         channel_(std::move(channel)),
         is_bad_(false),
-        last_use_time_(std::chrono::steady_clock::now()) {}
+        last_use_time_(Clock::now()) {}
 
   // Not copyable or moveable.
   Session(Session const&) = delete;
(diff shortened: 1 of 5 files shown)


mr-salty:
___
*[google/cloud/spanner/internal/session_pool_test.cc, line 421 at r1](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1425#-M2nN64685M2zVPo9-Qv:-M2nnnOL9qCi6yOcMkqS:b-cipsir) ([raw file](https://github.com/googleapis/google-cloud-cpp-spanner/blob/044143176703f5c6773f5d870264f131170b0fe0/google/cloud/spanner/internal/session_pool_test.cc#L421)):*
<details><summary><i>Previously, devbww (Bradley White) wrote…</i></summary><blockquote>

I would contend that the 5-second timer is significant.  I have to know that `RefreshExpiringSessions()` has been called at least once, and for coverage reasons I want to know that it has been called at most once.
</blockquote></details>

I meant, I am not sure the duration matters at all when using a `MockCompletionQueue`

PR #1438, opened 2020-03-19
testing: improvements for Backup related tests

### LINKED ISSUE #1429: Consider running integration tests for backups with shorter timeouts
We recently disabled the backup integration tests because they were flaking out after running for 90 (**!!**) minutes.  We discussed re-enabling the tests but with shorter timeouts.  I prototyped this idea in:

https://github.com/googleapis/google-cloud-cpp-spanner/compare/master...coryan:shorten-retry-policies-in-tests?expand=1

But I think it might not work as well as we would like. Yes, the tests will finish earlier, and it is preferable to flake in 30 minutes over 90, but they will still flake.  We could ignore errors caused by really long timeouts, but then the test loses most of its value? And cleanup becomes **very** difficult, because one cannot delete a Cloud Spanner Database with pending backups and/or restores.

### LINKED ISSUE #1432: Delete backups in an instance first before deleting a stale instance
We have some cleanup code in some integration tests for deleting stale Spanner instances. It turned out that we can not delete instance with any backups. We need to recursively delete backups in an instance first before deleting the instance itself.

tmatsuo:
* enable backup related test for the coverage build
* use shorter timeout for admin clients for backup tests
* correctly cleanup stale instances by deleting backups first

Fixes #1432 and fixes #1429

We may still want to disable backup related tests for the coverage build. If so, we can run the backup related tests in nightly build.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1438)

### DIFF
--- a/ci/kokoro/docker/build.sh
+++ b/ci/kokoro/docker/build.sh
@@ -84,8 +84,8 @@ elif [[ "${BUILD_NAME}" = "coverage" ]]; then
   export DISTRO_VERSION=31
   : "${RUN_INTEGRATION_TESTS:=$DEFAULT_RUN_INTEGRATION_TESTS}"
   if [[ -z "${RUN_SLOW_INTEGRATION_TESTS:-}" ]]; then
-    # Defaults to "instance" for the coverage test.
-    export RUN_SLOW_INTEGRATION_TESTS=instance
+    # Defaults to run all the slow tests for the coverage test.
+    export RUN_SLOW_INTEGRATION_TESTS=instance,backup
   fi
   in_docker_script="ci/kokoro/docker/build-in-docker-cmake.sh"
 elif [[ "${BUILD_NAME}" = "publish-refdocs" ]]; then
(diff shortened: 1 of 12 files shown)


tmatsuo:
Once all the test pass, maybe I'll trigger 5 coverage builds almost simultaneously.

tmatsuo:
The coverage build took 35:43, better than before :)

@coryan Thanks for the review, I'll split the PR, but before doing so, let me run 5 simultaneous coverage builds as an endurance test.

tmatsuo:
I started 5 coverage-presubmit builds

tmatsuo:
All the 5 coverage-presubmit builds passed with the latency of 33:21, 35:14, 35:49, 34:52, and 36:11.

Much better :)

tmatsuo:
FYI, it successfully deleted an instance `temporary-instance-2020-03-18-hogehoge` with a backup in my project :)

coryan:
Pro tip: to fix more than one bug you cannot just say `Fixes #NNN #MMM` you need to say `Fixes #NNN and fixes #MMM`.

PR #1449, opened 2020-03-26
feat: allow injecting a Clock into the `SessionPool` for testing

### LINKED ISSUE #1428: SessionPool needs a way to inject a clock for testing
In https://github.com/googleapis/google-cloud-cpp-spanner/pull/1425, @coryan said:

> > google/cloud/spanner/internal/session_pool_test.cc, line 389 at r1 (raw file):

> > // NOTE: This test runs in real time. SessionPool does not currently provide

> Maybe create a bug and leave a TODO(#...) here to change SessionPool? We could change it 
> (SessionPool) to be a template class, e.g.:
>
> `template <typename Clock = std::chrono::steady_clock> class SessionPool`
> as this class is in the internal namespace that would not break anything, and we only need to use 
> it with a weird clock in the tests.

Alternatively, we may not need to make the whole class a template to inject the clock, we may be able to simply inject a function that produces a "now", maybe something like `std::function<std::chrono::steady_clock::time_point()> now`

mr-salty:
fixes #1428

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1449)


google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,40 @@
...
+#include <chrono>
+#include <functional>
+
+namespace google {
+namespace cloud {
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+
+/// Until we need something more complicated, use `steady_clock` as our Clock.
```
devbww: Can we just make the return type a particular (sufficiently precise) `time_point` instead of saying anything about `Clock` and `Clock::time_point`?  Then the "now source" could be backed by `system_clock`, `steady_clock`, or anything else.  The fake clock, after all, doesn't give wall time, and neither is it steady.

google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,40 @@
...
+using Clock = std::chrono::steady_clock;
+
+/// A function type that is expected to yield a `time_point` representing "now".
+using ClockNowSource = std::function<Clock::time_point()>;
```
devbww: We already live in a world, with our retry and polling loops(*), where non-test code wants to "sleep" until the clock reaches some point (or an interval has passed).  So, I would suggest that we should support a richer interface than just "now".

[*] At the moment we have to override the policy parameters to avoid long-blocking tests.

devjgm:
I believe this PR is no longer needed since https://github.com/googleapis/google-cloud-cpp-spanner/pull/1452

PR #1452, opened 2020-03-26
feat: provide a mockable `Clock` class

### LINKED ISSUE #1428: SessionPool needs a way to inject a clock for testing

mr-salty:
Like #1449 but different; this should make it easier to add operations
in the future (like sleep/wait), at the cost of having to pass it around
as a `shared_ptr` even for the `RealClock`.  (In #1449 we only needed
the `shared_ptr` in the fake case.)

fixes #1428

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1452)


google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,78 @@
...
+class Clock {
+ public:
+  using time_point = std::chrono::steady_clock::time_point;
+  using duration = std::chrono::steady_clock::duration;
+
+  virtual ~Clock() = default;
+  virtual time_point Now() = 0;
+};
+
+class RealClock : public Clock {
```
devbww: Should sub-class definitions be separate from the base class?

google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,78 @@
...
+ public:
+  time_point Now() override { return std::chrono::steady_clock::now(); }
```
devbww: Should this be in terms of `Clock::time_point` (or some similar base-class definition)?

google/cloud/spanner/internal/clock.h
devbww: `RealClock` may be too close to "real time clock" (i.e., "wall time"), which this is not.

google/cloud/spanner/internal/session.h
```diff
@@ -36,11 +37,13 @@ namespace internal {
  */
 class Session {
  public:
-  Session(std::string session_name, std::shared_ptr<Channel> channel)
+  Session(std::string session_name, std::shared_ptr<Channel> channel,
+          std::shared_ptr<Clock> clock = std::make_shared<RealClock>())
```
devbww: Q. Why do we need to go to `shared_ptr`?  Shouldn't there be one clock?

google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,78 @@
...
+};
+
+class FakeClock : public Clock {
+ public:
+  /// Construct a `FakeClock` with an optional starting `time_point` of `now`.
+  explicit FakeClock(time_point now = {}) { now_ = now; }
```
devbww: Given the choice of `steady_clock::time_point`, and the insignificance of its absolute values, I question the need for the ctor argument and `SetTime()`.

google/cloud/spanner/internal/clock_test.cc
```diff
@@ -0,0 +1,57 @@
...
+namespace spanner {
+inline namespace SPANNER_CLIENT_NS {
+namespace internal {
+namespace {
+
+TEST(Clock, RealClock) {
+  std::shared_ptr<Clock> clock = std::make_shared<RealClock>();
+  auto now = clock->Now();
+  auto now2 = clock->Now();
+  EXPECT_LE(now, now2);
```
devbww: How do you know that?  I see no such guarantee.

google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * All implementations of this class are required to be thread-safe.
+ */
+class Clock {
+ public:
+  using time_point = std::chrono::steady_clock::time_point;
+  using duration = std::chrono::steady_clock::duration;
+
+  virtual ~Clock() = default;
+  virtual time_point Now() = 0;
```
devjgm: As discussed offline, I think `Now()` is logically a const operation as it's not modifying any logical state of the clock instance. It's simply reading the clocks state. If the clocks state is changed between successive calls of `Now()`, that's simply because "something else" mutated the clock's state. That has nothing to do with what `Now()` does.

That said, I know this may be a controversial opinion. I'm certainly not going to block this PR on my opinion. I'm happy to defer to @mr-salty or @devbww on this point.

google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * A simple `Clock` class that can be overridden for testing.
+ *
+ * All implementations of this class are required to be thread-safe.
+ */
+class Clock {
+ public:
+  using time_point = std::chrono::steady_clock::time_point;
+  using duration = std::chrono::steady_clock::duration;
```
devjgm: This alias isn't used.

google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,78 @@
...
+namespace internal {
+
+/**
+ * A simple `Clock` class that can be overridden for testing.
+ *
+ * All implementations of this class are required to be thread-safe.
+ */
+class Clock {
+ public:
+  using time_point = std::chrono::steady_clock::time_point;
```
devbww: Aside: I looked around for our use of `system_clock`.  It seems that `Timestamp` and backups certainly need it.  `CompletionQueue` could probably move to `steady_clock` for its timers.

google/cloud/spanner/internal/session_pool.cc
```diff
@@ -135,15 +138,14 @@ void SessionPool::MaintainPoolSize() {
 void SessionPool::RefreshExpiringSessions() {
   std::vector<std::pair<std::shared_ptr<SpannerStub>, std::string>>
       sessions_to_refresh;
-  Session::Clock::time_point now = Session::Clock::now();
-  Session::Clock::time_point refresh_limit =
-      now - options_.keep_alive_interval();
+  Clock::time_point now = clock_->Now();
+  Clock::time_point refresh_limit = now - options_.keep_alive_interval();
```
devjgm: Optional nit: consider `auto` here if you think it reads better.

google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * `SteadyClock` is a monotonic clock where time points cannot decrease as
+ * physical time moves forward. It is not related to wall clock time.
+ */
+using SteadyClock =
+    ::google::cloud::spanner::internal::Clock<std::chrono::steady_clock>;
+
+/**
+ * `SystemClock` reprsents the system-wide real time wall clock.
```
devbww: s/reprsents/represents/

google/cloud/spanner/internal/clock.h
```diff
@@ -0,0 +1,66 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_INTERNAL_CLOCK_H
+#define GOOGLE_CLOUD_CPP_SPANNER_GOOGLE_CLOUD_SPANNER_INTERNAL_CLOCK_H
+
+#include "google/cloud/spanner/version.h"
+#include <chrono>
+#include <mutex>
```
devbww: Unused.

devbww:
Looks like there are some header-guard mismatches.

mr-salty:
Header guards fixed as per https://github.com/googleapis/google-cloud-cpp/issues/3603 (see #1456)

PR #1459, opened 2020-04-01
chore: update CHANGELOG for release

scotthart:
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1459)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,32 +3,29 @@
 ## v1.2.0 - TBD
 
 ## v1.1.0 - 2020-04
+> **NOTICE:** This repo will be merging with the
+[googleapis/google-cloud-cpp](https://github.com/googleapis/google-cloud-cpp)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,34 @@
 # Changelog
 
-## v1.1.0 - TBD
+## v1.2.0 - TBD
+
+## v1.1.0 - 2020-04
+
```
devjgm, line 6: I think you want to include the following here:

```
> **NOTICE:** This repo will be merging with the
[googleapis/google-cloud-cpp](https://github.com/googleapis/google-cloud-cpp)
monorepo in 60 days, around **June 1, 2020**. This change should not break any
users who are using a released version. We expect this new monorepo
organization to be better for our existing and new customers going forward.
Please see
[googleapis/google-cloud-cpp#3612](https://github.com/googleapis/google-cloud-cpp/issues/3612)
for more details.
```

This will match what we did in `-common`

PR #1462, opened 2020-04-02
fix: fully qualify spanner names outside namespace-alias scope

### LINKED ISSUE #1461: namespace alias may be problematic
In ...

1. [google/cloud/spanner/benchmarks/benchmarks_config.cc](https://github.com/googleapis/google-cloud-cpp-spanner/blob/d50058713f80d00e4954c9da75218a9b21be1340/google/cloud/spanner/benchmarks/benchmarks_config.cc#L27)
2. [google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc](https://github.com/googleapis/google-cloud-cpp-spanner/blob/d50058713f80d00e4954c9da75218a9b21be1340/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc#L50)
3. [google/cloud/spanner/benchmarks/single_row_throughput_benchmark.cc](https://github.com/googleapis/google-cloud-cpp-spanner/blob/d50058713f80d00e4954c9da75218a9b21be1340/google/cloud/spanner/benchmarks/single_row_throughput_benchmark.cc#L31)
4. [google/cloud/spanner/samples/mock_execute_query.cc](https://github.com/googleapis/google-cloud-cpp-spanner/blob/d50058713f80d00e4954c9da75218a9b21be1340/google/cloud/spanner/samples/mock_execute_query.cc#L29)

we introduce a namespace-scoped, namespace alias for `spanner`
```
namespace spanner = ::google::cloud::spanner;
```
When any of these files includes a header that we don't fully control, or which doesn't provide sufficient cleanliness guarantees, we run the risk of clashing with an existing `namespace spanner`.

In cases 2 and 3. it is the (transitive) inclusion of generated protobuf headers that actually does cause a problem when the code is built in an environment that many of the repo owners will be familiar with.

It is probably worth our while to do something to avoid that particular clash.

devbww:
Do not expect a namespace alias declared within an anonymous namespace
to work outside of the anonymous namespace, as it could be ambiguous in
that context.

Fixes #1461.

---
This change is [Reviewable](https://reviewable.io/reviews/googleapis/google-cloud-cpp-spanner/1462)

### DIFF
--- a/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
+++ b/google/cloud/spanner/benchmarks/multiple_rows_cpu_benchmark.cc
@@ -147,14 +147,14 @@ int main(int argc, char* argv[]) {
     config.database_id =
         google::cloud::spanner_testing::RandomDatabaseName(generator);
   }
-  spanner::Database database(config.project_id, config.instance_id,
-                             config.database_id);
+  google::cloud::spanner::Database database(
+      config.project_id, config.instance_id, config.database_id);
 
   // Once the configuration is fully initialized and the database name set,
   // print everything out.
   std::cout << config << std::flush;
 
-  spanner::DatabaseAdminClient admin_client;
+  google::cloud::spanner::DatabaseAdminClient admin_client;
   std::vector<std::string> additional_statements = [&available, generator] {
     std::vector<std::string> statements;
     for (auto const& kv : available) {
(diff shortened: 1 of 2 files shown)


coryan:
Looks like Kokoro missed this PR, consider rebasing and pushing again, that usually wakes Kokoro up.

